PR #2, opened 2018-08-14
Add initial django project

sliverc:


docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '2'
+services:
+  db:
+    image: postgres
```
anehx: Lock version?

.travis.yml
```diff
@@ -0,0 +1,23 @@
+language: python
+
+python:
+  - "3.6"
+
+services:
+  - postgresql
```
anehx: Lock version to be the same as in `docker-compose.yml`

sliverc:
@anehx Locked and fixed CI

docker-compose.yml
```diff
@@ -1,20 +1,18 @@
...
   db:
-    image: postgres
+    image: postgres:9.3
     ports:
       - '5432:5432'
     environment:
-      - POSTGRES_USER=caluma
-      - POSTGRES_PASSWORD=caluma
+      - POSTGRES_USER={{cookiecutter.project_name}}
+      - POSTGRES_PASSWORD={{cookiecutter.project_name}}
```
anehx: That's not correct I think...

sliverc: Editor was playing fool with me... :) its fixed now.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '2'
```
anehx: Why not the latest version `3`?

setup.py
```diff
@@ -0,0 +1,13 @@
+"""Setuptools package definition."""
+
+from setuptools import find_packages, setup
+
+setup(
+    name='caluma',
+    version='0.0.0',
+    author='Adfinis SyGroup AG',
+    author_email='https://adfinis-sygroup.ch/',
```
anehx: Are we sure that "Adfinis SyGroup AG" should be the author? Also, this email is not an email address..

setup.py
```diff
@@ -0,0 +1,13 @@
...
+    description='Caluma Service providing GraphQL API ',
+    url='https://adfinis-sygroup.ch/',
+    packages=find_packages(),
```
anehx: Maybe add our license model here aswell? https://docs.python.org/2/distutils/setupscript.html#additional-meta-data

docker-compose.yml
sliverc: Fair point. Updated.

setup.py
sliverc: Removed author and author_emai. Not really relevant right now anyway.

setup.py
sliverc: license added.

sliverc:
ready for review again

PR #9, opened 2018-08-16
Add all forms connection

sliverc:
For jsonb minimum version of Postgres 9.4 is needed


caluma/form/tests/test_form.py
```diff
@@ -0,0 +1,23 @@
...
+            meta
+          }
+        }
+      }
+    }
+    """
+
+    result = schema.execute(query)
+    assert not result.errors
+    snapshot.assert_match(result.data)
```
anehx: Maybe add a test for fetching a single form (node graph) aswell?

sliverc: Added a generic global node test.

PR #15, opened 2018-08-20
Add form mutation save and delete

sliverc:


caluma/form/tests/snapshots/snap_test_form.py
```diff
@@ -0,0 +1,32 @@
...
+from __future__ import unicode_literals
+
+from snapshottest import Snapshot
+
+
+snapshots = Snapshot()
+
+snapshots["test_save_form 1"] = {
+    "saveForm": {
+        "clientMutationId": None,
```
anehx: Could you pass a `clientMutationId` to verify if it returns the same id as passed?

sliverc: Good point. Added test.

sliverc:
Ready for review again.

PR #20, opened 2018-08-23
Add question graph

### LINKED ISSUE #18: Errors not working correctly
Currently, the `saveForm` mutation does not succeed if the description is missing (#17) but fails to throw an error. The response simply returns `null` for the property `form`.

sliverc:
Fixes #18

Shows serializer validation errors as graphql errors


caluma/form/serializers.py
```diff
@@ -7,3 +8,26 @@ class FormSerializer(serializers.ModelSerializer):
...
+        jexl = JEXL()
+        # TODO: define transforms e.g. answer
+        errors = list(jexl.validate(value))
+        if errors:
+            raise exceptions.ValidationError(errors)
+
+        return value
+
+    def validate_is_hidden(self, value):
+        return self.validate_is_required(value)
```
anehx: Maybe a `_validate_jexl` method which is then called by `validate_is_required` and `validate_is_hidden` would feel a bit cleaner, don't you think?

sliverc: Fair enough. Changed it.

PR #26, opened 2018-08-27
 Localize name and description on form and question

sliverc:
* API allows all supported languages of django and doesn't set a default
* language is passed on as http header
* increase minimum Postgres version 9.6 to support HStore field


caluma/faker.py
```diff
@@ -0,0 +1,17 @@
+from faker import Faker
```
winged: I would rename this file. Depending on where you import it from, it may not be clear if the actual `faker` package is imported or our own.

caluma/form/migrations/0004_auto_20180827_1105.py
```diff
@@ -0,0 +1,43 @@
...
+
+    dependencies = [("form", "0003_auto_20180824_1334")]
+
+    operations = [
+        # needed till https://github.com/SectorLabs/django-localized-fields/pull/55
+        # is merged
+        HStoreExtension(),
+        migrations.AlterModelOptions(
+            name="formquestion", options={"ordering": ("-sort", "id")}
+        ),
```
anehx: Why is this in here?

sliverc: makemigrations created this as when this alteration was added makemigrations was not called (this change is actually not database relevant but makemigrations doesn't know this at time of creation)

caluma/settings.py
```diff
@@ -28,8 +28,10 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 # Application definition
 
 INSTALLED_APPS = [
-    "django.contrib.contenttypes",
     "django.contrib.postgres",
+    "localized_fields",
+    "psqlextra",
```
winged: Is the `psqlextra` app a dependency of `localized_fields`? Or why do we need this?

sliverc: yes localized_fields requires psqlextra (it add extra feature to HStoreField)

caluma/faker.py
sliverc: faker is a submodule of caluma so I would need to import as `from caluma.faker import ...`

same thing happens with models.py, serializers.py etc. which also have the same naming as upstream so I would prefer keeping the naming to keep it consistent with upstream.

PR #64, opened 2018-09-19
Refactor mutation to use serializers

### LINKED ISSUE #49: Use rest framework serializer in all create and update mutations
Currently only some mutations use `SerializerMutation` but it could actually be used by all just with different serializer.

This way validation could always happen in serializer and not with helper methods on the model.

### LINKED ISSUE #58: GlobalIDPrimaryKeyRelatedField should allow not encoded primary key as input type
This is needed to make static written mutations easier readible which could be loaded like written in #50 

For this to work in all mutations #49 needs to be implemented as well.

### LINKED ISSUE #63: Do not distinguish between trivial and non trivial change
API can not decide whether a change is trivial. This needs to be decided by the user and the API should not restrict it.

This means following mutation should always be allowed even when published:
* AddWorkflowSpecificationFlow
* RemoveWorkflowSpecificationFlow
* AddFormQuestion
* RemoveFormQuestion

SetWorkflowSpecificationStart mutation should be integrated into SaveWorkflowSpecification again.

sliverc:
Fixes #63 
Fixes #49 
Fixes #58 

Simplifying mutations and making serializer be responsible for validation.


caluma/mutation.py
```diff
@@ -127,12 +105,14 @@ def __init_subclass_with_meta__(
...
+        lookup_input_kwarg = cls._meta.lookup_input_kwarg
         model_class = cls._meta.model_class
 
         if model_class:
-            if "update" in cls._meta.model_operations and lookup_field in input:
+            if "update" in cls._meta.model_operations and lookup_input_kwarg in input:
                 instance = get_object_or_404(
-                    model_class, **{lookup_field: input[lookup_field]}
+                    model_class,
+                    **{lookup_field: extract_global_id(input[lookup_input_kwarg])}
```
anehx: As discussed, add some documentation for `lookup_input_kwarg`

sliverc: Good point. I have added documentation to all possible meta options on `SerializerMutation`.

PR #73, opened 2018-09-25
Introduce configuration options for integer, text and float questions

sliverc:
Validate answers whether those are correct according to the question type and configuration.


caluma/form/schema.py
```diff
@@ -62,11 +66,29 @@ class Meta:
...
 
 class IntegerQuestion(graphene.ObjectType):
+    max_value = graphene.Int()
+    min_value = graphene.Int()
+
+    def resolve_max_value(self, info):
+        return self.configuration["max_value"]
+
+    def resolve_min_value(self, info):
+        return self.configuration["min_value"]
```
anehx: You don't need those - delete 'em!

caluma/form/schema.py
```diff
@@ -62,11 +66,29 @@ class Meta:
...
 
 class FloatQuestion(graphene.ObjectType):
+    min_value = graphene.Float()
+    max_value = graphene.Float()
+
+    def resolve_max_value(self, info):
+        return self.configuration["max_value"]
+
+    def resolve_min_value(self, info):
+        return self.configuration["min_value"]
```
anehx: Same

caluma/form/schema.py
sliverc: True. Deleted. Also changed that max and min values are non null in query as we always have a default.

caluma/form/schema.py
sliverc: same as above

PR #75, opened 2018-09-25
Add options to checkbox and radio question types

sliverc:


caluma/form/schema.py
```diff
@@ -201,6 +241,7 @@ class Mutation(object):
     save_checkbox_question = SaveCheckboxQuestion().Field()
     save_float_question = SaveFloatQuestion().Field()
     save_integer_question = SaveIntegerQuestion().Field()
+    save_question_option = SaveQuestionOption().Field()
```
anehx: Maybe we should be able to delete it as well :wink:

sliverc:
Refactored as discussed. Options can now be saved before question.

Save checkbox and radio question now requires setting of options. Ordering of options is solved this way as well.

Ready for review again.

anehx:
CI is failing.. Otherwise LGTM :+1:

PR #82, opened 2018-09-28
Rename Form to FormSpecification and Document to Form

sliverc:
Fixes #41 

Notice:
* makes api more consistent
* integrates document app into form app
* needs recreaction of migrations 
* not backwards compatible, changes in form builder needed

PR #89, opened 2018-10-08
Add form as workflow data

### LINKED ISSUE #88: Implement workflow data pattern using form
Implement Workflow data pattern as described in http://www.workflowpatterns.com/patterns/data/visibility/wdp7.php

For data we use a form which is defined in workflow specification and instantiated when starting workflow.

sliverc:
Fixes #88 

* Form specification can optionally be defined during workflow specification
* Form is created when workflow is started

#82 needs to be merged first.


sliverc:
Closing this PR till #90 is implemented as this means different naming.

PR #95, opened 2018-10-09
 Add document as case data 

sliverc:
Fixes #89 

#94 to be merged first


caluma/workflow/serializers.py
```diff
@@ -14,7 +15,7 @@ def __init__(self, **kwargs):
 class SaveWorkflowSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Workflow
-        fields = ("slug", "name", "description", "meta", "start")
+        fields = ("slug", "name", "description", "meta", "start", "form", "form")
```
anehx: `form` is duplicated

sliverc: :smile: Fixed.

Issue #103, opened 2018-10-15
Add configuration options to README.md


sliverc:
All configuration options should be documented in README.md.

Especial notice should be taken for extension points (e.g. permission, visibility) to document them how to use it and issue should be filed if more extension points are needed.

Idea is that Caluma is used as a service and not as a django app.

PR #107, opened 2018-10-16
Add visibility extension layer

### LINKED ISSUE #104: Visibility classes extension
Queryset needs to be filtered per node type to restrict visibility. For this a new configuration option (extension point) is added named `VISIBILITY_CLASSES` where a list of visibility classes can be configured. Those classes will be chained.

Following classes would be implemented to start with:
`BaseVisibility` see below for details
`Authenticated` only authenticated users may see any nodes, otherwise none.
`All` all items are visible (default)

`BaseVisibility` allows overwriting of `get_queryset` for any node type by naming convention. e.g. for node form it could be overwritten with method `get_queryset_for_form`.

sliverc:
Fixes #104 

Documentation will be addressed with #103

### DIFF
--- a/caluma/conftest.py
+++ b/caluma/conftest.py
@@ -43,6 +43,7 @@ def snapshot(request):
 
 @pytest.fixture
 def info(rf):
+    """Mock for GraphQL resolve info embedding django request as context."""
     request = rf.get("/graphql")
     request.user = AnonymousUser()
 
@@ -62,6 +63,7 @@ def info(rf):
 
 @pytest.fixture
 def admin_info(rf):
+    """Mock for GraphQL resolve info embedding authenticated django request as context."""
     request = rf.get("/graphql")
     request.user = OIDCUser({"sub": "admin"})
 
--- a/caluma/conftest.py
+++ b/caluma/conftest.py
@@ -3,12 +3,14 @@ import inspect
 import pytest
 from factory import Faker
 from factory.base import FactoryMetaClass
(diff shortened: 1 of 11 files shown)


caluma/conftest.py
```diff
@@ -37,3 +39,41 @@ def assert_execution_result(self, result):
 
     with GraphQlSnapshotTest(request) as snapshot_test:
         yield snapshot_test
+
+
+@pytest.fixture
+def info(rf):
```
winged: What is this info fixture? Might need a docstring

caluma/settings.py
```diff
@@ -125,3 +126,12 @@ def parse_admins(admins):  # pragma: no cover
 OIDC_VALIDATE_CLAIMS_OPTIONS = env.dict(
     "OIDC_VALIDATE_CLAIMS_OPTIONS", cast={"value": bool}, default=None
 )
+
+# Extensions
+
+VISIBILITY_CLASSES = [
+    import_string(cls)()
```
winged: In my understanding, when you call `import_string(cls)()` like this, you'll get an instance, not a class. So either the setting's name is wrong or you should only fetch the classes

caluma/conftest.py
sliverc: Added docstring.

caluma/settings.py
sliverc: fair point. changed it so there are actual classes here.

PR #115, opened 2018-10-17
Introduce created_by_user and created_by_group for all objects

### LINKED ISSUE #114: Assign creating user and group
On all models user and group should be stored as `created_by_user` and `created_by_group`.

Such should be automatically set to currently logged in user with group or None if not authenticated or no group is available.

This data can then be used as multi tendency feature (e.g. may only see the resources created within your group)

sliverc:
Fixes #114 

* Renamed `created` to `createdAt` and `modified` to `modifiedAt` for consistency
* Introduce new configuration option OIDC_GROUPS_CLAIM
* Add CreatedByGroup visibility class to only view nodes which are created by node user belongs to


sliverc:
@anehx 
Would be great if you could have a look at this as I will need it for integration. Thanks.

caluma/tests/snapshots/snap_test_schema.py
```diff
@@ -39,8 +39,8 @@
 
 interface Answer {
   id: ID
-  created: DateTime!
-  modified: DateTime!
+  createdAt: DateTime!
+  modifiedAt: DateTime!
```
anehx: Where is `createdByUser` and `createdByGroup`?

caluma/tests/snapshots/snap_test_schema.py
```diff
@@ -361,8 +381,8 @@
 
 interface Question {
   id: ID!
-  created: DateTime!
-  modified: DateTime!
+  createdAt: DateTime!
+  modifiedAt: DateTime!
```
anehx: Where is `createdByUser` and `createdByGroup`?

caluma/tests/snapshots/snap_test_schema.py
sliverc: Very good point. I missed adding it to the interfaces. Changed now.

caluma/tests/snapshots/snap_test_schema.py
sliverc: added here as well.

anehx:
Well, sad story.. But I think if static data is the only alternative we should tolerate this.

PR #132, opened 2018-11-23
Implement hooks to override default permissions

sliverc:
This allows custom extensions to change default permission.

### DIFF
--- a/caluma/permissions.py
+++ b/caluma/permissions.py
@@ -4,6 +4,15 @@ from graphene.utils.str_converters import to_snake_case
 class BasePermission(object):
     """Basic permission class to be extended by any permission implementation."""
 
+    def has_base_permission(self, mutation, info):
+        """
+        Check base permission.
+
+        Override to implement default permission for all nodes.
+        Defaults to True.
+        """
+        return True
+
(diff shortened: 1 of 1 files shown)


sliverc:
Some code as an example:
```python
from caluma.permissions import BasePermission

class CustomPermission(BasePermission):
    def has_base_permission(self, mutation, info):
        roles = info.context.user.decoded_token.get("caluma_roles", [])
        return "admin" in roles or "company_admin" in roles

    def has_permission_for_start_case(self, mutation, info):
        return True

    def has_permission_for_complete_work_item(self, mutation, info):
        return True

    def has_permission_for_save_document(self, mutation, info):
        return True
```

In this case only admin and company admin are allowed to run all mutations, all other users may only start case, complete work item and save document.

If we had to implement it with the current `BasePermission` someone would need to overwrite `has_permission` and implement finding of overwriting methods itself.

If you see a better way then this approach in this PR please let me know. But as far as I see it this would be the easiest way.

sliverc:
As discussed api needs to be rewritten so it is typo safe.

Closing this PR and will open a new one later one.

PR #135, opened 2018-11-26
Make api of visibility and permission extensions more robust

sliverc:
Instead of relying on naming rather use decorator with referencing import directly. This way typos can be noticed during import.

### DIFF
--- a/caluma/mutation.py
+++ b/caluma/mutation.py
@@ -4,6 +4,7 @@ import graphene
 from django.conf import settings
 from django.http import Http404
 from django.shortcuts import get_object_or_404
+from django.utils.module_loading import import_string
 from graphene.relay.mutation import ClientIDMutation
 from graphene.types import Field, InputField
 from graphene.types.mutation import MutationOptions
@@ -66,6 +67,8 @@ class Mutation(ClientIDMutation):
 
     class Meta:
         abstract = True
+
+    permission_classes = [import_string(cls) for cls in settings.PERMISSION_CLASSES]
 
     @classmethod
     def __init_subclass_with_meta__(
@@ -171,13 +174,13 @@ class Mutation(ClientIDMutation):
 
     @classmethod
     def check_permissions(cls, root, info):
-        for permission_class in settings.PERMISSION_CLASSES:
+        for permission_class in cls.permission_classes:
             if not permission_class().has_permission(cls, info):
                 raise exceptions.PermissionDenied()
 
     @classmethod
     def check_object_permissions(cls, root, info, instance):
-        for permission_class in settings.PERMISSION_CLASSES:
+        for permission_class in cls.permission_classes:
             if not permission_class().has_object_permission(cls, info, instance):
                 raise exceptions.PermissionDenied()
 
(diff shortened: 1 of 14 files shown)


caluma/mutation.py
```diff
@@ -125,7 +128,7 @@ def __init_subclass_with_meta__(
         _meta.return_field_type = return_field_type
 
         input_fields = yank_fields_from_attrs(input_fields, _as=InputField)
-        super(SerializerMutation, cls).__init_subclass_with_meta__(
+        super(Mutation, cls).__init_subclass_with_meta__(
```
winged: Is there a reason why we're using python2-style `super()` here?

caluma/permissions.py
```diff
@@ -1,36 +1,87 @@
...
 
-    def has_object_permission(self, mutation, info, instance):
-        """
-        Check object permission of specific mutation or otherwise return True.
+        object_perm_fns = inspect.getmembers(
+            self, lambda m: hasattr(m, "_object_permission")
+        )
+        self._object_permissions = {
+            fn._object_permission: fn for _, fn in object_perm_fns
+        }
```
winged: It'd probably be a good idea to safeguard against duplicate permission methods here. Throw an exception if two methods are tagged for the same permission class

caluma/visibilities.py
```diff
@@ -1,31 +1,54 @@
...
 
-        Checks for node specific implement by its snake name.
 
-        e.g. for Node WorkItem it will call `get_queryset_for_work_item` if available.
-        """
-        queryset = self.get_base_queryset(node, queryset, info)
+class BaseVisibility(object):
+    """Basic visibility classes to be extended by any visibility implementation.
+
+    In combination with the decorator `@queryset_for` a custom visibility class
```
winged: Naming - the decorator is called `filter_queryset_for`, not `queryset_for`

caluma/visibilities.py
```diff
@@ -1,31 +1,54 @@
...
+        self._filter_querysets_for = {
+            fn._filter_queryset_for: fn for _, fn in queryset_fns
+        }
 
-        method_name = f"get_queryset_for_{to_snake_case(node.__name__)}"
-        get_queryset_for_node = getattr(self, method_name, None)
-        if get_queryset_for_node is not None:
-            queryset = get_queryset_for_node(node, queryset, info)
+    def get_queryset(self, node, queryset, info):
+        for cls in reversed(node.mro()):
```
winged: Just out of curiosity: Why are you iterating in reversed MRO here? I think filtering is "additive", so shouldn't really matter

caluma/mutation.py
sliverc: Changed

caluma/permissions.py
sliverc: Added safeguarding

caluma/visibilities.py
sliverc: well spotted. Changed.

caluma/visibilities.py
sliverc: Removed it as it makes no difference and is just more code... :)

sliverc:
@winged 
Changes for loading `permission_classes` and `visibility_classes` only once app is ready needs a lot of files being moved around.
Hence I will create a single PR for this and this PR can be reviewed/merged before.

PR #136, opened 2018-11-27
 Load visibility and permission classes once app is ready 

sliverc:
For this to work a new `core` app needed to be introduced.

#135 needs to be merged first


caluma/core/mutation.py
```diff
@@ -171,13 +174,16 @@ def get_object(cls, root, info, queryset, **input):
 
     @classmethod
     def check_permissions(cls, root, info):
-        for permission_class in settings.PERMISSION_CLASSES:
+        assert (
```
winged: `assert` statements may be stripped if user runs it with optimisation, potentially skipping this runtime check. See https://docs.python.org/2.0/ref/assert.html

sliverc: Changed all asserts to ImproperlyConfigured exception. Mind having a look again?

PR #142, opened 2018-12-05
Allow canceling of case

sliverc:
Allows to withdraw from a case at any point. When withdrawal is allowed is specified by permission.
All not completed work items are canceled when case is canceled.


sliverc:
depends on fix in #143

PR #144, opened 2018-12-05
Add possibility to start sub-workflow on a work item

sliverc:
Enforcement of whether a sub-workflow may be created on a certain task is handled on the permission layer.


caluma/workflow/models.py
```diff
@@ -84,6 +84,17 @@ class WorkItem(UUIDModel):
     task = models.ForeignKey(
         Task, on_delete=models.DO_NOTHING, related_name="work_items"
     )
-    case = models.ForeignKey(Case, related_name="work_items", on_delete=models.CASCADE)
     status = models.CharField(choices=STATUS_CHOICE_TUPLE, max_length=50, db_index=True)
     meta = JSONField(default={})
+
+    case = models.ForeignKey(Case, related_name="work_items", on_delete=models.CASCADE)
+    child_case = models.OneToOneField(
```
winged: I'd reverse that relationship, or set the `on_delete` to `models.SET_NULL`. Otherwise if you delete the child case, the parent will be deleted with it.

sliverc: changed it to cascade SET_NULL. reasoning behind is that from a work item perspective it is necessary to check whether a child case is not null each time a work item is completed. The other way around is not really used unless requested in a GraphQL query.

winged: :+1:

PR #151, opened 2018-12-12
Add documentation on how to get started and contribution

### LINKED ISSUE #103: Add configuration options to README.md

sliverc:
Fixes #103 

This includes how to configure Caluma.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,66 @@
+# Contributing
+
+Contributions to Caluma Service are very welcome! Best have a look at the open [issues](https://github.com/projectcaluma/caluma/issues)
+and open a [GitHub pull request](https://github.com/projectcaluma/caluma/compare). See instructions below how to setup development
+environment.
+
+## Setup development environment
+
+### Clone
+
+To work on Caluma you first need to clone
+
+```bash
+git clone https://github.com/projectcaluma/caluma.git
+cd caluma
+```
+
+### Open Shell
+
+Once it is cloned you can easily open a shell in the docker container to
+open an development environment.
+
+```bash
+# needed for permission handling
+# only needs to be run once
+echo UID=$UID > .env
+# open shell
+docker-compose run --rm caluma bash
+```
+
+### Testing
+
+Once you have shelled in docker container as described above
+you can use common python tooling for formatting, linting, testing
+etc.
+
+```bash
+# linting
+flake8
+# format code
+black .
+# running tests
+pytest
+# create migrations
+./manage.py makemigrations
+```
+
+### Install new requirements
+
+In case you needed to add new requirements you simply need to build the docker container
+again for those to be installed and re-open shell.
+
+```bash
+docker-compose build --pull
+```
+
+### Setup pre commit
+
+Pre commit hooks is an additional option instead of executing checks in your editor of choice.
+
+```bash
+# create virtualenv with tool of your choice
+pip install pre-commit
+pip install -r requiements-dev.txt -U
+pre-commit install
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Contributions to Caluma Service are very welcome! Best have a look at the open [issues](https://github.com/projectcaluma/caluma/issues)
 and open a [GitHub pull request](https://github.com/projectcaluma/caluma/compare). See instructions below how to setup development
-environment.
+environment. Before writing any code, best discuss your proposed change in a GitHub issue to see if the proposed change makes sense for the project.
 
 ## Setup development environment
 
@@ -45,9 +45,11 @@ pytest
 ./manage.py makemigrations
 ```
 
+Writing of code can still happen outside the docker container of course.
+
 ### Install new requirements
 
-In case you needed to add new requirements you simply need to build the docker container
+In case you're adding new requirements you simply need to build the docker container
 again for those to be installed and re-open shell.
 
 ```bash
@@ -58,8 +60,9 @@ docker-compose build --pull
 
 Pre commit hooks is an additional option instead of executing checks in your editor of choice.
 
+First create a virtualenv with the tool of your choice before running below commands:
+
 ```bash
-# create virtualenv with tool of your choice
 pip install pre-commit
 pip install -r requiements-dev.txt -U
 pre-commit install
--- a/README.md
+++ b/README.md
@@ -32,17 +32,17 @@ query and mutate form and workflow entities which are described below.
 
 ### Entities
 
-Caluma is split into two parts, form and workflow. The form can be used without a workflow and visa versa, but the full power of Caluma comes when combining the two.
+Caluma is split into two parts, form and workflow. The form can be used without a workflow and vice versa, but the full power of Caluma comes when combining the two.
 
-Each part consists of different entities, whereas an entity usually comes in pairs, a blueprint and a specific instance.
+Each part is based on several entities, which usually come in pairs: One defines a "blueprint" and is instance-independent, while the other one represents the blueprint in relation to a specific instance. The two entities are usually connected by a one-to-many relationship, e.g. there can be many concrete `cases` (instance-specific) following a certain `workflow` (global blueprint).
 
 #### Form entities
 
 **Form** defines the structure of a document, basically a container of questions.
 
-**Question** defines a single input of a form whereas a question has different implementation types defining how a question should be represented and validated.
+**Question** defines a single input of a form, such as a text input, checkbox or similar. A question has different implementation types defining how a question should be represented and validated.
 
-**Document** is a specific instance of a form being a container of answers.
+**Document** is a specific instance of a form. Since a form is a container of questions, a document can be visualized as a container of answers.
 
 **Answer** containing user input of a specific question. There are different implementations of answer covering different data types.
 
@@ -54,7 +54,7 @@ Naming and concept of workflow follows concept of [Workflow Patterns Initiative]
 
 **Task** is a definition of work being part of a workflow.
 
-**Flow** is a definition on what task resp. tasks follows another.
+**Flow** defines the ordering and dependencies between tasks.
 
 **Case** is a specific instance of a workflow.
 
@@ -70,7 +70,7 @@ Different environment variable types are explained at [django-environ](https://g
 
 A list of configuration options which you might need to configure to get Caluma started in your environment.
 
-* `SECRET_KEY`: A secret key used for cryptography. See [more](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SECRET_KEY)
+* `SECRET_KEY`: A secret key used for cryptography. This needs to be a random string of a certain length. See [more](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SECRET_KEY)
 * `ALLOWED_HOSTS`: A comma separated list of strings representing host/domans Caluma will be served on. See [more](https://docs.djangoproject.com/en/2.1/ref/settings/#allowed-hosts)
 * `DATABASE_NAME`: Name of database to use (default: caluma)
 * `DATABASE_USER`: username to use when connecting to the database (default: caluma)
@@ -79,9 +79,9 @@ A list of configuration options which you might need to configure to get Caluma 
 * `DATABASE_PORT`: port to use when connecting to Postgres database (default: 5432)
 * `LANGUAGE_CODE`: default language defined as fall back (default: en)
 
-#### OpenID Connect
+#### Authentication and authorization
 
-When you want to connect Caluma to your [IAM](https://en.wikipedia.org/wiki/Identity_management) supporting OpenID connect you need following settings.
+If you want connect Caluma you need a [IAM](https://en.wikipedia.org/wiki/Identity_management) supporting OpenID Connect. If not available you might want to consider using [Keycloak](https://www.keycloak.org/).
 
 * `OIDC_VERIFY_ALGORITHM`: Token verification algorithm (default: HS256)
 * `OIDC_CLIENT`: Client name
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
 
-In case you needed to add new requirements you simply need to build the docker container
-again for those to be installed and re-open shell.
+**Task** is a definition of work being part of a workflow.
+
+**Flow** is a definition on what task resp. tasks follows another.
+
+**Case** is a specific instance of a workflow.
+
+**WorkItem** is a specific instance of a task a resource has to work off.
```
winged: Generally I feel like this section could benefit from a concrete example, especially for readers who are not familiar with the concepts. Something along the following lines:

> A workflow could represent your company's hiring procedures. A task may be "evaluate application documents", another "send invitation" and another "send rejection". The flow would then define "send invitation" and "send rejection" as followups to the "evaluate" task.
> The Case would then represent the application of a single person who wants to start work at your company. "WorkItem" would mean specifically "review this person's documents".

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+### Configuration
+
+Caluma is a [12factor app](https://12factor.net/) so it is configurable per environment variables.
+Different environment variable types are explained at [django-environ](https://github.com/joke2k/django-environ#supported-types)
+
+#### Common
+
+A list of configuration options which you might need to configure to get Caluma started in your environment.
+
+* `SECRET_KEY`: A secret key used for cryptography. See [more](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SECRET_KEY)
```
winged: Add a note that this should be set to a random string of a certain length

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+* `DATABASE_NAME`: Name of database to use (default: caluma)
+* `DATABASE_USER`: username to use when connecting to the database (default: caluma)
+* `DATABASE_PASSWORD`: password to use when connecting to database
+* `DATABASE_HOST`: host to use when connecting to Postgres database (default: localhost)
+* `DATABASE_PORT`: port to use when connecting to Postgres database (default: 5432)
+* `LANGUAGE_CODE`: default language defined as fall back (default: en)
+
+#### OpenID Connect
+
+When you want to connect Caluma to your [IAM](https://en.wikipedia.org/wiki/Identity_management) supporting OpenID connect you need following settings.
```
winged: What if I don't? Do we support a standalone authorisation/authentication scheme?

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+
+#### Extension points
+
+Caluma is meant to be used as a service with a clean API hence it doesn't provide a Django app.
+For customization some clear extension points are defined. In case a customization is needed
+where no extension point is defined, best [open an issue](https://github.com/projectcaluma/caluma/issues/new) for discussion.
+
+##### Visibility classes
+
+Visibility classes define what user can read per node. Such can be be configured as `VISIBILITY_CLASSES`.
```
winged: This is the first time you mention nodes. Maybe a little explanation for people who are not familiar with GraphQL vocabulary would be helpful.

I'd also add a short conceptual explanation on how the permission system works:

> The visibility part defines what you can see at all. Anything you cannot see, you're implicitly also not allowed to modify. The visibility classes define what you see depending on your roles, permissions, etc. Building on top of this follow the permission classes (see below) that define what you can do with the data you see.

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+
+Following pre-defined classes are available:
+* `caluma.core.visibilities.Any`: Allow any user without any filtering
+* `caluma.user.visibilities.Authenticated`: Filter results to none for not authenticated users
```
winged: Wording: "Only show data to authenticated users"

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+* `caluma.user.visibilities.CreatedByGroup`: Filter to group node has been created for
```
winged: Requires too much knowledge: Does each node belong to a group? Suggestion:

> Only show data that belongs to the same group as the current user

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+    def filter_queryset_for_form(self, node, queryset, info):
+        return queryset.exclude(slug='protected-form')
+"""
 
-```bash
-docker-compose build --pull
 ```
 
-### Setup pre commit
+Save your visibility module as `visibility.py` and inject it Docker container as volume to path `/app/caluma/extensions/visibility.py`.
```
winged: Maybe add an example docker commandline?

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
-pip install pre-commit
-pip install -r requiements-dev.txt -U
-pre-commit install
+##### Permission classes
+
+Permission classes define who may perform which mutation. Such can be configured as `PERMISSION_CLASSES`.
+
+Following pre-defined classes are available:
+* `caluma.user.permissions.IsAuthenticated`: only allow authenticated users
+* `caluma.core.permissions.AllowAny`: allow any users to perform any mutation.
```
winged: Not a documentation feedback, but: Wouldn't it be useful to have an analog to the visibility class `CreatedByGroup` here? As in "only edit stuff you created" for example?

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+permission class defining per mutation and mutation object what is allowed.
+
+Example:
+```python
+from caluma.form.schema import SaveForm
+from caluma.mutation import Mutation
+
+class CustomPermission(BasePermission):
+    @permission_for(Mutation)
+    def has_permission_default(self, mutation, info):
```
winged: To meaningfully implement my own permission class/method, I need to know how to make my decision, ie. where to get my information. 
Please add some documentation explaining what's in the info parameter (I assume request, user, maybe more?) Also what does the mutation contain? Do I have access to the queryset?

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
 
-You can now access GraphiQL at [http://localhost:8000/graphql](http://localhost:8000/graphql)
+You can now access [GraphiQL](https://github.com/graphql/graphiql) at [http://localhost:8000/graphql](http://localhost:8000/graphql) which includes a schema documentation. The API allows to
+query and mutate form and workflow entities which are described below.
 
-## Development
+### Entities
 
-### Clone
+Caluma is split into two parts, form and workflow. The form can be used without a workflow and visa versa, but the full power of Caluma comes when combining the two.
```
czosel: vice versa

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
 
-To work on Caluma you first need to clone
+Each part consists of different entities, whereas an entity usually comes in pairs, a blueprint and a specific instance.
```
czosel:
```suggestion
Each part is based on several entities, which usually come in pairs: One defines a "blueprint" and is instance-independent, while the other one represents the blueprint in relation to a specific instance. The two entities are usually connected by a one-to-many relationship, e.g. there can be many concrete `cases` (instance-specific) following a certain `workflow` (global blueprint). 
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
+# Contributing
+
+Contributions to Caluma Service are very welcome! Best have a look at the open [issues](https://github.com/projectcaluma/caluma/issues)
+and open a [GitHub pull request](https://github.com/projectcaluma/caluma/compare). See instructions below how to setup development
+environment.
```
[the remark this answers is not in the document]
sliverc: Good point. I have added a more general sentence that this should always be done not just because the project is young.

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+# needed for permission handling
+# only needs to be run once
+echo UID=$UID > .env
+# open shell
+docker-compose run --rm caluma bash
+```
+
+### Testing
+
+Once you have shelled in docker container as described above
```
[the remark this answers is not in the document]
sliverc: Will do

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+black .
+# running tests
+pytest
+# create migrations
+./manage.py makemigrations
+```
+
+### Install new requirements
+
+In case you needed to add new requirements you simply need to build the docker container
```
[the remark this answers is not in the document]
sliverc: Changed.

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+docker-compose build --pull
+```
+
+### Setup pre commit
+
+Pre commit hooks is an additional option instead of executing checks in your editor of choice.
+
+```bash
+# create virtualenv with tool of your choice
```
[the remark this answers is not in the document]
sliverc: Good point

README.md
```diff
@@ -6,9 +6,15 @@
 [![Black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/projectcaluma/caluma)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](https://opensource.org/licenses/MIT)
 
-A collaborative form editing service. See [project site](https://projectcaluma.github.io/) for more details.
+A collaborative form editing service.
```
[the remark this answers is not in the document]
sliverc: I guess this will need to be discussed in a broader context as this is also written this way on caluma.io

README.md
sliverc: :smile:

README.md
sliverc: Integrated.

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
-cd caluma
-```
+#### Form entities
 
-### Open Shell
+**Form** defines the structure of a document, basically a container of questions.
 
-Once it is cloned you can easily open a shell in the docker container to
-open an development environment.
+**Question** defines a single input of a form whereas a question has different implementation types defining how a question should be represented and validated.
```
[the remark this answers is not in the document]
sliverc: integrated.

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+Naming and concept of workflow follows concept of [Workflow Patterns Initiative](http://www.workflowpatterns.com/).
 
-### Install new requirements
+**Workflow** defines a structure of a business process. It uses forward moving flows referencing tasks to model it.
 
-In case you needed to add new requirements you simply need to build the docker container
-again for those to be installed and re-open shell.
+**Task** is a definition of work being part of a workflow.
+
+**Flow** is a definition on what task resp. tasks follows another.
```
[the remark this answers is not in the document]
czosel: A **flow** defines a connection (directed graph?) between two or more tasks.

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+**Question** defines a single input of a form whereas a question has different implementation types defining how a question should be represented and validated.
 
-```bash
-# needed for permission handling
-# only needs to be run once
-echo UID=$UID > .env
-# open shell
-docker-compose run --rm caluma bash
-```
+**Document** is a specific instance of a form being a container of answers.
```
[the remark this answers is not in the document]
sliverc: Integrated.

README.md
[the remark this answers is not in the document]
sliverc: Integrated.

README.md
sliverc: Good point

README.md
```diff
@@ -21,68 +27,147 @@ After installing and configuring those requirements, you can download [docker-co
...
+**Flow** is a definition on what task resp. tasks follows another.
+
+**Case** is a specific instance of a workflow.
+
+**WorkItem** is a specific instance of a task a resource has to work off.
+
+
+### Configuration
+
+Caluma is a [12factor app](https://12factor.net/) so it is configurable per environment variables.
```
czosel:
```suggestion
Caluma is a [12factor app](https://12factor.net/) so it is configured using environment variables.
```

README.md
sliverc: good point. Rewrote it that for authentication an IAM is required. If not available suggesting Keycloak.

README.md
sliverc: well summarized. Integrated.

README.md
sliverc: :smile: integrated.

README.md
sliverc: Sounds good. Integrated.

README.md
sliverc: Good idea. I could actually use this for Fromarte. Will do this in another PR I guess.

README.md
sliverc: Discussed with @czosel and a tutorial style way of explaining should be done in another PR and some later point and would push the boundaries of this initial documentation PR.

README.md
sliverc: Adapted.

README.md
sliverc: Added example to docker-compose.yml file which is used for installation and referenced it.

README.md
sliverc: added some docs for arguments

sliverc:
@winged @kaldras @czosel 
I have integrated above comments please have a look again.
Please keep in mind though that this PR is a first step forward which simply explains how Caluma works in general so we should keep the scope of the PR to this.

In the future there will certainly need to improve this with diagrams, Graphql example queries, step to step tutorial, extracting documentation to e.g. ReadTheDocs etc.

sliverc:
Thanks @winged @czosel 
Totally agree on the structure which needs to be flatten. Let's consider this when thinking about moving documentation to ReadTheDocs or similar.

PR #156, opened 2018-12-14
Add complete workflow form task type

sliverc:
Implements workflow data pattern [wdp7](http://www.workflowpatterns.com/patterns/data/visibility/wdp7.php)


caluma/workflow/migrations/0003_auto_20181214_1317.py
```diff
@@ -8,7 +8,7 @@
 
 class Migration(migrations.Migration):
 
-    dependencies = [("workflow", "0001_initial")]
```
winged: Why the migration reordering? This looks like trouble for existing databases

caluma/form/validators.py
```diff
@@ -0,0 +1,94 @@
...
+
+from . import jexl
+
+
+class AnswerValidator:
+    def _validate_question_text(self, question, value):
+        max_length = (
+            question.max_length if question.max_length is not None else sys.maxsize
+        )
+        if not isinstance(value, str) or len(value) > max_length:
```
winged: Doesn't Django have built-in validators for such cases?
See https://docs.djangoproject.com/en/2.1/ref/validators/#built-in-validators

sliverc: It would be but those validators are per field and do not have access to the full context of the model.

If we wanted to use those we would have to normalize the database to have  one column or actually even table for each value type. Makes the code and database model more complicated so we decided to simplify this on this part and use a `JSONField` for values. Hence this validator is which has access to the question as well.

caluma/workflow/migrations/0003_auto_20181214_1317.py
sliverc: Yep something went wrong when merging the migrations. Will fixed it.

sliverc: Fixed migration issue.

PR #159, opened 2018-12-17
Add answer transform to question jexl

sliverc:
This allows is_required and is_hidden to be dependen on answers of other questions.

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -10,7 +10,14 @@ from ..core.types import DjangoObjectType, Node
 
 
 class QuestionJexl(graphene.String):
-    """Question jexl represents a jexl expression returning boolean."""
+    """Question jexl expression returning boolean.
+
+    Following transform can be used:
+    * answer - access answer of document by question slug
+
+    Example:
+    'question-slug'|answer
+    """
 
     pass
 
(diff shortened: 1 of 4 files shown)


caluma/form/schema.py
```diff
@@ -10,7 +10,14 @@
 
 
 class QuestionJexl(graphene.String):
-    """Question jexl represents a jexl expression returning boolean."""
+    """Question jexl expression returning boolean.
+
+    Following transform can be used:
+    * answer - access answer of document by question slug
+
+    Example:
```
winged: I feel all the JEXL stuff should be documented somewhere as well (ie. where can I use jexl, what transforms/data do I have available, etc).

caluma/form/validators.py
```diff
@@ -81,7 +81,7 @@ def validate(self, *, form, answers, **kwargs):
         answer_by_question = {answer.question.slug: answer.value for answer in answers}
```
winged: The way this is written, you only get access to the answers currently being validated. Is this what you want? Or wouldn't it be better to have all questions of the form available (and only look them up from the QS when needed)?

Obviously this would require some overloading, so you get the "new" answer if available, etc..

sliverc: all answers always get validated when whole document gets validated (reason it could be that configuration of a question has changed after a answer has already given but before document is being completed). Hence what you are describing is already the case.

caluma/form/schema.py
sliverc: I have added a small section on the README.

Small note though: Docstrings like this will actually be automatically extracted into a GraphQL schema documentation which is referenced how to open it on the README already.

sliverc:
@winged 
Ready for review again.

Issue #163, opened 2018-12-27
Add GraphQL example for supported workflow patterns


sliverc:
Depends on #162

PR #164, opened 2018-12-27
Implement workflow pattern distribution by offer

sliverc:
See [wrp13](http://www.workflowpatterns.com/patterns/resource/push/wrp13.php)

* addressGroups jexl on task defines at design time what groups given task
  will be offered to
* addressedGroups on work item defines what groups may process work items
* assignedUsers defines who actually works on work item

Part of documentation is added into structures which already exist. Extended example documentation will follow with #163


caluma/workflow/jexl.py
```diff
@@ -6,6 +6,12 @@
 from ..core.jexl import ExtractTransformSubjectAnalyzer
 
 
+class GroupJexl(JEXL):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.add_transform("groups", lambda spec: spec)
```
winged: This transform doesn't really do anything.. is this correct? Or shouldn't it actually *do* something? (I see that some other transforms do the same, but still..

caluma/workflow/models.py
```diff
@@ -104,7 +106,18 @@ class WorkItem(UUIDModel):
         Task, on_delete=models.DO_NOTHING, related_name="work_items"
     )
     status = models.CharField(choices=STATUS_CHOICE_TUPLE, max_length=50, db_index=True)
-    meta = JSONField(default={})
+    meta = JSONField(default=dict)
+
+    addressed_groups = ArrayField(models.CharField(max_length=150), default=list)
+    """
+    Offer work item to be processed by a group of users, such are not committed to process
```
winged: Wouldn't it make sense to use the `help` field parameter instead for this?

winged: Also, what does "committed" mean here? If I give an item to a group of people, the word "required" would probably be a better match.

caluma/form/schema.py
```diff
@@ -15,8 +15,10 @@ class QuestionJexl(graphene.String):
     Following transform can be used:
     * answer - access answer of document by question slug
 
-    Example:
-    'question-slug'|answer
+    Examples:
+    'answer' == 'question-slug'|answer
+    'answer' in 'list-question-slug'|answer
```
winged: Shouldn't the second line be the plural transform?

caluma/workflow/jexl.py
winged: On another note regarding "task" vs "tasks" (plural) transforms below: I'd suggest validating the input to be an actual list in the plural transform variant(s). Otherwise you (or the user) will get a nasty surprise if the plural-form is used on a singular input (or vice versa)

caluma/form/schema.py
sliverc: there is no plural transform in this jexl.

caluma/workflow/jexl.py
sliverc: Those transforms are used as markers. This way a jexl expression can be analyzed and groups resp. tasks extracted without knowing the input data.

The input type of a transform should be validated I agree. I will add this to the respective validator.

caluma/workflow/models.py
sliverc: Good point with help attribute. I will change this.

required would mean someone has to take care of such work items. But this is not the case it is simply saying whether a user of a group wants to take care. If they want to they can ignore this offer. Hence used address.

caluma/workflow/jexl.py
sliverc: Added additional validation for transform subject types.

caluma/workflow/models.py
sliverc: move comment to `help_text`. Now it is even visible in GraphQL documentation schema :+1:

sliverc:
@winged All conversations are addressed and PR ready for review again

PR #165, opened 2018-12-28
Use docker-compose.yml as full deployment example

sliverc:
docker-compose.yml is already referenced on README.md to be used as an deployment example.

This change makes `docker-compose.yml` a self explaining example.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,12 @@
 version: "3.4"
 services:
   db:
-    image: postgres:9.6
-    ports:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,12 +1,12 @@
 version: "3.4"
 services:
   db:
-    image: postgres:9.6
-    ports:
-      - "5432:5432"
+    image: postgres:9.6-alpine
```
winged: Is there a reason why not newer?

PR #225, opened 2019-01-31
Use enum type for choices arguments

sliverc:
This change makes api more consistent that arguments and query node types use the same uppercase naming. It additionally allows also autocompletion for argument filters.

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -11,6 +11,7 @@ from django.utils import translation
 from django_filters.constants import EMPTY_VALUES
 from django_filters.fields import ChoiceField
 from django_filters.rest_framework import (
+    ChoiceFilter,
     Filter,
     FilterSet,
     MultipleChoiceFilter,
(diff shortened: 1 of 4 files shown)


caluma/core/filters.py
```diff
@@ -263,3 +273,40 @@ def description(self):
...
+
+    name = to_camel_case(field.label)
+    choices = list(get_choices(field.choices))
+    named_choices = [(c[0], c[1]) for c in choices]
+    named_choices_descriptions = {c[0]: c[2] for c in choices}
+
+    class EnumWithDescriptionsType(object):
+        @property
+        def description(self):
+            return named_choices_descriptions[self.name]
```
czosel: Why the inline class definition?

caluma/core/filters.py
```diff
@@ -164,7 +165,16 @@ def __init__(self, label, *args, fields=tuple(), **kwargs):
 
 
 class FilterSet(GrapheneFilterSetMixin, FilterSet):
-    pass
+    @classmethod
+    def filter_for_lookup(cls, field, lookup_type):
+        filter_class, params = super().filter_for_lookup(field, lookup_type)
+        if issubclass(filter_class, ChoiceFilter):
+            meta = field.model._meta
+            # Prefixing newly created filter with Argument to avoid conflicts
```
czosel: Nitpick:
```suggestion
            # Postfixing newly created filter with Argument to avoid conflicts
```

PR #238, opened 2019-02-05
Remove obsolete publish and archive mutations

### LINKED ISSUE #199: Publish and archive mutation should not be separated
Currently publish and archive are seperate mutations. This way it is not possible to unpublish or unarchive.

Initially this was meant as a protection layer, but there is actually no need to restrict Caluma not to unpublish or unarchive e.g. a form as it is the responsibility of the admin to decide whether this is a good idea or not.

Hence the mutations should be removed and instead integrated into the specific mutations as flags (e.g. saveWorkflow will have a isPublished and isArchived input field)

sliverc:
Fixes #199 

API user needs to have the ability to set publish and archive flag
to False so it doesn't make sense to have its own mutation for this.

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -183,12 +183,6 @@ class SaveForm(UserDefinedPrimaryKeyMixin, Mutation):
         serializer_class = serializers.SaveFormSerializer
 
 
-class ArchiveForm(Mutation):
-    class Meta:
-        lookup_input_kwarg = "id"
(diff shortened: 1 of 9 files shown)


sliverc:
This is not backwards compatible and form builder needs to be adjusted, best to use checkboxes instead of buttons to archive and publish.

See also https://github.com/projectcaluma/ember-caluma-form-builder/issues/87

PR #245, opened 2019-02-06
Visibility filters: First function "wins"

czosel:
Before, we we're running _all_ matching visibility functions. Running
only the first (most specific) function allows overriding a generic
`Node` visibility function for specific models.

This is a backwards-incompatible change. Call your `Node`s visibility
function directly if you want to emulate previous behavior.

### DIFF
--- a/caluma/core/tests/test_visibility.py
+++ b/caluma/core/tests/test_visibility.py
@@ -3,7 +3,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db.models import CharField
 
 from .. import models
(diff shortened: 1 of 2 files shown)


sliverc:
CI fails because of linting issues:
```
./caluma/core/tests/test_visibility.py:6:1: I001 isort found an import in the wrong position
./caluma/core/tests/test_visibility.py:70:1: D200 One-line docstring should fit on one line with quotes
./caluma/core/tests/test_visibility.py:70:1: D400 First line should end with a period
```

PR #262, opened 2019-02-08
Ensure that question is added at the end of the form

### LINKED ISSUE #259: Add form question does not add at the end
It would be expected that a question will be added at the end of the list but currently it will add it in order of the slug.

Not as critical as it can still be reordered afterwards but still annoying.

sliverc:
Fixes #259 

This broke when UUIDs were introduced as adding relied on sequential order of primary key.


caluma/form/models.py
```diff
@@ -31,7 +31,7 @@ class FormQuestion(UUIDModel):
     sort = models.PositiveIntegerField(editable=False, db_index=True, default=0)
 
     class Meta:
-        ordering = ("-sort", "id")
+        ordering = ("-sort",)
```
winged: I'd still leave the "id" (or alternatively, "name") as secondary ordering field, so it remains deterministic even if the user/frontend gives the same "sort" value for multiple rows

caluma/form/serializers.py
```diff
@@ -107,7 +120,7 @@ def update(self, instance, validated_data):
                 "Input questions are not the same as current form questions"
             )
 
-        for sort, question in enumerate(reversed(questions)):
+        for sort, question in enumerate(reversed(questions), start=1):
```
winged: Out of curiosity: Why do we `reverse()` the questions here? Why are they not in "sorted" order already?

caluma/form/models.py
sliverc: this is hidden from the user interface. User interface resp. api user doesn't see the sort and this is only handled internally where we make sure those numbers are never the same.

caluma/form/serializers.py
sliverc: same as above sort flag is not exposed to user interface. API user simply sends a list which is ordered and this representation will need to be represented in the database.

sliverc: reverse is used that we do not have to figure out the largest sort id by a query which could be racy but can simply add new questions with sort order 0

caluma/form/models.py
winged: Aah. If it's not even exposed via API, then I'm fine with it, as we can enforce consistency

caluma/form/serializers.py
winged: I understand why you start at 1 (for inserting later), but the reverse I still don't quite get

sliverc: there is a mutation AddFormQuestions which adds a single question at the end of the list, as we use descending sort order this can simply be set to zero and it will be safe.

sliverc: one more thing:
input list comes in ascending order but sort order is in descending. Hence reversed.

winged: Aah okay, thanks for explaining!

PR #265, opened 2019-02-08
Rename question type radio to choice and checkbox to multiple choice

### LINKED ISSUE #215: Rename TYPE_RADIO and TYPE_CHECKBOX
Because in the frontend, we'll want to use `TYPE_RADIO` for dropdown fields as well.

Suggested naming:
EXCLUSIVE_CHOICE
MULTIPLE_CHOICE

sliverc:
Fixes #215 

Radio and checkbox are frontend layout decision terms and should not be
used in the api.

Changing it to more generic term whereas frontend implementation can choose
between a radio or select view but always using the same api question type
choice.


sliverc:
I have chosen choice and multiple choice because all other python libraries we use in Caluma name it exactly this way:

https://django-filter.readthedocs.io/en/master/ref/filters.html#choicefilter
https://www.django-rest-framework.org/api-guide/fields/#choice-selection-fields
https://docs.djangoproject.com/en/2.1/ref/forms/fields/#django.forms.ChoiceField

PR #275, opened 2019-02-12
Validation extension point

### LINKED ISSUE #271: Add validation extension point
Visibility extension point restricts what user can see, whereas permission end point defines what can be done on a specific form.

The validation extension point completes these extensions points where input resp. payload data can be validated or adjusted when need to.

sliverc:
Fixes #271 

This completes visibility and permission extensions so not only can be defined what a user can access and what mutations on what object can be executed but also process input data of any mutation.

This is especially helpful when defining custom meta data per project.


README.md
```diff
@@ -197,14 +199,51 @@ class CustomPermission(BasePermission):
...
 
 Save your permission module as `permissions.py` and inject it as Docker volume to path `/app/caluma/extensions/permissions.py`,
 see [docker-compose.yml](https://github.com/projectcaluma/caluma/blob/master/docker-compose.yml) for an example.
 
 Afterwards you can configure it in `PERMISSION_CLASSES` as `caluma.extensions.permissions.CustomPermission`.
 
+##### Validation classes
+
+Validation classes validate or amend input data of any mutation. It completes permission classes which allow execution of mutations based on updating object whereas validation classes
+process input data.
```
czosel: Validation classes can validate or amend input data of any mutation. Each mutation is processed in two steps:

1. Permission classes check if a given mutation is allowed based on the object being mutated, and
2. Validation classes process (and potentially amend) input data of a given mutation

README.md
```diff
@@ -197,14 +199,51 @@ class CustomPermission(BasePermission):
...
+
+A custom validation class defining validation per mutation can be defined like the following:
```
czosel:
```suggestion
A custom validation class defining validations for various mutations looks like this:
```

README.md
```diff
@@ -197,14 +199,51 @@ class CustomPermission(BasePermission):
...
+    @validation_for(SaveForm)
+    def validate_save_form(self, mutation, data, info):
+        if data['meta'] and info.context.group != 'admin':
+          raise exceptions.ValidationException('May not change meta on form')
+        return data
+```
+
+Arguments:
+* `mutation`: mutation class
+* `data`: input data whereas relation are resolved (e.g. a form ID is represented as actual form object)
```
czosel:
```suggestion
* `data`: input data with resolved relationships (e.g. a form ID is represented as actual form object)
```

caluma/core/validations.py
```diff
@@ -0,0 +1,66 @@
...
+        return _decorated
+
+    return decorate
+
+
+class BaseValidation(object):
+    """Basic validation class to be extended by any validation implementation.
+
+    In combination with the decorators `@validation_for` a custom validation class
+    can define validation on basis of mutations and its subclasses.
```
czosel:
```suggestion
    can define validations both for specific mutations as well as the base `Mutation` class.
```

caluma/core/validations.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Per default it doesn't do any validation but default can be overwritten to define
+    a validation for `Mutation`.
```
czosel: (included in suggestion above)

caluma/core/validations.py
sliverc: Not so sure on this change as it makes it inconsistent with other extension points and basically just repeats what the code already way more simply states.

caluma/core/validations.py
sliverc: ahh I see what you we trying to do with above comment... should review first all comments before recommenting... :smile: let me try to rephrase this

czosel: :smile:

PR #278, opened 2019-02-12
Record closed date, user and group on work item and case

### LINKED ISSUE #277: Introduce closed date and closed by
For statistical reason and to verify whether a work item or case has been processed within a given time frame it is important to register when a work item or case has been closed.

It can be closed either by canceling or completing it. Additional is the user recorded who closed it.

sliverc:
Fixes #277 

A case and work item is defined as closed when it either got completed or canceled. With this additional information the throughput time can be calculated.


sliverc:
@winged Yeah your are right, those fields could need a help_text so it is visible in the GraphQL API documentation. Let me add this.

PR #279, opened 2019-02-12
Calculate deadline of work item based on task lead time 

### LINKED ISSUE #276: Introduce deadline dates
It is common to specify a deadline for a work item in a process model. Hence we should introduce a lead time on a task which will be calculated to a deadline date when work item is created.

Reference http://www.workflowpatterns.com/patterns/exception/exceptiontypes.php

How hooks will be implemented for expiry will need to be considered at a later point.

sliverc:
Fixes #276 

For full flexibility lead time is defined in seconds.


winged:
I find "lead time" as an expression rather unclear. It's not obvious what this is for, when not referencint the PR, the associated issue and the link there.
I know (after googling) that it's the correct technical term, so maybe instead of another name, some more explanatory help text / documentation would be in order.

sliverc:
@winged 
I have adjusted the `help_text` of lead time so it should be understandable by users who do not know the term lead time.

Also in terms of documentation I am not sure whether you have had a look at the GraphiQL interface http://localhost:8000/graphql when Caluma is started already?
A user resp. dev of Caluma service will certainly work with this tool and this nicely guides the user with auto completion and additional hints (like lead_time) what to do.

PR #291, opened 2019-02-20
Support multiple instance pattern

### LINKED ISSUE #290: Support multiple instance pattern
Workflowpatterns defines a pattern [Multiple Instances without a priori Run-Time Knowledge](http://workflowpatterns.com/patterns/control/multiple_instance/wcp15.php) which seems to be a good fit processes where multiple instances (work items) of the same task created and assigned to various groups.

The implementation could be done as follows:

- Add a new Interface `MultipleInstanceTask` that implements `Task` and is implemented by `MultipleInstanceTaskFormTask` (initially needed), `MultipleInstanceSimpleTask` (potentially needed in the future) etc.
- Adapt logic in `completeWorkItem` mutation: If associated task is a `MultipleInstanceTask`, it is only completed when _all_ the associated work items have been completed.
- Add mutation `createWorkItem(Case, MultipleInstanceTask)` which creates a new work item for a given Case and MultipleInstanceTask, provided that the given case has _at least one_ open work item of this task.

czosel:
Fixes #290 

I went for a flag `is_multiple_instance` instead of new task types, because the interface `MultipleInstanceTask` could not be enforced on the `createWorkItem` mutation and didn't add much value on it's own. Instead, it caused quite wordy names like `MultipleInstanceCompleteTaskFormTask` which can by avoided with the flag.


czosel:
@sliverc ready for review

caluma/workflow/tests/snapshots/snap_test_task.py
```diff
@@ -47,6 +47,66 @@
     }
 }
 
+snapshots["test_query_all_tasks[multiple_instance_complete_task_form] 1"] = {
+    "allTasks": {
+        "edges": [
+            {
+                "node": {
+                    "__typename": "MultipleInstanceCompleteTaskFormTask",
```
sliverc: I assume this is a deprecated snapshot? Good to run all tests with --snapshot-update to remove deprecations.

caluma/workflow/tests/test_task.py
```diff
@@ -1,10 +1,9 @@
 import pytest
 
-from .. import models, serializers
+from .. import serializers
 from ...core.tests import extract_serializer_input_fields
 
 
-@pytest.mark.parametrize("task__type", [models.Task.TYPE_SIMPLE])
```
sliverc: why is this removed?

caluma/workflow/tests/test_work_item.py
```diff
@@ -287,6 +334,57 @@ def test_complete_work_item_with_next_multiple_tasks(
...
+            clientMutationId
+          }
+        }
+    """
+
+    inp = {"input": {"id": work_item.pk}}
+    result = schema_executor(query, variables=inp)
+
+    assert not result.errors
+    assert case.work_items.count() == 4
```
sliverc: this line is not quite clear why... better to check for completed (1) and ready work items (3)?

caluma/workflow/validators.py
```diff
@@ -5,15 +5,15 @@
...
-    def _validate_task_complete_task_form(self, task, case, document):
-        DocumentValidator().validate(answers=document.answers, form=case.document.form)
-
     def _validate_task_simple(self, task, case, document):
         pass
 
     def _validate_task_complete_workflow_form(self, task, case, document):
         self._validate_task_complete_task_form(task, case, case.document)
 
+    def _validate_task_complete_task_form(self, task, case, document):
```
sliverc: I guess this change is not needed in the diff

caluma/workflow/tests/snapshots/snap_test_task.py
czosel: Indeed, fixed.

caluma/workflow/tests/test_task.py
czosel: No reason, restored :wink:

caluma/workflow/tests/test_work_item.py
czosel: Got point, done

caluma/workflow/validators.py
czosel: True, reverted.

PR #293, opened 2019-02-20
WorkItem validator: fix reference to task form

czosel:


czosel:
Oops, fixed it! Decided to move the method a bit down because the order seems more logical that way.

PR #298, opened 2019-02-22
Filter workitems by assigned_groups

### LINKED ISSUE #292: Filter workitems by assigned_groups
I'd like to get all work items which have overlapping addressed_groups with the current user's groups. Example:
My groups: A, B
Work items's `addressed_groups` in DB:
1) B, C -> hit
2) C, D -> miss

I'd call the filter just `addressedGroups`.

Possible query:
```graphql
query AllWorkItems {
  allWorkItems (addressedGroups: ["A", "B"]) {
```

See also: https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#overlap

open-dynaMIX:
Fixes #292

This filter uses the `__overlap` lookup expression.

This commit implements a function `_generate_list_filter_class(inner_type)`
to generate Filter classes that will resolve into graphene type lists.

For now we only need a `StringListFilter`.

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -1,5 +1,7 @@
 from functools import reduce
 
+import django.forms
+import graphene
 from django import forms
 from django.contrib.postgres.fields import JSONField
 from django.contrib.postgres.fields.hstore import KeyTransform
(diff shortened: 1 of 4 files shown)


czosel:
CI probably just needs a `pytest -k test_schema --snapshot-update` to go green

PR #317, opened 2019-03-08
Do not exclude documents in form schema

open-dynaMIX:
This change allows to have access to documents and even cases from
forms.

This is useful for triggering cases based on form properties (e.g.
meta).

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -176,7 +176,7 @@ class Form(DjangoObjectType):
     class Meta:
         model = models.Form
         interfaces = (relay.Node,)
-        exclude_fields = ("documents", "workflows", "tasks")
+        exclude_fields = ("workflows", "tasks")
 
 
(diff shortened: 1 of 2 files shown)


open-dynaMIX:
@winged I know what you mean, but honestly am a bit lost as to how to test this.

Any inputs would be very much appreciated.

winged:
> @winged I know what you mean, but honestly am a bit lost as to how to test this.
> 
> Any inputs would be very much appreciated.

I don't think we explicitly need to test this. It's just a question on how graphene applies filters. To test this, you'd need an elaborate permissions setup I think

open-dynaMIX:
@winged This seems to be handled in the VisibilityClasses, as can be seen [here](https://git.adfinis-sygroup.ch/fromarte/fromarte/blob/master/caluma/extensions/visibilities.py).

So I'd say, there are no security implications.

czosel:
Yep, that's also what I was thinking :+1:

PR #318, opened 2019-03-08
Implement meta-field key filtering

open-dynaMIX:
This commit implements a "meta_HasKey"-filter.

For this to work, the meta-field is now a "generic.GenericScalar()".

This implementation is up for discussion and only implemented for Form.meta atm. If we decide that this is the way to go, we can implement it for the other models as well.

### DIFF
--- a/caluma/form/filters.py
+++ b/caluma/form/filters.py
@@ -1,14 +1,14 @@
 from . import models
 from ..core.filters import (
-    FilterSet,
     GlobalIDFilter,
     GlobalIDMultipleChoiceFilter,
+    MetaFilterSet,
     OrderingFilter,
     SearchFilter,
 )
 
 
-class FormFilterSet(FilterSet):
+class FormFilterSet(MetaFilterSet):
     search = SearchFilter(fields=("slug", "name", "description"))
     order_by = OrderingFilter(label="FormOrdering", fields=("name",))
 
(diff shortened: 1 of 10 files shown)


caluma/form/filters.py
```diff
@@ -11,6 +12,7 @@
 class FormFilterSet(FilterSet):
     search = SearchFilter(fields=("slug", "name", "description"))
     order_by = OrderingFilter(label="FormOrdering", fields=("name",))
+    meta__has_key = CharFilter(lookup_expr="has_key", field_name="meta")
```
winged: I would get rid of the double underscore here, as it leaks out to the graphQL representation, and is derived from an implementation (django queryset filter syntax).

open-dynaMIX: Done.

open-dynaMIX:
Do we want this filter for all models with a meta-field?

caluma/form/filters.py
czosel: I agree, `meta_HasKey` looks quite funny on the API. This feature will always be restricted to `meta` fields, so some an alternative name might be `metaHasKey`.

I'm wondering though: Can we imagine a way to filter for things like "meta property X equals/is bigger than/is smaller than Y"? Potentially, for multiple properties at once? Maybe something like

`filterMeta: '{"X": { "operator": "equals", "value": "Y" }, "Z": {...}}'

czosel:
I think this can be useful everywhere, yeah.

PR #322, opened 2019-03-12
Fix unlinking last document in table answer

### LINKED ISSUE #313: Last row in table answer can't be unlinked
When I try unlinking the last row of a TableAnswer with a query like this
```gql
mutation SaveDocumentTableAnswer {
  saveDocumentTableAnswer(input:{
    question: "table-question-1",
    document: "RG9jdW1lbnQ6MzE0ZWE4MjItZTA0My00OTMzLWJjODItYjRmMTJlYzk0ZGRm",
    value: [] # <---- empty value list
  }) {
    clientMutationId
  }
}
```
an error is returned:
```
{
  "errors": [
    {
      "message": "'ManyRelatedManager' object is not iterable",
      "locations": [
        {
          "line": 42,
          "column": 3
        }
      ],
      "path": [
        "saveDocumentTableAnswer"
      ]
    }
  ],
  "data": {
    "saveDocumentTableAnswer": null
  }
}
```

czosel:
When an empty document list is passed to the validate function, a
ManyRelatedManager object was used as fallback for validation instead of the actual
documents queryset.

Fixes #313

### DIFF
--- a/caluma/form/serializers.py
+++ b/caluma/form/serializers.py
@@ -416,7 +416,10 @@ class SaveDocumentTableAnswerSerializer(SaveAnswerSerializer):
 
     def validate(self, data):
         documents = (
-            data.get("documents") or self.instance and self.instance.documents or []
+            data.get("documents")
+            or self.instance
+            and self.instance.documents.all()
+            or []
         )
         question = data.get("question") or self.instance and self.instance.question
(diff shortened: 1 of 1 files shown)


czosel:
Ah, that's good to know :wink:

PR #332, opened 2019-03-19
Add removeDocument mutation

### LINKED ISSUE #314: Add RemoveDocument mutation
For TableAnswers, each row is represented by a document. When a document is removed in the user interface, all we can currently do is unlink it from the answer, but the actual document is kept in the database. I would prefer to be able to remove these documents.

I wouldn't try limiting the possibility to remove documents to those referenced in TableAnswers, because removing documents in general seems to be a useful feature.

czosel:
Fixes #314

### DIFF
--- a/caluma/form/serializers.py
+++ b/caluma/form/serializers.py
@@ -644,3 +644,24 @@ class RemoveAnswerSerializer(serializers.ModelSerializer):
     class Meta:
         fields = ("answer",)
         model = models.Answer
+
+
+class RemoveDocumentSerializer(serializers.ModelSerializer):
+    document = serializers.GlobalIDField(source="id")
+
+    def update(self, instance, validated_data):
+        if hasattr(instance, "case"):
+            raise Exception("You cannot remove a Document, if it's attached to a case.")
+
+        for answer in instance.answers.filter(
+            question__type=models.Question.TYPE_TABLE
+        ):
+            answer.documents.all().delete()
+
+        instance.delete()
+
+        return instance
+
+    class Meta:
+        fields = ("document",)
+        model = models.Document
--- a/caluma/form/tests/test_document.py
+++ b/caluma/form/tests/test_document.py
@@ -4,7 +4,7 @@ from graphql_relay import to_global_id
 from ...core.relay import extract_global_id
 from ...core.tests import extract_serializer_input_fields
 from ...core.visibilities import BaseVisibility, filter_queryset_for
-from ...form.models import Answer, Question
+from ...form.models import Answer, Document, Question
 from ...form.schema import Document as DocumentNodeType
 from .. import serializers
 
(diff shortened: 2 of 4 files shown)


sliverc:
Not so sure this is a good idea as with this mutation a case may never be completed when a case document is removed.

czosel:
@sliverc Good point. What do you think, shall we only allow deleting Documents which aren't referenced by any (running?) case?

czosel:
As discussed with @sliverc: check cascading (protected)

caluma/form/serializers.py
```diff
@@ -508,3 +508,15 @@ def update(self, instance, validated_data):
 
     class Meta(SaveAnswerSerializer.Meta):
         pass
+
+
+class RemoveDocumentSerializer(serializers.ModelSerializer):
+    document = serializers.GlobalIDField(source="id")
+
+    def update(self, instance, validated_data):
+        models.Document.objects.get(pk=instance.pk)
```
sliverc: This should delete but doesn't do anything at this point.

czosel:
After discussing with @kaldras we noticed that this should probably recursively delete all child documents.

open-dynaMIX:
@czosel This PR is ready for review. I can't set you as a reviewer, because you've opened it.

caluma/form/tests/test_document.py
```diff
@@ -875,3 +875,33 @@ def filter_queryset_for_document(self, node, queryset, info):
         assert result.data["documentValidity"] is None
     else:
         assert len(result.data["documentValidity"]["edges"]) == 1
+
+
+@pytest.mark.parametrize("remove_case", [True, False])
+def test_remove_document(db, remove_case, document, answer, case, schema_executor):
```
winged: This parametrisation is so invasive, I figure it'd be cleaner to have two separate tests for this

caluma/form/serializers.py
```diff
@@ -644,3 +644,19 @@ def update(self, instance, validated_data):
...
+
+
+class RemoveDocumentSerializer(serializers.ModelSerializer):
+    document = serializers.GlobalIDField(source="id")
+
+    def update(self, instance, validated_data):
+        if hasattr(instance, "case"):
+            raise Exception("You cannot remove a Document, if it's attached to a case.")
+
+        instance.delete()
```
winged: From the discussion - shouldn't we recursively delete subdocuments if the document contains table answers?

If it makes sense, maybe parametrize the mutation so the client can select what to do in that case

czosel: Good catch! To me it seems most consistent to also remove subdocuments.

caluma/form/tests/test_document.py
open-dynaMIX: Done.

caluma/form/serializers.py
open-dynaMIX: It's implemented. Without a parameter for now.

winged: Thanks! However this should probably be recursive. I know nobody does nested tables at the moment, but one day people might do it and get a surprising behaviour..

open-dynaMIX: AFAIK the consensus is to NOT support nested tables. @czosel

winged: Fair enough, let's just not forget that :)

czosel: Right - I think we should discuss this in person in more detail once. On the one hand we consciously decided to avoid the complexity involved in nested tables for the moment (both front- and backend), on the other hand we do have a (non-caluma) project where is is currently a requirement. It might be worth discussing what kind of alternative designs might be possible, so that we're prepared for the day where this becomes important :wink:

PR #348, opened 2019-03-27
Add "has_answer" filter

winged:


winged:
RFC: Where else is the AnswerValueFilter needed? Proposal on how to find answers in FormQuestion "subforms"?

czosel:
> Where else is the AnswerValueFilter needed?

`Case` and `Document` seem like a good place to start to me, can't think of other entities right now.

> Proposal on how to find answers in FormQuestion "subforms"?

How about nesting the filters like this?
```
(allDocuments: "{"question_slug": {"subquestion_slug": "value"}")
```
Unfortunately, we'll need this feature pretty much from the start :confused:

caluma/core/tests/test_visibility.py
```diff
@@ -165,3 +166,78 @@ class ConfiguredUnion(Union):
...
+        @filter_queryset_for(Form)
+        def filter_queryset_for_form(self, node, queryset, info):
+            # Extract our "custom" filter value
+            # TODO: can we do this in a more sane, straightforward kind of way?
+            args = [
+                arg.value.value
+                for sel in info.operation.selection_set.selections
+                for arg in sel.arguments
+                if arg.name.value == "custom"
+            ]
```
czosel: Oliver suggested passing the filters as an additional to `filter_queryset`. Looking into the code a little bit, I noticed that `filter_queryset` is currently called from `get_queryset` on the base `Node` object (see core/types.py:13) - and my impression that on that level there won't be much more context about filters available. It might be possible to get that information by extending `graphene.Connection`?

/cc @sliverc

sliverc: Just been thinking about this.

visibility extension is properly not the best idea to do this as it is not only used for reading but also as a lookup for mutations. When using as lookup there are not input arguments but just data.

Also are we not coupling those services a bit too tightly? Could this problem not be solved in the frontend?

If a filter (e.g. status) is set in the frontend, frontend calls 3rd party api first to get all instances which then be passed on to caluma api.

The visibility extensions will still need to implement the security part what may be read of course.

The only draw back would be performance but I guess this would need to be tested whether it is an issue before doing engineering in caluma I guess.

Just an idea to think about.

sliverc: Also as an idea for the answer filter.

hasAnswer is I guess not what we want in the end (same counts for hasMetaKey) but rather filtering for an actual answer.

Have we already checked whether it is possible to use a complex type (e.g. typed key, value) as an argument? That would be more Graphql like but otherwise a jsonString would be a workaround (even when a bit ugly).

czosel: @winged I didn't even consider introducing a complex type for the filters, guess I also still have to get used to thinking in GraphQL more. Seems to be the best solution to me, even though it'll probably be some effort to implement.

```
interface AnswerFilter {
  slug: String!
  operator: eq | lt | gt | ?
}

type StringAnswerFilter {
  value: String!
}

type IntegerAnswerFilter {
  ...
}
```
With this, we could filter by `AnswerFilter[]`.

Initially, we might get away with just implementing the `eq` operator. But in the future we'll need `lt` and `gt`, especially for `DateAnswer`.

winged: Agree, this would be much cleaner. I'm still reading up on graphene/graphql, so did a first throw of an implementation "the way I knew how". will look into a proper typed interface this week.

winged: @sliverc we discussed offline for a long time regarding how to solve this. Problem is we have data in the old software that we may need sorting/filtering/paging on, and data in Caluma. Depending on which column is searched, the frontend would need to decide which service to ask first. Also, we'd still need to ask the other service so we know what we're allowed to show.. 

Instead of the "custom" no-op filter, the alternative is passing data via an additional HTTP header. I kind of gravitate towards that solution now

winged:
I've spent some time trying to make the "has_answer" filter typed. I don't think that's easily done, but maybe I just don't know exactly how :)

There's some more code lying around in #365 where I tried to make it typed, but didn't work yet, so we'll live with this for the time being

sliverc:
@winged Didn't we implement this now already in a different way and this can be closed?

winged:
Yes this was implemented in  #430

PR #349, opened 2019-03-31
Add mutation to create document families

czosel:
When a form contains FormQuestions, those _always_ need to have a sub
document attached - otherwise, the form can never be complete.
The creation of subforms could be done in the frontend (in fact that was
my first approach), but that is tedious, requires many requests, and
also feels somewhat redundant.

This adds a "CreateDocumentFamily" mutation that recursively creates
subdocuments for all FormQuestions and attaches them to the newly
created main document as answers.


czosel:
RFC: Should we really introduce a new mutation for this, or just override `create` of `SaveDocument`? The only use-case that would not be possible anymore if we decide to override `SaveDocument` is saving special `meta` data for sub-documents.

czosel:
I think I'm leaning more towards just changing `SaveDocument`, because saving `meta` on sub-documents currently seems like an unlikely use-case.

czosel:
Since @kaldras also agreed to use `SaveDocument`, i made the change. Ready for review! :tada:

caluma/form/serializers.py
```diff
@@ -390,7 +390,34 @@ class Meta:
         model = models.Option
 
 
+def create_and_link_child_documents(form, document):
```
winged: Nitpick - I'd probably move this to a method within `DocumentSerializer`, as it's not used anywhere else. Unless you expect this to happen in the near future. But then, it may need a more "general" location

czosel: Good point! Fixed.

PR #365, opened 2019-04-09
Braindump: add filters typed

winged:
This does not work and is NOT intended to be merged, more like a "letter of intent":

The "has_answer" filter from #348 should be typed, so we can match arbitrary question types, in a clean manner.

This PR just contains some code I've been messing with - probably I was on the wrong path all together..


sliverc:
@winged Is this brain dump still needed or has this actually been implemented?

winged:
Nope, it can go. See #430

PR #370, opened 2019-04-11
Bugfix: When creating a document tree, set the family

winged:
The family field represents the identifier of the topmost document of
the structure. When recursively creating a document structure, we need
to pass down the family, otherwise it will be set anew for each sub
document, defeating the purpose.

### DIFF
--- a/caluma/form/tests/test_document.py
+++ b/caluma/form/tests/test_document.py
@@ -3,7 +3,7 @@ from graphql_relay import to_global_id
 
 from ...core.relay import extract_global_id
 from ...core.tests import extract_serializer_input_fields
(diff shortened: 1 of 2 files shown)


caluma/form/tests/test_document.py
```diff
@@ -598,5 +599,26 @@ def test_create_document_with_children(
     sub_document = result.data["saveDocument"]["document"]["answers"]["edges"][0][
         "node"
     ]
+
+    def _uuid_from_id(id_string):
+        return base64.b64decode(id_string).decode("utf-8").split(":")
```
czosel: use `extract_global_id` from `core/relay.py` instead

PR #373, opened 2019-04-14
Make question and answer factories dynamic (type-dependent)

czosel:


czosel:
Definitely. Let's do that in another PR though :wink:

PR #377, opened 2019-04-15
Add data_sources for dynamic options in select fields

### LINKED ISSUE #320: Dynamic options in select fields
Sometimes, a dropdown field in a form needs to show dynamic content, e.g.

* All groups in the system
* All users in the system
* A reference to a already existing case ID in caluma ("linking")

-> The data for the options can come from internal or external data sources / APIs. Maybe we can consider "registering" custom option providers in Caluma as an extension point. These providers can then, using the full power of python,

* make API calls and parse responses
* access data inside the caluma DB
* something else we can't imagine yet :wink:

open-dynaMIX:
This commit implements the `data_sources` extension point and the new
question-types `DynamicChoiceQuestion` and
`DynamicMultipleChoiceQuestion`.

There are two queries:
1. allDataSources --> Get all configured sources
2. dataSource --> get data from source

Closes #320


caluma/data_source/data_source_handlers.py
```diff
@@ -0,0 +1,68 @@
...
+class DataSourceException(Exception):
+    pass
+
+
+def is_iterable_and_no_string(value):
+    try:
+        iter(value)
+        if isinstance(value, str):
+            return False
+        return True
```
czosel: Nitpick: How about
```python
return not isinstance(value, str)
```

caluma/data_source/data_source_handlers.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Data:
+    def __init__(self, data):
+        self.data = data
+        self.option, self.value = self.load()
+
+    def load(self):
+        if is_iterable_and_no_string(self.data):
+            if not 0 < len(self.data) < 3:
```
czosel: Why `< 3` ?

caluma/data_source/data_source_handlers.py
```diff
@@ -0,0 +1,68 @@
...
+        return str(self.data), str(self.data)
+
+
+def get_data_sources():
+    sources = []
+    for name in dir(data_sources):
+        cls = getattr(data_sources, name)
+        if (
+            isinstance(cls, type)
+            and name.endswith("DataSource")
```
czosel: Shall we still explicitly mention in the docs that the datasources have to end with "DataSource" in order to be picked up?

czosel: Additionally we could also add a note on this to the "No data_source found" error message.

winged: Hmm I find this a rather strange convention that's not used in any other place in Django. Sure we need this? I mean requiring the name to end with a given suffix to be picked up. Could lead to surprising results and/or support cases

caluma/data_source/data_source_handlers.py
```diff
@@ -0,0 +1,68 @@
...
+    key = f"data_source_{name}_{info.context.user.username}"
+    raw_data = cache_handler(
+        ds.get_data, timedelta(seconds=ds.timeout), ds.default, key, info=info
+    )
+    if not is_iterable_and_no_string(raw_data):
+        raise DataSourceException(f"Failed to parse data from source: {name}")
+
+    data = [Data(d) for d in raw_data]
+
+    return data
```
czosel: No need for `data` variable

caluma/data_source/data_sources.py
```diff
@@ -0,0 +1,47 @@
...
+    A custom data source class could look like this:
+    ```
+    >>> from caluma.core.data_sources import BaseDataSource
+    ... import requests
+    ...
+    ...
+    ... class CustomDataSource(BaseDataSource):
+    ...     info = 'User choices from "someapi"'
+    ...     timeout = 3600
+    ...     default = []
```
czosel: I'd drop this line since `[]` is the default anyway

caluma/data_source/schema.py
```diff
@@ -0,0 +1,35 @@
+from graphene import Connection, ConnectionField, String
+from graphene.types import ObjectType
+
+from .data_source_handlers import get_data_source_data, get_data_sources
+
+
+class DataSource(ObjectType):
+    info = String()
```
czosel: Since `name` is technical, should we use this field for display in the form builder UI? This might need to be multilingual then.

README.md
```diff
@@ -271,6 +271,34 @@ will return a presigned `uploadUrl`, which the client can use to upload the file
...
+## Custom data sources
+For Choice- and MultipleChoiceQuestions it's sometimes necessary to populate the choices
+with calculated data or data from external sources.
+
+For this you can use the data_source extension point.
+
+An example data_source looks like this:
+
+ ```python
+from caluma.core.data_sources import BaseDataSource
```
Yelinz: Isn't the import `caluma.data_source.data_sources`?

caluma/form/serializers.py
```diff
@@ -258,6 +260,50 @@ def update(self, instance, validated_data):
         return instance
 
 
+class SaveDynamicQuestionOptionsMixin(object):
```
Yelinz: Is this intended and what might the use cases be?

czosel: Just to provide some context: We were wondering if the dynamic question types really should support additional, statically configured options.

caluma/data_source/data_source_handlers.py
open-dynaMIX: Absolutely, will do!

README.md
open-dynaMIX: Nope, ` caluma.core.data_sources.BaseDataSource` is the BaseClass you have to subclass in `caluma.data_source.data_sources`.

czosel: @Yelinz already started on the frontend integration for this :wink:

caluma/data_source/data_source_handlers.py
open-dynaMIX: Absolutely! I can't say why, but I implemented auto-discovery for data_sources.

This is fixed now.

README.md
open-dynaMIX: > @Yelinz already started on the frontend integration for this

Great! It's fixed now.

caluma/form/serializers.py
open-dynaMIX: Ha! I really needed this couple of days off, things look much clearer now ;)

The options are now fetched every time the question is accessed. No options are created for those questions in the DB.

caluma/data_source/data_sources.py
open-dynaMIX: Done!

caluma/data_source/schema.py
open-dynaMIX: Done. Also the labels for the options is now multilingual.

PR #379, opened 2019-04-16
Add "placeholder" and "infoText" properties to Question

### LINKED ISSUE #364: Add help text field to Question
Naming?  "info" / "hint" / "description" / "help"

* multilingual
* on every question type

czosel:
Fixes #364


caluma/form/models.py
```diff
@@ -68,6 +68,8 @@ class Question(SlugModel):
     is_required = models.TextField(default="false")
     is_hidden = models.TextField(default="false")
     is_archived = models.BooleanField(default=False)
+    placeholder = LocalizedField(blank=True, null=True, required=False)
```
kaldras: `placeholder` should not be field on the question itself, it should be a dynamic property like, for example `max_length`, right? Or why did you make this an own field?

czosel: Because I need the placeholder to be localized, which wouldn't work with the JSONField. On the API the placeholder only shows up with the questions that can have one.

PR #380, opened 2019-04-17
Add hide_label field to choice questions

### LINKED ISSUE #368: Allow hiding the label of ChoiceQuestions and MultipleChoiceQuestions
Often, typically at the end of a form, there are one or more questions that ask the user to confirm something (terms of service, privacy policy, ...). Those questions usually consist of a single checkbox and a label, e.g.:

 I have read and understood the terms of service
The current MultipleChoiceQuestion doesn't support this use-case very well: It requires a global question label which will be displayed on top.

The easiest way to bring this into the caluma form would be a "Hide label" option for Choice and MultipleChoiceQuestions. Then it is possible to add add a "Confirm" type of question as a MultipleChoiceQuestion with a single option.

Backend tasks:

- [ ] add `hideLabel` Boolean field to ChoiceQuestion and MultipleChoiceQuestion

Yelinz:
Closes #368

### DIFF
--- a/caluma/form/models.py
+++ b/caluma/form/models.py
@@ -120,6 +120,14 @@ class Question(SlugModel):
     def min_value(self, value):
         self.configuration["min_value"] = value
 
+    @property
+    def hide_label(self):
+        return self.configuration.get("hide_label")
+
+    @min_value.setter
+    def hide_label(self, value):
+        self.configuration["hide_label"] = value
+
 
 class QuestionOption(UUIDModel):
     question = models.ForeignKey("Question", on_delete=models.CASCADE)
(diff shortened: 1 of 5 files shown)


caluma/form/models.py
```diff
@@ -120,6 +120,14 @@ def min_value(self):
     def min_value(self, value):
         self.configuration["min_value"] = value
 
+    @property
+    def hide_label(self):
+        return self.configuration.get("hide_label")
+
+    @min_value.setter
```
open-dynaMIX: Copy/paste mistake? :)

czosel:
Did you also consider just using `meta` for this? The backend doesn't really care about this feature, does it?

Yelinz:
No, but you're correct. I will change it to meta.

czosel:
I think I discussed this with @kaldras, but I don't remember the conclusion :sweat_smile: If you go for `meta`, no backend change is required.

Yelinz:
Yes, I will close this PR for now, we can reopen it if the need arises.

PR #393, opened 2019-04-26
New filter on MetaFilterSet: metaValue

winged:
This new filter is able to filter any node that has a meta field by the
meta field's value, as follows:

    query foo {
      allDocuments (metaValue: {key: "foo", value: "hello world"}) {
        ...
        }
    }

This will return all documents whose meta field has a key named "foo"
with the value "hello world".
An optional parameter can be given to the filter to configure the lookup
type:

    metaValue: {key: "foo", value: "hello", lookup:STARTSWITH}

This would change the filter to startswith (prefix search). Also
available lookup methods are EXACT (default if parameter is left out,
and CONTAINS (which does a case-insensitive lookup)).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ target/
 
 # IDE specific configurations
 .idea/
+.vscode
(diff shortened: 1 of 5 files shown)


czosel:
@winged Can you fix the conflict?

PR #396, opened 2019-04-28
feat(form): add question for static content

czosel:
This adds a question type that can be used for inserting static content
like subheadings, additional text, images, videos, links etc. inside
forms (it will be rendered as markdown in the frontend).
Since it is not an actual question, no corresponding answer type
is required.


caluma/form/schema.py
```diff
@@ -235,6 +241,8 @@ class Meta:
             "row_form",
             "sub_form",
             "placeholder",
+            "static_content",
+            "static_content",
```
winged: Should only be listed once

czosel: Nice find, fixed!

PR #398, opened 2019-04-29
feat(jexl): add support for nested forms

czosel:
See https://github.com/projectcaluma/ember-caluma/pull/154 for the corresponding frontend implementation.

### DIFF
--- a/caluma/form/jexl.py
+++ b/caluma/form/jexl.py
@@ -19,15 +19,12 @@ class QuestionJexl(JEXL):
         self.add_transform("answer", self.answer_transform)
         self.add_transform("mapby", lambda arr, key: [obj[key] for obj in arr])
 
-    def answer_transform(self, *args):
-        question = args[0]
-        path = args[1] if len(args) > 1 else None
-
+    def answer_transform(self, question_with_path):
         current_context = self.context
-        if path:
-            parts = path.split(".")
-            for part in parts:
-                current_context = current_context.get(part)
+        segments = question_with_path.split(".")
+        question = segments.pop()
+        for segment in segments:
+            current_context = current_context.get(segment)
 
         return current_context.get(question)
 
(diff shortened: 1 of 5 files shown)


caluma/form/jexl.py
```diff
@@ -15,8 +15,21 @@ class QuestionJexl(JEXL):
     def __init__(self, answer_by_question={}, **kwargs):
         super().__init__(**kwargs)
 
-        self.add_transform("answer", lambda question: answer_by_question.get(question))
+        self.context = answer_by_question
+        self.add_transform("answer", self.answer_transform)
         self.add_transform("mapby", lambda arr, key: [obj[key] for obj in arr])
 
+    def answer_transform(self, *args):
```
winged: I'd prefer explicit arguments - they have specific meanings anyway, judging from the code.

You can do `def answer_transform(self, question, path=None):` to get the same effect

caluma/form/tests/test_jexl.py
```diff
@@ -19,3 +19,22 @@
 def test_question_jexl_validate(expression, num_errors):
     jexl = QuestionJexl()
     assert len(list(jexl.validate(expression))) == num_errors
+
+
+@pytest.mark.parametrize(
+    "expression,result",
+    [
+        ('"a1"|answer', "A1"),
```
winged: That interface feels awkward. Why not unify the path with the question slug? Something like `"parent.form_b.b1"|answer`

caluma/form/jexl.py
czosel: Right, that's what I was looking for! My python is a bit rusty :blush:

caluma/form/tests/test_jexl.py
czosel: Good point. 
/cc @fkm @kaldras

kaldras: Yeah, I like this. Feels simpler and does the same!

czosel: Fixed!

caluma/form/jexl.py
czosel: Fixed

PR #400, opened 2019-04-29
fix(API): use GenericScalar for meta fields

czosel:
BREAKING CHANGE: If you've been manually parsing the JSONString of a
meta field, you can remove that code - meta is automatically expanded now.

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -80,6 +80,8 @@ class Question(Node, graphene.Interface):
 
 
 class Option(DjangoObjectType):
+    meta = generic.GenericScalar()
+
     class Meta:
         model = models.Option
         interfaces = (relay.Node,)
@@ -619,6 +621,7 @@ class Document(DjangoObjectType):
     answers = DjangoFilterSetConnectionField(
         AnswerConnection, filterset_class=filters.AnswerFilterSet
     )
+    meta = generic.GenericScalar()
 
     class Meta:
         model = models.Document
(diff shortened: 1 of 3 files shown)


winged:
What's the rationale behind this change? Why is it necessary?

czosel:
We're expanding `meta` on several entities already (`GenericScalar` instead of `JSONString`). I think we should do that everywhere, since it makes it easier for the frontend and comes with no disadvantage.

winged:
> We're expanding `meta` on several entities already (`GenericScalar` instead of `JSONString`). I think we should do that everywhere, since it makes it easier for the frontend and comes with no disadvantage.

In that case, why not change it on all entities?

czosel:
I was afraid you'd say that :smile:

czosel:
@winged done. Do you have an opinion about whether meta should be required?

winged:
> @winged done. Do you have an opinion about whether meta should be required?

Meta should definitely not be required. It's only used for certain purposes, so if you're not using it, you shouldn't need to set it to some value

winged:
Hmm I've just seen that `Question.meta` is required, while all others are not. Any reason for that?

czosel:
No idea! Some of the `meta` field in the workflow schema are defined as `meta = JSONField(default=dict)` - that might also be reasonable?

/cc @open-dynaMIX

winged:
I don't think the default has any use, really. It's a suggestion to use it as a dict, but it's not really required to do so (except that then filters like the one from #393 won't work in that situation). I'd rather leave it NULL instead if not used (saves a wee bit of storage as well)

czosel:
Ok. Shall we move on with this then?

czosel:
@winged Wanna approve it then? :wink:

PR #405, opened 2019-04-29
Refactor of caching for DataSources

open-dynaMIX:
This commit implements a simple caching decorator, that allows for
caching based on DataSource name.

This replaces the `caluma.data_source.utils.cache_handler()`.


README.md
```diff
@@ -298,8 +300,8 @@ This class needs also to be added to the `DATA_SOURCE_CLASSES` environment varia
 ### Properties
 
 * `info`: Descriptive text for the data source (can also be a multilingual dict)
-* `timeout`: Time you want to cache the data
-* `default`: The default value to be returned if execution of `get_data()` fails
```
czosel: Why is this removed? As far as I can see, `timeout` is still supported?

open-dynaMIX: `timeout` is now set in the decorator and not as a property of the DataSource class anymore.

I did not mention the timeout arg in the README because I think in the context of a caching decorator, the word "timeout" pretty much says it all.

czosel: Alright :+1:

caluma/data_source/data_sources.py
```diff
@@ -37,11 +43,23 @@ class BaseDataSource:
...
-    default = []
+    default = None
 
     def __init__(self):
         pass
 
     def get_data(self, info):  # pragma: no cover
         raise NotImplementedError()
+
+    def __call__(self, info):
```
winged: I'm not a fan of unnecessary use of magic functions. Why not name it `try_get_data_with_fallback()` or similar?

README.md
```diff
@@ -309,6 +311,13 @@ this value will also be used as label.
 
 For the label, it's possible to use a dict with translated values.
 
+### `data_source_cache` decorator
+This decorator allows for caching the data based on the DataSource name.
+
+Django's cache framework is used, so you can also implement your own caching logic. When
+doing so it's, advisable to use the `data_source_` prefix for the key in order to avoid
```
winged:
```suggestion
doing so, it is advisable to use the `data_source_` prefix for the key in order to avoid
```

caluma/data_source/data_sources.py
open-dynaMIX: I share your concern regarding magic functions. The reason I went with that is, that we don't want to reserve method-names. I guess if we go with `try_get_data_with_fallback()` the risk is very small.

czosel: In JS, I'd prefix the method with an underscore. Does a similar convention exist in python?

open-dynaMIX: Yes, but then it's considered a "private"-method and should not be part of the API.

PR #407, opened 2019-04-30
fix(validation): ignore hidden required fields

czosel:


czosel:
For now that's guaranteed, yeah. For required machine-filled hidden fields, there is also no way for the user to correct that issue; so if we have that use-case in the futurem, we'll have to think about that.

PR #410, opened 2019-05-01
fix(validation): add type mapping for validating file questions

czosel:


czosel:
@open-dynaMIX coverage should be good now.

caluma/form/serializers.py
```diff
@@ -681,11 +681,8 @@ def create(self, validated_data):
         return super().create(validated_data)
 
     def update(self, instance, validated_data):
-        if instance.file and instance.file.name is not validated_data["file"]:
-            instance.file.delete()
-            validated_data = self.set_file(validated_data)
-        elif not instance.file:
-            validated_data = self.set_file(validated_data)
+        instance.file.delete()
+        validated_data = self.set_file(validated_data)
```
open-dynaMIX: I think this was a bit too much. We don't want to recreate the File-object if we just want to modify the `meta`-field on the Answer. Or any other field besides `file` for that matter.

I would change this to:

```python
def update(self, instance, validated_data):
    if instance.file.name is not validated_data["file"]:
        instance.file.delete()
        validated_data = self.set_file(validated_data)
    return super().update(instance, validated_data)
```

So only if the file name changed, we recreate the File-object. For replacing the file with another one with the same name, there is no mutation needed, as the frontend just can fetch a new `uploadUrl`.

czosel: Oops, you're right. Do you want to create the PR?

open-dynaMIX: Done!

PR #413, opened 2019-05-01
Add validators for dynamic choice options questions

### LINKED ISSUE #412: Dynamic choice options questions are missing validators
The Questions DynamicChoiceQuestion and DynamicMultipleChoiceQuestion are missing answer validators in `caluma/form/validators.py`

Yelinz:
This PR adds validators for answers to DynamicChoiceQuestion and
DynamicMultipleChoiceQuestion.

For this to work, all validators need access to the `info`-object.

Closes #412

### DIFF
--- a/caluma/form/validators.py
+++ b/caluma/form/validators.py
@@ -4,6 +4,7 @@ from rest_framework import exceptions
 
 from . import jexl
 from .models import Question
+from caluma.data_source.data_source_handlers import get_data_source_data
 
 
 class AnswerValidator:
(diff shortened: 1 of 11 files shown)


caluma/form/validators.py
```diff
@@ -70,6 +71,27 @@ def _validate_question_multiple_choice(self, question, value, **kwargs):
                 f"Should be one of the options [{', '.join(options)}]"
             )
 
+    def _validate_question_dynamic_choice(self, question, value, **kwargs):
+        options = get_data_source_data(document, question.data_source).values_list(
```
open-dynaMIX: Are we sure we want to do this? The idea of dynamic options is, well, that they are dynamic :)
So a valid option that existed at the time of loading the question could be an invalid one at the time of validation.

I would reduce this validation to just:

```python
if not isinstance(value, str):
    raise Exception()
```

Same goes for `_validate_question_dynamic_multiple_choice`.

caluma/form/validators.py
```diff
@@ -70,6 +71,27 @@ def _validate_question_multiple_choice(self, question, value, **kwargs):
...
+        if not isinstance(value, str) or value not in options:
+            raise exceptions.ValidationError(
+                f"Invalid value {value}. "
+                f"Should be of type str and one of the options {'.'.join(options)}"
+            )
+
+    def _validate_question_dynamic_multiple_choice(self, question, value, **kwargs):
+        options = get_data_source_data(
+            kwargs["document"], question.data_source
+        ).values_list("slug", flat=True)
```
open-dynaMIX: `get_data_source_data` returns a list of `Data`-objects, not a queryset (as the dynamic options are not in the db). This means `values_list` and all other queryset-features are not available.

caluma/form/validators.py
czosel: Hmm, good point. Our current use case would prefer this to be valid though, the list is not expected to change frequently.

czosel:
@open-dynaMIX Can you take over here? @Yelinz is not around until next week.

caluma/form/validators.py
open-dynaMIX: So we expect this to fail just unfrequently ;)

I think we need to come up with a good solution for this.

Personally I'd prefer a lax validation over an error-prone.

czosel: I see your point. In our specific use-case, validation is also crucial though, because decides the next assigned group in the workflow. Idea: Shall we make this configurable in the data-source? Something like `enforce_validation: True/False`?

Issue #382, opened 2019-04-18
Wrong type of `parentAnswers` of the `Document`-type


open-dynaMIX:
Since the merge of #353 the type of `parentAnswers` of the `Document`-type is `[FileAnswer]` but should be `[FormAnswer]`:

```
type Document implements Node {
  createdAt: DateTime!
  modifiedAt: DateTime!
  createdByUser: String
  createdByGroup: String
  id: ID!
  form: Form!
  meta: JSONString!
  answers(before: String, after: String, first: Int, last: Int, question: ID, search: String, createdByUser: String, createdByGroup: String, metaHasKey: String, orderBy: [AnswerOrdering]): AnswerConnection
  parentAnswers: [FileAnswer]  # <---
  case: Case
  workItem: WorkItem
}
```

Somehow, this is due to the order of declaration of the Answers in the schema.

If the declaration of `FormAnswer` is moved below the declaration of `FileAnswer`, then this ist fixed, but then the type of the answer of `File` is changed as well:

```
type File implements Node {
  createdAt: DateTime!
  modifiedAt: DateTime!
  createdByUser: String
  createdByGroup: String
  id: ID!
  name: String!
  answer: FormAnswer  # <--
  uploadUrl: String
  downloadUrl: String
  metadata: GenericScalar
}
```

The types should not depend on the order of declaration, but be set explicitly.

czosel:
Wow, that's pretty strange. Looks a bit like a bug in graphene. I'd be ok with just leaving this documented for the moment though - we can try to create a minimal reproduction and open an issue in graphene about this in the future.

open-dynaMIX:
#424 caused a regression and had to be reverted (#429).

open-dynaMIX:
This issue is tracked in #432.

PR #416, opened 2019-05-02
Fix returning __typename for removeAnswer mutation

open-dynaMIX:
The removeAnswer mutation always returned `FileAnswer` as __typename.

This commit fixes that, so that the returned `__typename` always matches
the answer that was removed.

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -751,6 +751,7 @@ class RemoveAnswer(Mutation):
     class Meta:
         lookup_input_kwarg = "answer"
         serializer_class = serializers.RemoveAnswerSerializer
+        return_field_type = Answer
 
 
 class Mutation(object):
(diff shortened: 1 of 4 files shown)


caluma/form/schema.py
```diff
@@ -751,6 +751,7 @@ class RemoveAnswer(Mutation):
     class Meta:
         lookup_input_kwarg = "answer"
         serializer_class = serializers.RemoveAnswerSerializer
+        return_field_type = Answer
```
czosel: Small change, big effect :tada:

open-dynaMIX: Unfortunately #382 is a different issue :(

czosel: Pity. I'll go ahead and approve this one anyway, then.

PR #426, opened 2019-05-06
Implement flexible format validators

### LINKED ISSUE #360: Flexible format validations
Form validations typically include format validations for fields things like emails, social security numbers, credit card and phone numbers, etc.

After discussing with @kaldras, we propose adding a new feature which functions similarly to #320 for this:

- Add new extension point, e.g. `format-validations.py` (maybe find better name to avoid confusion with `validations.py`)
- Add some pre-defined validations to caluma core for common stuff (e.g. email)
- Add a new query to the API `allFormatValidations` which will be used in the form builder UI
- Add a new field to `formatValidation` to `TextQuestion`, `IntegerQuestion`, `FloatQuestion`
- Integrate new validations in `form/validators.py`
- A format validation is defined by
  - a slug (e.g. email)
  - a name (translated for display in the form builder, e.g. E-Mail)
  - a definition of the actual validation rule. Open question: Would this be regex? Python and JS have different engines - I'm not sure how different they are in practice. A possible escape hatch would be (if the two engines are too different) to use node.js for the backend side.

open-dynaMIX:
This commit adds flexible format validators.

There is one new query: `all_format_validators` that enables querying
all available format validators.

A text- or textarea-question can have one or multiple format_validators
assigned.

When querying a question, there is a new `FormatValidatorsConnection`,
containing all FormatValidator-objects assigned on the question.

For this there is a new `QuestionValidator` that ensures, referenced
format_validators exist.

`translate_value()` has been moved to `core.utils`, as this is used in the
`data_source`- and `form`-app.

There is a new extension-point `format_validators.py`.

When adding the tests for this, it became obvious, that there was something
wrong with the parameters for `DataSource`. As I was touching those tests
anyway, I fixed this as well.

Added commented out extension volume to docker-compose.yml. Also added
forgotten data_sources extension volume.

Additionally this commit also cleans up the snapshots for document and
question as there were quite a lot old unused snapshots.

Closes #360


open-dynaMIX:
@czosel @winged What do you think?

Let's agree on the design, before I go ahead and implement the other base FormatValidators.

czosel:
The API looks great to me as well! 👍

open-dynaMIX:
I've added the info about overriding `validate()` to the README and implemented the other base validator (there was just the phone-number left :)).

PR #430, opened 2019-05-07
Filter by answer

winged:


winged:
@czosel what do you think of this, as a first draft?

caluma/core/filters.py
```diff
@@ -217,17 +218,161 @@ def filter(self, qs, value):
...
+        return qs
+
+    def apply_expr(self, qs, expr):
+        question_slug = expr["question"]
+        match_value = expr["value"]
+
+        # hierarchy lookup mode
+        hierarchy = expr.get("hierarchy", AnswerLookupMode.FAMILY)
+        if not hierarchy:
+            hierarchy = AnswerLookupMode.FAMILY
```
czosel: Isn't this case already covered by the second argument of `get`? (same below)

caluma/core/filters.py
```diff
@@ -217,17 +218,161 @@ def filter(self, qs, value):
...
+        lookup = expr.get("lookup", self.lookup_expr)
+        if not lookup:
+            lookup = self.lookup_expr
+
+        filters = {}
+        if "." in question_slug:
+            split_slug = question_slug.split(".")
+            assert (
+                len(split_slug) == 2
+            ), "Cannot match full slug path, only form_slug.question_slug"
```
czosel: Nitpick: "full" depends on the form configuration, how about
```suggestion
            ), "Cannot match more than one level of nesting, only form_slug.question_slug"
```

caluma/core/filters.py
winged: get() could return None still, that's why

caluma/core/filters.py
winged: Changed the message a bit

PR #440, opened 2019-05-10
Implement totalCount for connections

### LINKED ISSUE #404: Implement totalCount for pagination
It looks like totalCount is not yet implemented in Caluma.

I found [a mixing](https://github.com/graphql-python/graphene-django/issues/162) but I am not sure if that is what we need.

open-dynaMIX:
This commit implements `totalCount` for connections as per graphQL spec.

Closes #404


fkm:
Thank you!

PR #450, opened 2019-05-14
Fix various evaluation issues

winged:
* When accessing a document's answers, we need values for all the questions, not just for the answers that exist.
* Also, in this context add a fallback `Answer` object, even if it doesn't exist in the DB (note it won't get saved!)


caluma/form/validators.py
```diff
@@ -183,38 +175,66 @@ def get_answers_by_question(document):
...
-                answer.question.pk: self.get_answer_value(answer, document)
-                for answer in document.answers.all()
-            }
 
         if answer.value is None:
             if answer.question.type in (
                 Question.TYPE_DYNAMIC_MULTIPLE_CHOICE,
+                Question.TYPE_DYNAMIC_CHOICE,
                 Question.TYPE_MULTIPLE_CHOICE,
+                Question.TYPE_CHOICE,
```
czosel: Hmm, single choice questions don't return ListAnswers ...

PR #455, opened 2019-05-16
fix(meta filter): use GenericScalar for meta filter value

czosel:
This is required if we want to filter by e.g. numbers instead of
strings.


winged:
Probably needs a `pytest --snapshot-update`. LGTM otherwise

czosel:
See #456

PR #461, opened 2019-05-20
Fix HasAnswerFilter: correct field for date types

winged:
Date answers use a dedicated field, so we should use it instead of
`value` when searching

### DIFF
--- a/caluma/form/tests/test_filter_by_answer.py
+++ b/caluma/form/tests/test_filter_by_answer.py
@@ -111,11 +111,18 @@ def test_query_all_questions(
     ans_subform = document.answers.get(question__slug="subform")
 
     for qtype, question in sub_questions.items():
-        answer_factory(
-            question=question,
-            document=ans_subform.value_document,
-            value=TEST_VALUES[qtype][form_value],
-        )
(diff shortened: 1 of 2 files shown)


caluma/form/tests/test_filter_by_answer.py
```diff
@@ -111,11 +111,18 @@ def test_query_all_questions(
...
+                question=question,
+                document=ans_subform.value_document,
+                date=TEST_VALUES[qtype][form_value],
+            )
+        else:
+            answer_factory(
+                question=question,
+                document=ans_subform.value_document,
+                value=TEST_VALUES[qtype][form_value],
+            )
```
czosel: This should already be implemented in `AnswerFactory`:
```
    date = Maybe("is_date", yes_declaration=Faker("date"), no_declaration=None)
```

czosel: @winged

winged: This is only for the faker. What we need is predictable values to verify that the search works.

czosel: Ah right, i missed the loop.

Issue #469, opened 2019-05-27
hasPreviousPage does not work


open-dynaMIX:
Extracted from #459 

The problem lies in `graphql-relay-py`. See the [issue](https://github.com/graphql-python/graphql-relay-py/issues/12) and the corresponding [PR](https://github.com/graphql-python/graphql-relay-py/pull/14).

`graphql-relay-py` seems pretty abandoned and the mentioned PR is from Nov, 2017 with no activity since January 2018.

[Fun fact] It's not even violating the [spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo.Fields):

> If the client is paginating with first/after, then the client **may** return true if edges prior to after exist, if it can do so efficiently, otherwise **may** return false.

czosel:
Would it be possible to integrate the fix from the PR somehow? Preferably without forking graphql-relay-py but by overwriting some method(s)?

PR #470, opened 2019-05-27
feat(pagination): implement offset pagination

### LINKED ISSUE #459: Offset pagination
Currently, we only have a cursor based navigation. To cover certain requirements in our project we should also offer a classic offset pagination.

Also, I think the `hasPreviousPage` does not really work right now.

open-dynaMIX:
This commit adds offset pagination with `limit` and `offset` filters.

This can alternatively be used instead of the already implemented cursor based
pagination.

Closes #459

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -171,9 +171,23 @@ class OrderingFilter(OrderingFilter):
         )
 
 
+class IntegerFilter(Filter):
+    field_class = forms.IntegerField
+
+
 class FilterSet(GrapheneFilterSetMixin, FilterSet):
     created_by_user = CharFilter()
     created_by_group = CharFilter()
+    offset = IntegerFilter(method="filter_offset")
+    limit = IntegerFilter(method="filter_limit")
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


open-dynaMIX:
As this is just an alternative to our cursor based pagination, @winged and I decided to merge it.

sliverc:
Just a quick comment as I have been CCed... :wink: 

For simplicity and usability of an API alternatives are to be avoided unless there is a heavy use case. Otherwise the API gets very cluttered.

So I think the question would be is there a real use case for limit/offset pagination which cannot be done with a cursor (the related issue doesn't really explain this) or is it a simple a workaround for #469 in what case I would rather fix that issue.

Just a thought... :smile:

anehx:
The problem with the cursor pagination is that we can't do the common pagination where the user is able to directly go to page XY. Instead we'd have to do a simpler version in which the user could only click next or previous page.. Since such a pagination is a common use case we thought offset pagination would be a good alternative to the current pagination.

sliverc:
@anehx 
What is nowadays best UI practice for paging then? Is it not infinite scrolling?

PR #471, opened 2019-05-27
Implement case ordering based on answer values

### LINKED ISSUE #460: Case ordering
The `allCases` graph should be able to be sorted by answers

open-dynaMIX:
This commit extends the `CaseFilterSet` in order to support ordering
by answer values to a specific question. The question slug is used a input
for this filter. If prefixed with a `-`, the ordering is reversed.

Closes #460


open-dynaMIX:
@czosel This has still some issues I have to work out, before removing `WIP:`.

Do we need to be able to sort by multiple columns?

open-dynaMIX:
@winged Would be great to have a short talk about this tomorrow.

open-dynaMIX:
@winged @czosel This is ready for review now. Much simpler without the whole nesting :tada:

caluma/workflow/filters.py
```diff
@@ -53,6 +56,48 @@ class CaseFilterSet(MetaFilterSet):
 
     has_answer = HasAnswerFilter(document_id="document__pk")
     status = case_status_filter(lookup_expr="in")
+    order_by_question_answer_value = CharFilter(
+        method="filter_order_by_question_answer_value",
+        label=(
+            "Expects a question slug. If the slug is prefixed with a dash, "
```
czosel: Nitpick: I think `dash` refers to another character ("longer minus", in german "Gedankenstricht / Halbgeviertstrich") - what we're expcing is called "hyphen" or "minus sign".
https://en.wikipedia.org/wiki/Dash
https://en.wikipedia.org/wiki/Hyphen

caluma/workflow/filters.py
```diff
@@ -53,6 +56,48 @@ class CaseFilterSet(MetaFilterSet):
...
+            "Table questions are not supported at this time."
+        ),
+    )
+
+    @staticmethod
+    def filter_order_by_question_answer_value(queryset, _, value):
+        order_by = "-order_value" if value.startswith("-") else "order_value"
+        value = value.lstrip("-")
+
+        path = value.split(".")
```
czosel: Can't we drop this altogether?

caluma/workflow/filters.py
```diff
@@ -53,6 +56,48 @@ class CaseFilterSet(MetaFilterSet):
...
+        answer_value = "value"
+        if question.type in not_supported:
+            raise RuntimeError(
+                f'Questions with type "{question.type}" are not supported '
+                f'by "filterOrderByQuestionAnswerValue"'
+            )
+        elif question.type == Question.TYPE_DATE:
+            answer_value = "date"
+        elif question.type == Question.TYPE_FILE:
+            answer_value = "file__name"
```
czosel: This question type to value field mapping is already at several places in the codebase, right? It might be worth extracting this into a function?

caluma/workflow/filters.py
open-dynaMIX: Absolutely!

caluma/workflow/filters.py
open-dynaMIX: Totally!

caluma/workflow/filters.py
open-dynaMIX: You are right, but it is much simpler to address this, after #530 is resolved.

czosel: :+1: Alright, lets solve it in another PR then.

PR #473, opened 2019-05-29
Introduce family filter for documents

winged:
This is intended to improve loading speed so you don't need to build a
large nested query anymore, which should speed up data retreival:
```graphql
        query foo ($root: ID!) {
          allDocuments(rootDocument: $root) {
            edges {
              node {
                id
              }
            }
          }
        }
```
This should get you all documents regardless of nesting level. Fetching the corresponding answers for them should then allow you to get the whole structure in a flat result

### DIFF
--- a/caluma/core/forms.py
+++ b/caluma/core/forms.py
@@ -1,4 +1,6 @@
+from graphene import List, String
 from graphene_django import forms
+from graphene_django.forms.converter import convert_form_field
 
 
 class GlobalIDFormField(forms.GlobalIDFormField):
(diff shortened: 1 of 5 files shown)


winged:
@anehx can you check whether this allows you to improve document query performance?

anehx:
Would be nice if this PR also included #474

winged:
@anehx last commit implements #474

caluma/core/forms.py
```diff
@@ -23,3 +26,20 @@ class GlobalIDMultipleChoiceField(forms.GlobalIDMultipleChoiceField):
 
     def valid_value(self, value):
         return True
+
+
+class SlugMultipleChoiceField(GlobalIDMultipleChoiceField):
+    """
+    Slug multiple choice field which allows listing multiple slugs.
+
+    Disable cleaning as we dont' know which slugs may be valid down
```
anehx:
```suggestion
    Disable cleaning as we don't know which slugs may be valid down
```

PR #490, opened 2019-06-06
allCases query: Filter for multiple status

### LINKED ISSUE #351: allCases query: Filter for multiple status
At the moment it is only possible to filter for exactly one case status. For listings it should be possible to filter for multiple status at the same time. This can be achieved by changing the filter type of status from `String!` to `List of Strings [String!]`.

open-dynaMIX:
This commit implements a case status filter, that accepts a list of case
status and filters with the `in` lookup expression.

Closes #351


czosel:
This is a breaking change, right?

open-dynaMIX:
@czosel Yes. Was this too fast? Should I revert?

czosel:
No, it's fine. Let's start with regular releases and a proper CHANGELOG as soon as we can though.

PR #505, opened 2019-06-14
feat(caluma): Remove FormAnswer type

### LINKED ISSUE #504: Answers to FormQuestions must be flat
There was a conceptual misunderstanding that led us to implement
nested answers in FormQuestions. This created performance and complexity
issues.

The discussion about this started in #489 and the following telco made it evident, that we have to revert this.

Answers to sub-questions of FormQuestions must remain flat. In turn this means, a Question can only once be answered per document and all answers in a case share the same document.

This makes the FormAnswer type obsolete.

open-dynaMIX:
There was a conceptual misunderstanding that led us to implement
nested answers. Which in turn created performance and complexity
issues.

This commit removes the FormAnswer type. Additionally also all
implementations of recursive lookups and form path handling (that only existed
as a result of this misconception) are removed.

There is a data migration that flattens the document tree.

Closes #504

BREAKING CHANGE:
 - The FormAnswer type is removed and answers to a question
   in a sub-form do not have their own document.
 - In order to access parent questions from within a table-sub-question
   JEXL, there is no need to prefix the question_slug with `parent.`
   anymore.
 - The JEXL expression `rootForm` is replaced by `form`.


open-dynaMIX:
/cc @czosel @winged @kaldras 

Let's begin :)

I'm sure I've missed some things...

caluma/form/tests/test_validators.py
```diff
@@ -44,77 +44,6 @@ def test_validate_special_fields(
...
-    answer_factory(
-        document=sub_2_document, question=sub_form_question_2.question, value="foo"
-    )
-
-    if should_throw:
-        error_msg = f"Questions {sub_form_question_1.question.slug} are required but not provided"
-        with pytest.raises(ValidationError, match=error_msg):
-            DocumentValidator().validate(main_document, info)
-    else:
-        DocumentValidator().validate(main_document, info)
```
czosel: Shouldn't we adapt this test case to something that is still meaningful with the refactor? After all we can still have nested forms, just not nested answer trees.

open-dynaMIX: This tests the DocumentValidator() which does not have to handle nesting.

open-dynaMIX: Done!

anehx:
:heart: :100: :fire:

caluma/form/serializers.py
```diff
@@ -563,6 +532,17 @@ class SaveDocumentTableAnswerSerializer(SaveAnswerSerializer):
         help_text="List of document IDs representing the rows in the table.",
     )
 
+    def _get_document_tree(self, document_id):
```
winged: Since you're not returning a nested structure here, you could just find all documents with the same family and return the bunch, doing one single query instead of recursively descending down the tables.

Also, as you're not really returning a tree, I'd suggest a different name (`_get_family_documents` comes to mind..)

caluma/form/serializers.py
```diff
@@ -612,7 +592,7 @@ def update(self, instance, validated_data):
         for answer_document in models.Document.objects.filter(
             pk__in=answer_documents.values("document")
         ):
-            children = _get_document_tree(answer_document.pk)
+            children = self._get_document_tree(answer_document.pk)
             models.Document.objects.filter(pk__in=children).update(
```
winged: Out of context for this PR, but why do we detach the table children from the toplevel family? This doesn't make any sense IMHO, and puts rocks in our path when fetching the document for validation etc

caluma/form/validators.py
```diff
@@ -157,18 +152,21 @@ def validate(self, *, question, document, info, **kwargs):
...
         # empty values are allowed
         # required check will be done in DocumentValidator
         if value:
-            validate_func = getattr(self, f"_validate_question_{question.type}")
-            validate_func(
-                question, value, document=document, info=info, answer_tree=answer_tree
-            )
+            if not isinstance(answer_tree, list):
+                answer_tree = [answer_tree]
+            for at in answer_tree:
```
winged: I'm not sure what happens here: Why is `answre_tree` a list? Is this to validate in context of table questions?

If I understand correctly, you'd get `documents` in `kwargs` and would need to iterate over those, not multiple answer trees, no?

caluma/form/validators.py
```diff
@@ -286,34 +282,36 @@ def _get_answer_value(self, answer, document, parent):
...
-                is_hidden = jexl.QuestionJexl(
-                    answer_tree, document.root_form.slug
-                ).evaluate(question.is_hidden)
+                is_hidden = jexl.QuestionJexl(answer_tree, document.form.slug).evaluate(
+                    question["is_hidden"]
+                )
                 if is_required and not is_hidden:
-                    if answer_tree.get(question.slug, None) in EMPTY_VALUES:
-                        required_but_empty.append(question.slug)
+                    if answer_tree.get(question["slug"], None) in EMPTY_VALUES:
```
winged: I'd work on the naming a bit: In some places, `answer_tree` is a list, in others, it's a dict.

caluma/form/serializers.py
open-dynaMIX: #524

caluma/form/validators.py
open-dynaMIX: Thanks for pointing this out! I've meant to come back to this and forgot about it.

caluma/form/serializers.py
open-dynaMIX: Related to #524

open-dynaMIX:
@winged Ready for review!

PR #507, opened 2019-06-14
WIP: django-management-command that executes graphql

ganwell:
This command allows to have graphql-based fixtures during development.

I use it to write an importer for a caluma-predecessor.

WIP: needs to be polished, if you are interested in this PR.


ganwell:
I just realized that GraphML can't do dependent-queries. This needs some magic added or it isn't too useful. I leave the issue open, till I wrote the importer, in case something good falls out.

ganwell:
My work yields nothing reusable

PR #512, opened 2019-06-19
Performance: Create indexes on answer 'value' and 'date'

winged:
This is used extensively in many filters, including permission layers,
so it's a good idea to have a general index on it.

If a project needs custom lookups within the value field for example,
they should still create their own indexes


czosel:
LGTM, can you still fix the linting errors?

PR #518, opened 2019-06-21
Experimental PyPy version of Caluma

winged:
This uses a separate Dockerfile, so users will still have the choice

### DIFF
--- a/caluma/settings.py
+++ b/caluma/settings.py
@@ -1,8 +1,16 @@
 import os
 import re
+import sys
 
 import environ
 from django.conf import global_settings
+
+if hasattr(sys, "pypy_version_info"):  # pragma: no cover
+    # in pypy, override psycopg2 with cffi version
+    from psycopg2cffi import compat
+
+    compat.register()
+
 
 env = environ.Env()
 django_root = environ.Path(__file__) - 2
--- a/pypy/Dockerfile
+++ b/pypy/Dockerfile
@@ -1,0 +1,35 @@
+FROM pypy:3.6
+
+# pypy image still has /usr/bin/python being cpython, we don't want that
+RUN rm /usr/bin/python && cp /usr/local/bin/pypy3 /usr/local/bin/python
+
+WORKDIR /app
+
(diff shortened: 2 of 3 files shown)


winged:
The plan also is to add this as a different tag to Dockerhub, so switching between python implementations becomes easy

sliverc:
@winged 
Not sure this is a good idea. Just simply adding a pypy docker file doesn't really add pypy support. We would need to adjust CI to run the tests with pypy as well where some code could break as well. Also the docker file is a copy of the Dockerfile which means in the future several dockers files need to be adjusted - or worse they could get out of sync.

All in all I do not see how supporting pypy brings any advantage to a caluma user as the app runs in a docker container. Or what do you see the advantages are?

ganwell:
TLDR: It depends.

@sliverc well, it is supporting a fellow open-source project more or less. They have done an awesome job ensuring cpython compatibility. I am currently trying to get pypy tested/supported on musl, just because both projects musl-libc and pypy are awesome.

I agree that this might cause additional time and effort for caluma-maintainers and caluma won't profit from pypy as long as it isn't on heavy load (JIT) or under memory pressure (pypy has more efficient memory-representation of data-structures). Also coverage-py is very slow on pypy, because it is built for cpython, so CI-cycles will slow-down.

sliverc:
maybe this better happen outside of caluma repository to experiment with PyPy and Caluma and see whether it actually brings some performance improvement in certain situations. If it works out well we can always consider to integrate it into the main repo even replacing CPython as default if the experiment shows massive performance improvements without major drawbacks.

Simply supporting PyPy in the main repo just for the sake of supporting it I wouldn't do as it increases CI run time and maintenance.

open-dynaMIX:
First benchmarks by @winged surprisingly show that with cpython caluma is about 3-4 times faster.

I don't know if he further investigated why that is the case.

ganwell:
For me pypy is a little faster running pytest, but psycopg2cffi https://github.com/chtd/psycopg2cffi/issues/106 isn't well maintained.

Pypy:
```
real    3m45.620s
user    2m48.987s
sys     0m37.897s
```
CPython:
```
real    4m5.379s
user    3m1.226s
sys     0m44.560s
```

Issue #509, opened 2019-06-15
Validate documents without workflow


czosel:
Currently, it is impossible to check if a document has been filled out correctly without configuring a workflow (specifically, without a `CompleteWorkflowFormTask` or `CompleteTaskFormTask`). For scenarios where the form part of caluma is integrated in another application it would be helpful to expose the validation state of a document in a query.

The most straight-forward implementation would probably be a new `isValid` property on the `Document` node, but since computing the validation state can be computationally intense, we might have to consider implementing this as a separate query, along the lines of
````graphql
isDocumentValid(id: ID!): DocumentValidationState

DocumentValidationState(
  isValid: Boolean
)
```

open-dynaMIX:
I absolutely agree that form validation should not depend on a workflow.

czosel:
@open-dynaMIX :+1: any feedback on the design? /cc @sliverc

sliverc:
Hmm I am not so sure about this.

Because with a `isDocumentValid` the frontend is still the one enforcing validation by asking backend whether a document is valid  instead of doing it itself. And there is no guarantee that a saved document is actually valid.

I only see this to be enforced by the backend with a status resp. when there is a submit action of a document. And that's currently done with a workflow. So I have a feeling the workflow is still the best even though not the most convenient method or do you have any other suggestions?

czosel:
You're right, if the assumption is that caluma is the "first class" backend. The feature I suggested is useful for the following case:

* There is another backend system which maintens the workflow state of an instance.
* Caluma is used as "form store"
* -> The other backend system needs to ask caluma about the validation state of a document before allowing specific state transitions

In summary, this would be an "integration feature" for cases where a specific part of caluma (the form) is integrated into an existing application.

open-dynaMIX:
If we decide implementing this, I don't think there is a reason for another query. With an `isValid` property on the document, the validation will only run, if this property is actually queried.

sliverc:
Just thought how you wanna make sure that a document is not edited once its submitted but this is solved the way that the visibility layer calls the other backend right?

I guess it is not a big deal to implement this if it helps but I think from a Caluma perspective this shouldn't be something to be promoted.

For `isValid` property or `validateDocument` mutation the difference would be that `isValid` would be a simple boolean as with `validateDocument` we could return error messages as well.

open-dynaMIX:
[...]

PR #519, opened 2019-06-21
New: query for validating documents

winged:
This allows validating documents without trying to complete a work item.

The query will return a list of errors from the validator (as far as
possible, as it currently works with exceptions when something happens)

Implements #509 via this query:

```graphql

query ValidateFoo {
  documentValidity(
    id: "RG9jdW1lbnQ6MDFjNTgwYmItZmU1Zi00ZmY2LTlmMDctMWVlNDJlNjVhNmNl"
  ) {
    edges {
      node {
        id
        isValid
        errors {
          slug
          errorMsg
        }
      }
    }
  }
}
```

resultin in a response like this
```json
{
  "data": {
    "documentValidity": {
      "edges": [
        {
          "node": {
            "id": "01c580bb-fe5f-4ff6-9f07-1ee42e65a6ce",
            "isValid": true,
            "errors": []
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/caluma/form/schema.py
+++ b/caluma/form/schema.py
@@ -1,4 +1,5 @@
 import graphene
+from django.shortcuts import get_object_or_404
 from graphene import relay
 from graphene.types import ObjectType, generic
 from graphene_django.rest_framework import serializer_converter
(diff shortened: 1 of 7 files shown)


open-dynaMIX:
@winged Nice!

Would it make sense to make these changes on top of #505? #505 contains a lot of changes to `caluma/form/validators.py`

PR #520, opened 2019-06-21
Performance improvements

winged:
Profiling the validation process showed that a lot of time is spent in
parsing (not even executing!) the JEXL expressions.

Caching the parsed expressions (and shortcutting simple "true" / "false"
expressions) improved the validation time for a large-ish document
structure from roughly 46s on my machine down to 23s.

Of the remaining 23s, 17s are spent fetching data from the DB, which
will be sped up when the "flattening" of the document structure is
merged

### DIFF
--- a/caluma/core/jexl.py
+++ b/caluma/core/jexl.py
@@ -1,7 +1,49 @@
+from itertools import count
+
 import pyjexl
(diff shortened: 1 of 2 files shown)


winged:
WIP because cache is a simple dict that doesn't evict old data yet

PR #522, opened 2019-06-21
Audit log

### LINKED ISSUE #516: Add audit logging
We could integrate django-reversion or something similar to provide audit logging. My current use case is focused on the `Answer` table, but probably we should offer this for all tables.

open-dynaMIX:
This commit adds `django_simple_history` in order to record changes to the
records.

This is implemented in the `BaseModel`, so the history is available in
all models.

Changes due to the [limitations](https://django-simple-history.readthedocs.io/en/2.7.2/common_issues.html) when using `django_simple_history`:

* Replaced `bulk_create` with `bulk_create_with_history` where used.
* Replaced `queryset.update()` with a for loop where used.

Other:

* Fixed a race condition in `caluma.workflow.tests.test_case.test_status_filter()`.

Closes #516


caluma/core/models.py
```diff
@@ -1,9 +1,40 @@
...
+        history_user_setter=_history_user_setter,
+        history_user_getter=_history_user_getter,
+    )
+
+    class Meta:
+        abstract = True
+
+
+Base = models.Model
+if settings.SIMPLE_HISTORY_ACTIVE:
```
winged: What does this do with regards to the migrations? Will enabling this at any time create any tables, and does it do so correctly? Or will it end up requiring additional migrations?

open-dynaMIX: When switching `SIMPLE_HISTORY_ACTIVE`, it's necessary to run `makemigrations` and `migrate`.

open-dynaMIX: Unfortunately there seems to be no "master switch". There is [save-without-a-historical-record](https://django-simple-history.readthedocs.io/en/2.7.2/querying_history.html#save-without-a-historical-record), but that's not really what we need.

winged: > When switching `SIMPLE_HISTORY_ACTIVE`, it's necessary to run `makemigrations` and `migrate`.

That's kind of a killer. If a client decides to switch, they can't expect to go into the container and do that manually.

I'm OK with switching on/off not being possible, but we need to instruct users properly. Consider the following cases:

| Project status     | Want the feature?    |   Action required
|----------------|-------------------------|-----------------------------------|
| New           |  No                          |   None                                |
| New           |  Yes                          |   Set the flag correctly from start  |
| Existing      |  No                           | None                                 |
| Existing      |  Yes                          | Set the flag exactly at the same time when Caluma is updated |

All this trouble and possible errors by operators make me consider simply not providing the switch and enabling the feature permanently

open-dynaMIX: I absolutely agree. The switch is removed.

PR #527, opened 2019-06-26
Fix migration for flattening document hierarchies

winged:
The previous version would choke if the ordering of answers wasn't
strictly top-to-bottom. Also, answers would only be moved one level up,
while they should be moved to the top document instead


caluma/form/migrations/0018_remove_answer_value_document_data_migration.py
```diff
@@ -12,13 +12,34 @@
...
-            a.value_document.delete()
-        a.delete()
+                try:
+                    logger.debug(
+                        f"moving answer {answer.pk} from {answer.document.pk} to root {a.document.family}"
+                    )
+                    answer.document_id = a.document.family
+                    answer.save()
+                except Exception as exc:
+                    __import__("pdb").set_trace()
```
open-dynaMIX: Wanna keep this in here?

winged: yeah nope.. removed :)

PR #528, opened 2019-06-26
Order by meta

winged:


caluma/settings.py
```diff
@@ -201,3 +201,7 @@ def parse_admins(admins):  # pragma: no cover
         "": {"handlers": ["console"], "level": env.str("LOG_LEVEL", default="INFO")}
     },
 }
+
+# Configure the fields you intend to use in the "meta" fields. This will
+# provide corresponding constants in the ordreBy filter.
+META_FIELDS = env.list("META_FIELDS", default=["camac-instance-id"])
```
open-dynaMIX: I assume the default here should be `[]`.

winged: Indeed, fixed

PR #533, opened 2019-06-27
feat: enforce conventional commits

### LINKED ISSUE #532: Use conventional commits
In order to simplify releasing Caluma, we want to enforce [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/). This will help us to auto generate release notes.

As discussed, we will not automate the release itself.

open-dynaMIX:
This commit adds enforcement of [conventional
commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) to the
pipeline.

Additionally it adds a `commit-msg` hook for git.

Closes #532

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,11 @@ before_install:
 
 install:
   - docker-compose up -d --build
+  - npm install @commitlint/{config-conventional,travis-cli}
 
 script:
   - docker-compose exec caluma black --check .
   - docker-compose exec caluma flake8
+  - ./node_modules/.bin/commitlint-travis
   - docker-compose exec caluma ./manage.py makemigrations --check --dry-run --no-input
   - docker-compose exec caluma pytest --no-cov-on-fail --cov --create-db -vv
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,1 @@
+module.exports = {extends: ['@commitlint/config-conventional']};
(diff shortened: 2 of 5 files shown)


czosel:
@winged I agree to both points (offline disussion and quality concerns). It does help however already if it's just used as a "starting point" for writing a release note by generating all the commits with their descriptions - if the quality is lacking, one can still do manual post-processing.

open-dynaMIX:
As @czosel said, this is meant as a starting point and to help semi-automatically write release notes.

But I think a general discussion about this in the team is very important!

anehx:
@open-dynaMIX @czosel @winged What's the status here? I think we discussed this properly offline, right?

commitlint.config.js
```diff
@@ -0,0 +1 @@
+module.exports = {extends: ['@commitlint/config-conventional']};
```
anehx: Isn't this the default?

.travis.yml
```diff
@@ -28,9 +28,11 @@ before_install:
 
 install:
   - docker-compose up -d --build
+  - npm install @commitlint/{config-conventional,travis-cli}
```
anehx: Wow, I honestly didn't know you could do that :joy:

commitlint.config.js
open-dynaMIX: @anehx According to the [docs](https://commitlint.js.org/#/guides-local-setup?id=install-commitlint) this is needed.

PR #542, opened 2019-07-03
Feature: multi meta filters (Maintenance branch pre-505)

winged:
/cc @anehx


anehx:
You don't need to do this for pre 505..

winged:
In that case, closing the PR again

PR #544, opened 2019-07-03
feat(filters)!: metaValue filter now takes a list

winged:
BREAKING CHANGE: check used metaValue and change them from single value to list.

In some projects, we need to filter by multiple meta values. This change
allows passing a list of meta filters instead

Obsoletes #543

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -317,10 +317,16 @@ class MetaValueFilter(Filter):
     def filter(self, qs, value):
         if value in EMPTY_VALUES:
             return qs
-        meta_key = value["key"]
-        meta_value = value["value"]
-        lookup = value.get("lookup", self.lookup_expr)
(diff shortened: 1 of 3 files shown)


winged:
> Before we merge this is it not time to start creating a first release and then write a changelog for user relevant changes - including breaking changes? Otherwise we will have issues in the projects already using Caluma. As far as I understand we already do releases in ember caluma.

I think as long as we're the only actual users and we're still in 0.x mode, this can be handled in a relaxed manner. Generally speaking of course you're right!

anehx:
@winged IMHO we should start using releases ASAP even if we are the only users at the moment. We had **way** to many problems because we had an old image version and no one knew why. Releases would fix alot of our pain points and it's not that big of an effort..

sliverc:
I have changed my review to comment so you can decide whether you want to do a release before or not without my review blocking you.

PR #547, opened 2019-07-03
feat(filters): Meta lookup: extend to LT,GT,LTE,GTE

winged:
/cc @anehx


winged:
@anehx Dates should work as requested, provided they're in a sane format. The field is JSON, so cannot store "actual" dates

PR #549, opened 2019-07-03
chore(ci): Use pytest-flake8 and pytest-black

winged:
Instead of calling flake8 / black explicitly, add them to pytest.  This
should help with forgetting to call those tools before committing, as
it's only one commandline call now instead of three.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,5 @@ install:
   - docker-compose up -d --build
 
 script:
-  - docker-compose exec caluma black --check .
-  - docker-compose exec caluma flake8
   - docker-compose exec caluma ./manage.py makemigrations --check --dry-run --no-input
   - docker-compose exec caluma pytest --no-cov-on-fail --cov --create-db -vv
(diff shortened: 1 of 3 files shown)


sliverc:
I see the benefit this could be bring, though not so sure about this change because of following reasons:

* performance: potentially black, flake8 are run too many times (with pytest, editor, pre-commit etc.) see also https://github.com/tholo/pytest-flake8/issues/19
* flake8 doesn't seem to honour default flake8 configuration which would be needed for editor (see https://github.com/tholo/pytest-flake8/issues/22) - one of the reasons why CI is currently failing.


I think it's kind of developer taste which way to go so would it be an option to document usage of pytest-flake8 and pytest-black in [CONTRIBUTING.md](https://github.com/projectcaluma/caluma/blob/master/CONTRIBUTING.md) and a developer can then choose whether to use its own editor, pre-commit or pytest plugins. What do you think?

Otherwise if we want to integrate per default into Caluma it would be good to get the feedback of other regular committers.

winged:
Yeah makes sense

PR #550, opened 2019-07-03
feat(filters): Add hasAnswer filter on workitems

winged:
work items have two documents that are relevant to them: their own
(direct) document, optional, and the document attached to their case.

This adds the hasAnswer filter for both of them.

/cc @fkm


caluma/conftest.py
```diff
@@ -141,3 +141,13 @@ def data_source_settings(settings):
 @pytest.fixture
 def history_mock(mocker):
     mocker.patch.object(HistoricalRecords, "create_historical_record")
+
+
+@pytest.fixture
+def simple_case(case_factory, document_factory, question_factory, answer_factory):
+    question = question_factory()
+    document = document_factory()
+    answer = answer_factory(document=document, question=question)
```
sliverc: overlooked this, but ci says there is a linting issues as defined asnwer variable is actually not used.

winged: Indeed, fixed

PR #554, opened 2019-07-05
feat(filters): Text search across multiple answers

winged:
This introduces a filter named `searchAnswers` that allows you to search
text within documents/cases across multiple answers.

The way it works is you pass a list of questions to search, and a search
string. The search string is split by words. Each word must then be
found in at least one answer for a document/case to show up.

You may optionally provide a lookup type: `STARTSWITH`, `TEXT`,
`CONTAINS`.

Note that depending on the database, those may or may not perform well,
as we do not at this point in time preemptively optimize by creating
indexes etc.

/cc @fkm

### DIFF
--- a/caluma/form/filters.py
+++ b/caluma/form/filters.py
@@ -19,6 +19,7 @@ from ..core.filters import (
     SearchFilter,
     SlugMultipleChoiceFilter,
 )
+from ..form.models import Answer, Question
 from . import models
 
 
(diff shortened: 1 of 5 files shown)


winged:
Note this is currently WIP, as Django complains about the `__search` lookup with a deprecation warning. We should probably update Django (or was there some config I missed to enable FTS on PostgreSQL?)

sliverc:
I don't think search lookup is supported on json field. Might this why a warning is shown?

But in any case we should update to Django 2.2 as it is the latest LTS release.

winged:
> Shouldn't this filter be in the `form` app?

I've put it in `core` because it's used in both apps. But topic-wise, you're right of course

open-dynaMIX:
I'd put it in the forms app, as there is a clear dependency on forms.

IMO dependencies from `workflow` to `form` are ok, but core should never depend on `form` or `workflow` and `form` should never depend on `workflow`.

That's also what @sliverc told me.

winged:
> I'd put it in the forms app, as there is a clear dependency on forms.

Makes sense, the HasAnswer filter is there as well. Moved it

PR #558, opened 2019-07-08
chore(ci): Pin pydocstyle to fix CI

sliverc:
This works around issue [0] till it is fixed upstream

[0] https://gitlab.com/pycqa/flake8-docstrings/issues/36

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,6 +12,7 @@ flake8-isort==2.7.0
 flake8-string-format==0.2.3
 flake8-tuple==0.4.0
 isort==4.3.21
+pydocstyle==3.0.0
 pytest==5.0.1
 pytest-cov==2.7.1
 pytest-django==3.5.1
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -12,6 +12,7 @@ flake8-isort==2.7.0
 flake8-string-format==0.2.3
 flake8-tuple==0.4.0
 isort==4.3.21
+pydocstyle==3.0.0
```
winged: please add a comment as to why it's pinned low, and when it can be updated

sliverc: I rather add this into the PR which renovates open later on. That PR will fail anyway. And at some point it would be great to pin all deps using Pipenv...

winged: Fair enough

Would this be a best practice for our other projects as well? Maybe we should discuss it in the team

sliverc: I think we should get some experience with Pipenv (also in terms of renovate) and once this works well it would be good to discuss to make this the standard (could be done in Caluma as heavily used).

Using pipenv this error just wouldn't have appeared. I open an issue.

PR #561, opened 2019-07-08
Update to django 2.2 as latest LTS

sliverc:


caluma/core/visibilities.py
```diff
@@ -85,6 +85,6 @@ def filter_queryset(self, node, queryset, info):
                 result_queryset = result_queryset.union(class_result)
 
         if result_queryset is not None:
-            queryset = queryset.filter(pk__in=result_queryset)
+            queryset = queryset.filter(pk__in=result_queryset.values("pk"))
```
winged: Hm is the `.values()` really needed here?

sliverc: if this is not done it leads to an error 'subquery has too many columns'. Makes sense to me though that there is no auto magic anymore but one has to be specific with the column to compare.

PR #569, opened 2019-07-10
Avoid calling of filterset class when there are not filtering args

sliverc:
Constructing a django filter form is a performance hit and should only be done when there are actual filters.

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -470,6 +470,50 @@ class DjangoFilterConnectionField(
         queryset.query.select_related = default_queryset.query.select_related
         return queryset
 
+    @classmethod
+    def connection_resolver(
+        cls,
+        resolver,
+        connection,
+        default_manager,
+        max_limit,
+        enforce_first_or_last,
+        filterset_class,
+        filtering_args,
+        root,
+        info,
+        **args,
+    ):
+        # building form within filterset class is a performance hit
+        # and should only be done if there are actual filter arguments
+        filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}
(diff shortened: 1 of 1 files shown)


sliverc:
This is a fairly ugly workaround with a gain of 5% of time with queries with many nested connections.

So question is whether it's worth it or not? If yes and we do not see any downside in this patch I can also open a PR upstream where the fix would be much simpler.

czosel:
An upstream PR is always a good idea, no matter if we decide to merge this or now, right?

sliverc:
Doubled checked this patch won't be applicable upstream as it is also possible to overwrite default queryset in a FilterClass. A test is actually failing and it would not be backwards compatible.

We don't do this in Caluma so the patch would still be valid in our case.

caluma/core/filters.py
```diff
@@ -489,6 +489,59 @@ def merge_querysets(cls, default_queryset, queryset):
...
+        default_manager,
+        max_limit,
+        enforce_first_or_last,
+        filterset_class,
+        filtering_args,
+        root,
+        info,
+        **args,
+    ):
+        class QuerysetManager(object):
```
winged: This builds a new class for each request.. If we want to have a custom manager optimized for our use case, why not move it to the model itself, or at least out to module scope? That seems to be the better place for it

sliverc: this class uses outscope so don't quite see how this could be done on model level?

sliverc: updated so it actually calls super to skip filtering.

sliverc:
Ready for review again and as it is now I think this is mergeable.

PR #572, opened 2019-07-11
Convert options m2m field to array field

sliverc:
This reduces queries when querying options. Especially helpful on nested form questions with options.


sliverc:
@czosel This is now running with the down side that all django filters would need to implemented manually now.
With large queries the performance improvement is around 20%.

Whereas #569 is simply a bit ugly this PR actually makes maintaining caluma harder as filters will need to be implemented as Django Filter and potentially manual in python as well.

Please comment what you think how we should move forward.

winged:
I do like the idea, but did you look at the SQL queries generated? How optimized are `join`s via array fields? How did you measure the performance improvement?

Taking this a bit further, do we have some other fields / m:n connectoins that could profit from this?

sliverc:
As of change #596 data loaders do not work at the moment and this change also kind of breaks with consistency and maintainability.

Closing this PR. Query explosion should still be solved in the future but better with proper prefetching.

PR #595, opened 2019-07-19
Check authentication only once

sliverc:
Once a user is set on request authentication is already passed and we can exit.


sliverc:
this is an approximate performance gain of 10% on large queries.

PR #600, opened 2019-07-22
Optimize setting of document familiy

sliverc:
There is no need to set the family one by one but a simple update works just the same.


sliverc:
I overlooked commit 44e3bedd70420700b8e503f3ed189b3a52b983ed as it seems when running git blame.

Let me add a docstring as this piece of code is not obvious.

open-dynaMIX:
:+1:

sliverc:
Closing in favor of #601

PR #605, opened 2019-07-24
fix(table): fix flaky detaching of documents when updating TableAnswers

czosel:
Under certain conditions (which are not fully understood until now) the
main documents' family was changed when an enclosed table question was
updated.

This adds a test that demonstrates the flaky behaviour, and a bugfix
that fixes it and tries to clarify the code.

### DIFF
--- a/caluma/form/serializers.py
+++ b/caluma/form/serializers.py
@@ -534,12 +534,16 @@ class SaveDocumentTableAnswerSerializer(SaveAnswerSerializer):
     )
 
     def _get_document_tree(self, document_id):
-        answers = models.AnswerDocument.objects.filter(document_id=document_id).values(
-            "answer"
-        )
-        child_documents = models.Document.objects.filter(answers__in=answers).distinct()
+        """Get all child documents (table rows) of a given document, and the document itself."""
 
-        for child_document in child_documents:
+        answers = models.Answer.objects.filter(document_id=document_id)
+        child_documents = (
+            answers.filter(documents__isnull=False)
+            .values_list("documents", flat=True)
+            .distinct()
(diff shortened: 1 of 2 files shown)


czosel:
@sliverc I just had another look - yesterday's fix actually seems fine to me. We got confused because we were calling the return value of `_get_document_tree` `children`, but it also returns the document ID that was passed to it (`yield document_id`, `L545`). I renamed the variable to `tree` and added a comment for clarification.

I'll test this in one of our projects now, afterwards this should be good to go from my side.

czosel:
Test in customer project was successful :+1:

caluma/form/serializers.py
```diff
@@ -534,12 +534,16 @@ class SaveDocumentTableAnswerSerializer(SaveAnswerSerializer):
     )
 
     def _get_document_tree(self, document_id):
-        answers = models.AnswerDocument.objects.filter(document_id=document_id).values(
-            "answer"
+        """Get all child documents (table rows) of a given document, and the document itself."""
+
+        answers = models.Answer.objects.filter(document_id=document_id)
```
sliverc: better to be filter after type `TYPE_TABLE` as well? Could then call it table_answers to make the code better understood.

caluma/form/serializers.py
```diff
@@ -534,12 +534,16 @@ class SaveDocumentTableAnswerSerializer(SaveAnswerSerializer):
...
+        child_documents = (
+            answers.filter(documents__isnull=False)
```
sliverc: I have a feeling isnull on a m2m is always false... with change above this would not really be needed though.

PR #607, opened 2019-07-25
Set document family properly on whole family

sliverc:
Avoiding exception when tables in tables are used.


sliverc:
This is a follow up on #605 

I don't see why `documents__isnull=True` is needed. When added it does an inner join and if left out it uses `LEFT OUTER JOIN` so in that case it cannot be null.

PR #611, opened 2019-07-25
Use different tag for caluma development

ganwell:
If the user already downloaded the release image and then decides to start
development, docker will use the already cached image, no rebuild is triggered.
By using the tag latest-dev instead of latest, docker can't use the cached
image.

**Update**: Changing Title and Description because the root-cause is different, the old Issue was:

Title: Add a way to install dev_requirements to testing-how-to

Maybe I missed something, but I had to do this to run the testing-commands.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -6,6 +6,7 @@ services:
     ports:
       - "${DATABASE_PORT:-5432}:${DATABASE_PORT:-5432}"
   caluma:
+    image: projectcaluma/caluma:dev
     build:
       context: .
       args:


sliverc:
https://github.com/projectcaluma/caluma/blob/master/docker-compose.override.yml#L12

This is done there and not needed manually in case docker-compose is actually run.

Below it also says `docker-compose build --pull` which might solve your problem as the caluma container might be been created already without the override file.

What we should do though is to tag the image in override as `dev` or so those errors are clear that the image is not build yet and dc will build it.

ganwell:
@sliverc CI is ok with this change, removing WIP state.

docker-compose.override.yml
```diff
@@ -6,6 +6,7 @@ services:
     ports:
       - "${DATABASE_PORT:-5432}:${DATABASE_PORT:-5432}"
   caluma:
+    image: projectcaluma/caluma:latest-dev
```
sliverc: `latest` is a specific tag with a special name, other tags might also be the latest but with a different argument set for instance.

Therefore I would simply call this tag `dev`.

ganwell:
Changed to dev.

PR #615, opened 2019-07-26
Keep json field value type when ordering

### LINKED ISSUE #608: Meta ordering
Currently, Caluma supports ordering by meta fields as described [here](https://github.com/projectcaluma/caluma#common). However it orders all of those properties as string. We need to differentiate between integer, string and date etc. for ordering.

Current:
`3 > 20 > 100`

Expected:
`100 > 20 > 3`

sliverc:
Additionally this avoids annotate which would be added even though ordering
by meta field is not really used.

Fixes #608

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -1,4 +1,3 @@
-import itertools
 from functools import reduce
 
 import django.forms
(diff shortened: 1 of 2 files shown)


sliverc:
This is just a fix to get ordering working for different json types.

I think all in all we should rethink the meta field feature as it is very static and hence not really Caluma like. See https://github.com/projectcaluma/caluma/issues/616

PR #617, opened 2019-07-26
Create GinIndex for JSONFields

ganwell:
Improves performance for searching and filtering for meta/value JSONFields

@sliverc It seems we alread use GinIndexes for a array- and a text-type:  https://github.com/projectcaluma/caluma/blob/0083725d1a4fca67b59700839e3d7ee079b833c4/caluma/workflow/models.py#L192


caluma/form/models.py
```diff
@@ -220,6 +233,7 @@ class Meta:
         # a question may only be answerd once per document
         unique_together = ("document", "question")
         indexes = [models.Index(fields=["value"]), models.Index(fields=["date"])]
+        indexes = [GinIndex(fields=["meta", "value"])]
```
sliverc: there are two indexes definitions - there should only be one for this to work properly.

ganwell: oh dear, that was a bad one. would be interesting to get linter warning for that.

sliverc:
Run black and we are good to merge this.

PR #621, opened 2019-07-29
Return 401 status code when authentication failed

### LINKED ISSUE #477: 401 Error
Currently if the client is not authenticated we get a 200 status code with this response body:

```json
{
   "errors":[
      {
         "message":"401 Client Error: Unauthorized for url: http://keycloak.local/auth/realms/ebau/protocol/openid-connect/userinfo",
         "locations":[
            {
               "line":2,
               "column":3
            }
         ],
         "path":[
            "allCases"
         ]
      }
   ],
   "data":{
      "allCases":null
   }
}
```

Wouldn't it make more sense to send an actual 401 response?

sliverc:
Fixes #477

For this needed to replace middleware with an extended django view
where a response can be handled directly.

### DIFF
--- a/caluma/settings.py
+++ b/caluma/settings.py
@@ -158,10 +158,7 @@ ADMINS = parse_admins(env.list("ADMINS", default=[]))
 
 # GraphQL
 
-GRAPHENE = {
-    "SCHEMA": "caluma.schema.schema",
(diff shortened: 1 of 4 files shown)


anehx:
@open-dynaMIX Is there something preventing this from merging?

PR #624, opened 2019-07-30
 Do not swallow question-is-missing-exception 

### LINKED ISSUE #609: Document validity raises cryptic errors
The `documentValidity` graph throws way too cryptic errors:

```
Error while evaluating 'is_hidden' expression on question 'test': 'nonexistent'|answer == ':+1'
```

When the real error was thrown in `caluma/form/jexl.py:35`:

```
Question could not be found: 'nonexistent'
```

sliverc:
Fixes #609

As Caluma validates jexl when saving it technically the only exception
which can happen here is that an answer is missing.
Hence removing whole block.

It might make sense to have a possibility to validate all jexl
of a form to avoid Runtime error but this should happen outside
the document validation.

### DIFF
--- a/caluma/form/jexl.py
+++ b/caluma/form/jexl.py
@@ -27,10 +27,7 @@ class QuestionJexl(JEXL):
 
     def answer_transform(self, question):
         try:
-            return (
-                self.answer_by_question.get(question)
-                or self.answer_by_question.get("parent", {})[question]
-            )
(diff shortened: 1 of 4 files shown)


sliverc:
I have found some corner cases which are not properly validated when save question.

So this is more complicated then I thought. Let me split up the PR to the bits which are less controversial.

PR #625, opened 2019-07-30
Remove left overs of validation answer tree

sliverc:
* There is no need for answer_tree anymore as of #505
* parent doesn't exist and can be removed
* requiredness does not to be inherited as a form question does actually
  have answers.


open-dynaMIX:
@winged LGTM, you might want to have a look as well.

PR #627, opened 2019-07-30
Do not swallow question is missing exception

### LINKED ISSUE #609: Document validity raises cryptic errors

sliverc:
Fixes #609 

Based on #625 which needs to be merged first

### DIFF
--- a/caluma/form/jexl.py
+++ b/caluma/form/jexl.py
@@ -27,10 +27,7 @@ class QuestionJexl(JEXL):
 
     def answer_transform(self, question):
         try:
-            return (
-                self.answer_by_question.get(question)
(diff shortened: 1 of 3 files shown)


sliverc:
This fixes the issue but all in all in the long run I think validation of jexl expression needs to be improved so all this exceptions never actually happen during usage.

What I could think of is to provide a form validation possibility which evaluates all jexl using default answers.

PR #631, opened 2019-08-02
feat(WorkItems): add caseMetaValue filter

open-dynaMIX:
This commit transforms the `MetaValueFilter` into a more generic
`JSONValueFilter`.

This filter is then used for a `caseMetaValue` filter to the workitems. This
allows for filtering WorkItems based on meta field values.

Closes #610

BREAKING CHANGE: `MetaValueFilterType` has been renamed to
`JSONValueFilterType` and `MetaLookupMode` to `JSONLookupMode`.

PR #635, opened 2019-08-07
fix(filters): Fix meta filtering with boolean values

czosel:
This fixes a regression introduced in #631: `KeyTextTransform` casts the Boolean `True` to the String `"true"`, which is not what we want when we're filtering for boolean values.

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -255,8 +255,15 @@ class JSONValueFilter(Filter):
             # That's why we annotate the queryset with the value.
             # Some discussion about it can be found here:
             # https://code.djangoproject.com/ticket/26511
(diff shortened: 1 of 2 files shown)


czosel:
Done!

czosel:
Thanks for the review @winged :slightly_smiling_face:

czosel:
@open-dynaMIX even though this is merged already, your review is still appreciated :wink:

open-dynaMIX:
@czosel I already did take a look. Nice catch! I was hoping we could just switch `KeyTextTransform` with `KeyTransform` which doesn't cast the values into a TextField. But a short test revealed that then `CONTAINS` somehow fails. As this would be pure cosmetics and the code is working absolutely correct now, I didn't want to spend more time on it.

czosel:
Alright, thanks! :+1:

PR #657, opened 2019-08-21
feat(history): add historical API

### LINKED ISSUE #650: Expose historical tables on API
It would be nice if we'd have some queries that allow accessing the historical data that we're recording. This probably needs some more detailed design before we jump into implementation.

/cc @open-dynaMIX @winged

open-dynaMIX:
This commit adds a schema for querying historical records.

In order to handle historical FileAnswers, the files are not
automatically removed from the bucket anymore, but renamed, so they stay
accessible for historical queries. For this, the `Minio` storage client
now has a method `move_object()`.

Closes #650


sbor23:
LGTM

PR #686, opened 2019-09-16
feat(history): support asOf per type

### LINKED ISSUE #683: documentAsOf doesn't handle variables correctly
The `documentAsOf` graph doesn't handle aliased variables properly:

```graphql
query(documentId: $ID, date: DateTime!) {
  documentAsOf(id: $documentId, asOf: $date) {
    answers {
      edges {
        node {
          id
        }
      }
    }
  }
}
```
This fails because the API expects `asOf` to always be named like this in the variables: 
https://github.com/projectcaluma/caluma/blob/6449e13e750e8fca541a51271fa521412c1e5456/caluma/form/historical_schema.py#L154

open-dynaMIX:
This commit implements the argument `asOf` for all related types in the
historical schema.

This removes the need for accessing variables from other types (i.e. the
top level `documentAsOf`). But also it gives us much more flexibilty
when constructing queries.

Closes #683

### DIFF
--- a/caluma/form/historical_schema.py
+++ b/caluma/form/historical_schema.py
@@ -9,9 +9,9 @@ from . import models
 from .schema import (
     Answer,
     DateAnswer,
-    Document,
     FileAnswer,
     FloatAnswer,
+    FormDjangoObjectType,
     IntegerAnswer,
     ListAnswer,
     StringAnswer,
@@ -106,7 +106,9 @@ class HistoricalFile(ObjectType):
     name = graphene.String(required=True)
     download_url = graphene.String()
     metadata = generic.GenericScalar()
-    answer = graphene.Field("caluma.form.historical_schema.HistoricalFileAnswer")
+    historical_answer = graphene.Field(
+        "caluma.form.historical_schema.HistoricalFileAnswer"
+    )
     history_date = graphene.types.datetime.DateTime(required=True)
     history_user_id = graphene.String()
     history_type = graphene.String()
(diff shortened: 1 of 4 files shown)


open-dynaMIX:
@anehx IMO this is a much cleaner solution than what we had in mind first. With this you can query answers from multiple points in time in the same `documentAsOf` operation.

open-dynaMIX:
@anehx Yes it does. I did not mark it as breaking change, as this is not used anywhere yet. But technically it is.

Do you think I should?

open-dynaMIX:
@anehx @czosel I've declared the breaking change now. And also added another commit with another breaking change. Could you please review again?

caluma/form/historical_schema.py
```diff
@@ -9,9 +9,9 @@
 from .schema import (
     Answer,
     DateAnswer,
-    Document,
     FileAnswer,
     FloatAnswer,
+    FormDjangoObjectType,
```
anehx: What is this?

open-dynaMIX: This is our ObjectType we use for resolving question and answer types. It can be found[here](https://github.com/projectcaluma/caluma/blob/master/caluma/form/schema.py#L63). I was not able to inherit from the `Document` type, so I've made it analoguous to this class. We could also just use `DjangoObjectType` here but it doesn't hurt anyway.

anehx: Alright

PR #690, opened 2019-09-19
fix(model): use natural key for form_question model

czosel:
This enhances loading and dumping of the database (same argument as
for question_option)


open-dynaMIX:
@czosel This is a breaking change, right? AFAIK we can't just switch between UUID- and Char-fields. Did you try the migration with populated tables?

czosel:
Right - not yet. We need a data migration for that, right?

czosel:
Turns out the migration doesn't fail even with populated, but to change the primary keys to the new format you have to do some trickery - it seems that the following works:
```
>>> from caluma.form.models import FormQuestion
>>> for fq in FormQuestion.objects.all():
...     FormQuestion.objects.filter(form=fq.form, question=fq.question).delete()
...     fq.save()
```
Explanation:

* we can't just call `save()` on each record due to unique_together constraint (form, question)
* `fq.delete()` won't delete anything because `fk.pk` already returns the natural key (which doesn't exist in the DB yet - so we use `filter` to find the record
* just call `save` to add the record with the natural primary key

What do you think, is this acceptable? Then I'd put it in the migration.

open-dynaMIX:
Could be a way to go. Do the `Form` and `Question` tables pickup the changed foreignkeys? [This](https://code.djangoproject.com/ticket/25012) could be an issue here. We have to be very careful as this is pretty much open-heart surgery ;)

Also what's with the failing test?

winged:
> [...] Do the `Form` and `Question` tables pickup the changed foreignkeys?

The foreign key lies in the `FormQuestion` table, the `Form` and `Question` are being referenced. So the workaround migration proposed by @czosel should indeed work.

Luckily, Django runs migrations within transactions on postgres, so I suggest going ahead with the proposed migration.

czosel:
@open-dynaMIX @winged ready for review :tada:

PR #694, opened 2019-09-23
fix(history): do not hide removed records

open-dynaMIX:
This commit makes sure that also removed records are included in
historical responses.


anehx:
@winged @czosel could you both take a look at this so we can merge it?

PR #703, opened 2019-09-30
feat(history): add cleanup command

open-dynaMIX:
This commit adds a cleanup command for historical records:

```
usage: manage.py cleanup_history [-h] [--no-clean-dups] [--dry] [-k KEEP]
                                 [--version] [-v {0,1,2,3}]
                                 [--settings SETTINGS]
                                 [--pythonpath PYTHONPATH] [--traceback]
                                 [--no-color] [--force-color]

Cleanup historical records.

optional arguments:
  -h, --help            show this help message and exit
  --no-clean-dups
  --dry
  -k KEEP, --keep KEEP  Duration we want to keep the records. E.g. '6 months',
                        '1 year'. Uses dateparser.
```

### DIFF
--- a/caluma/core/management/commands/cleanup_history.py
+++ b/caluma/core/management/commands/cleanup_history.py
@@ -1,0 +1,42 @@
+from dateparser import parse
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+from simple_history.models import registered_models
+
+
+class Command(BaseCommand):
+    """Cleanup historical records."""
+
+    help = "Cleanup historical records."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry", dest="dry", action="store_true")
+        parser.add_argument(
+            "-k",
+            "--keep",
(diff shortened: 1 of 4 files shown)


caluma/core/management/commands/cleanup_history.py
```diff
@@ -0,0 +1,42 @@
...
+
+    help = "Cleanup historical records."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry", dest="dry", action="store_true")
+        parser.add_argument(
+            "-k",
+            "--keep",
+            dest="keep",
+            default="",
```
sbor23: A little to the party, but shouldn't the `keep` option be required without a default?
As of now, if the argument has been forgotten, the whole history would be deleted, as the filter in L33 isn't applied.
Or am I mistaken?

winged: Good catch! I'll create an issue from it

PR #708, opened 2019-10-02
Filter cleanup and negation

winged:
This introduces filters that can be chained (same filter multiple times)
and negated.  Currently this is VERY WIP, but shows what can be done in
the allDocuments edge.

Code needs to be cleaned up and applied to the other filterable edges as well,
and tests adjusted etc..

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -1,6 +1,6 @@
 # GraphQL
 
-As noteds before, Caluma uses a GraphQL API for communication between
+As noted before, Caluma uses a GraphQL API for communication between
 frontends, other services, and the Caluma service itself.
 
 The easiest way to get to know Caluma's data
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -1,0 +1,92 @@
+# GraphQL
+
+As noteds before, Caluma uses a GraphQL API for communication between
+frontends, other services, and the Caluma service itself.
+
+The easiest way to get to know Caluma's data
+structures is to start the service and navigate to
+[localhost:8000/graphql/](http://localhost:8000/graphql/) (assuming you
+have the development server running).
+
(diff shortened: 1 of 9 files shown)


winged:
Example of what will be possible with this:
```graphql
{
  allDocuments(
   filter: [
    {
      searchAnswers: {
        questions: ["some-text"],
        value: "stuff that I want", lookup: TEXT
      }
    },
    {
      searchAnswers: {
        questions: ["some-text"],
        value: "things that I don't want", lookup: TEXT
      },
      invert:true
    }
  ]
  ) {
    edges {
      node {
        id
      }
    }
  }
}
```

winged:
To easen the transition, I had the idea this weekend to provide the "old-style" filters as well for a release cycle or two. So instead of moving all the filters into the new structure, we would provde both interfaces. Shouldn't be too hard.

What do you guys think? @open-dynaMIX @anehx @kaldras @czosel

czosel:
I really like the idea! Since we have a first release now, it would be good to get used to a workflow that involves deprecations and control the pace of breaking changes. Ideally, we could mark the old filters as deprecated directly in the schema.

open-dynaMIX:
Great idea! :+1:

winged:
Added the "old-style" filters back now. However even though I tried almost all day, I couldn't figure out how to mark them as deprecated. I'm tempted to get this merged nonetheless and just doing it in the documentation instead.

docs/graphql.md
```diff
@@ -0,0 +1,92 @@
+# GraphQL
+
+As noteds before, Caluma uses a GraphQL API for communication between
```
open-dynaMIX:
```suggestion
As noted before, Caluma uses a GraphQL API for communication between
```

PR #721, opened 2019-10-06
chore(docs): add documentation on workflow concepts

czosel:
Turns out vacation is the perfect environment to write some documentation 😅 

[Rendered](https://github.com/projectcaluma/caluma/blob/bcab7071fd4954d59c206221b499ccfcacd8e167/docs/workflow-concepts.md)

This adds a draft for the first chapter for the documentation of Caluma's workflow system. I thought of the following structure:

1) concepts (workflow, tasks, cases, work items and how they fit together)
2) implementation (how those concepts can be applied on the GraphQL API)
3) supported workflow patterns

Where those docs go probably still needs to be discussed - I figured I'd just start on the content in the meantime 😉

I'd appreciate some feedback on this - I tried to set the barrier of entry relatively low and to explain the basics in detail. This is not meant as a technical reference, more as tutorial/guide. A bit like the [Ember.js guides](https://guides.emberjs.com/release/). If you agree with the direction this is headed, I'll add chapter two which exists as an analog draft already.

@open-dynaMIX @kaldras @anehx @ganwell @winged


winged:
Awesome work!

But that keming tho...

czosel:
Good call! I just Oxygen, the same font as on the caluma.io website. Probably not a good idea if the font is not embedded into the svg - I’ll look into it.

open-dynaMIX:
Great work! :100:

sbor23:
Looking good, certainly clear enough for n00b in caluma like me.
Eagerly awaiting the last chapters. ;)
:1st_place_medal:

winged:
Cleaned it up a bit in #723, and merged by @open-dynaMIX

PR #724, opened 2019-10-10
Filters: Syntactic separation of ordering and filtering

winged:
We want to separate filtering and ordering syntactically, so it doesn't
make sense to allow filtering in the filter[] list in the first place.

### DIFF
--- a/caluma/core/ordering.py
+++ b/caluma/core/ordering.py
@@ -1,0 +1,101 @@
+from typing import Any, Tuple, Union
+
+from django import forms
+from django.db.models.expressions import CombinedExpression, F, Value
+from django.db.models.query import QuerySet
+from django_filters.fields import ChoiceField
+from django_filters.rest_framework import Filter
+from graphene import Enum
+from graphene_django.forms.converter import convert_form_field
+from graphene_django.registry import get_global_registry
+from rest_framework import exceptions
+
+OrderingFieldType = Union[F, str]
+
+
+class CalumaOrdering(Filter):
+    """Base class for new-style ordering filters.
+
+    This is not really a filter in the rest framework sense,
+    but we use it's infrastructure to get conversion to GraphQL
+    types.
+
+    This is an abstract base class. To use it, you need to
+    implement the `get_ordering_value()` method to return a tuple
+    of `(queryset, ordering_field)`. The queryset may be modified,
+    for example if you need to annotate it to get to the desired
+    ordering field.
+    """
+
+    def get_ordering_value(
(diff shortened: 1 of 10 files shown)


winged:
Got a good bit ahead today. This is now possible:

```graphql
query new_ordering {
  allDocuments(
    filter: [
    	{metaHasKey: "camac-instance-id"}
    ],
    order: [
      {meta: "foobar", direction: ASC},
      {answerValue: "nutzungsart"}
    ]
  ) {
      ...
  }
}
```

winged:
This should now be ready for review.

Please have a good look at the new order types. Especially:
* Are the filters everywhere you expect them?
* Do the "attribute" filters cover all fields you potentially want to order by?
* Are all existing "orderBy" use cases covered by the new ones?

fkm:
Probable oversights:
* `order` seems to be missing from `allWorkItems`.
* There's an `oderBy` on the `WorkItemFilterSetType`.
* There's an `oderBy` on the `CaseFilterSetType`.
* There's an `orderByQuestionAnswerValue` on the `CaseFilterSetType`.
* There's an `oderBy` on the `QuestionFilterSetType`.

Questions:
* What does `documentForm` do on the `CaseFilterSetType`?
* Will cases be found recursively for `case`, `caseDocumentHasAnswer`, and `caseMetaValue` on  the`WorkItemFilterSetType`?

caluma/core/ordering.py
```diff
@@ -0,0 +1,159 @@
...
+    field_class = forms.CharField
+
+    def get_ordering_value(
+        self, qs: QuerySet, value: Any
+    ) -> Tuple[QuerySet, OrderingFieldType]:
+
+        return qs, CombinedExpression(F(self.field_name), "->", Value(value))
+
+
+class AnswerValueOrdering(CalumaOrdering):
```
open-dynaMIX: This should be in the `form` app, as `core` should never depend on `form`.

winged:
> Fair enough @ recursive case search! smiley
> 
> Concerning the legacy `orderBy`. Aren't those directly on `allWorkItems` and `allCases`?

They're almost everywhere

> 
> If we had a version with the `filter: { orderBy: "..." }` everything is good +1

No, `foo(filter: [{orderBy..}])` should never exist. If you want the legacy orderBy, go with `foo(orderBy:...)`.

I've explicitly removed them from the `filter:[]` type, as I want this to only do the filtering, not ordering

caluma/core/ordering.py
winged: Right you are! Fixed

Issue #729, opened 2019-10-14
Show old values of a dynamic data source


JohnL17:
I try to add a new dynamic data source which makes the last n documents selectable for document referencing.

The problem: if the selected document drops out of the last n documents (as example when a new document is created), the value isn't shown in the dynamic question anymore.

The solution: give along the answer object to the data source in caluma, so we can read and display the value.

I think this should be discussed before implementation, so we can assure that this is the best way.

JohnL17:
As disscussed with @open-dynaMIX and @kaldras I will add the answer object as a parameter to the data source's get_data method in caluma. Then I'll pass the answer object in the graphql query so the data source has access to it.

PR #732, opened 2019-10-15
feat(data_source): add the answer id to the data source

JohnL17:
This pr references issue #729

### DIFF
--- a/caluma/data_source/data_source_handlers.py
+++ b/caluma/data_source/data_source_handlers.py
@@ -57,12 +57,12 @@ def get_data_sources(dic=False):
     ]
 
 
-def get_data_source_data(info, name):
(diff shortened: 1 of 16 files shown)


czosel:
The commit message should probably also include an explanation how to migrate an existing app.

winged:
> The commit message should probably also include an explanation how to migrate an existing app.

I'm wondering if we should start working on an "upgrading" guide generally. To explain to users what needs to change in major version bumps. (Might be something for a separate PR however)

czosel:
Good point. What we're doing in ember-caluma is just use the commit messages, at least for a first draft - if further info is needed, the release notes can be edited directly on the GitHub releases page. The only downside of this approach is that the release notes are not directly version-controlled.

PR #736, opened 2019-10-17
chore(reuse): switch to AGPL-3.0-or-later, add reuse-tool

winged:
BREAKING CHANGE: switch to AGPL-3.0-or-later

* Initialize reuse-tool

* Add AGPL-3.0-or-later via dep5 for caluma/*.py

* Add CC0-1.0 via dep5 for metadata/config/info files

  - Files that are below the threshold of originality

* Add reuse lint to travis and pre-commit

(Followup from @ganwell's fork as he's on vacation)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,9 @@ repos:
       language: system
       entry: flake8
       types: [python]
+    - id: reuse
+      name: reuse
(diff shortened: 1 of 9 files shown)


winged:
Waiting for #732 as it will be backwards-incompatible as well to avoid unnecessary version bumping

Issue #731, opened 2019-10-15
Document validation ignores FormQuestions


anehx:
The document validation in the caluma service does not work the same way as in `ember-caluma`. The caluma service iterates over all questions (including questions of form questions) and then validates them. This is wrong, since the respective form question could be hidden or not required, which is completely ignored here:

https://github.com/projectcaluma/caluma/blob/b53cb4d360886dd9335b77e86266cda0e8f7c8ac/caluma/form/validators.py#L177-L298

This should work more like [the implementation in EC](https://github.com/projectcaluma/ember-caluma/blob/395e4e4f162dbbd5cd28e2acddd724aadf21aaa7/addon/lib/field.js#L260-L310)...

winged:
Hidden I understand, but if a question is not required, the answer should still be validated IF it's given. Correct?

PR #747, opened 2019-10-21
Fix 731 form validation

winged:
We should only validate answers that are visible.

### DIFF
--- a/caluma/form/tests/test_validators.py
+++ b/caluma/form/tests/test_validators.py
@@ -364,3 +364,45 @@ def test_validate_hidden_field(
     else:
         with pytest.raises(ValidationError):
             DocumentValidator().validate(document, info)
+
+
+@pytest.mark.parametrize("hide_formquestion", [True, False])
+def test_validate_hidden_subform(
+    db,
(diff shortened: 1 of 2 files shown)


winged:
Fixes #731

caluma/form/tests/test_validators.py
```diff
@@ -343,3 +343,24 @@ def test_validate_required_integer_0(
...
+    document = document_factory(form=form_question.form)
+
+    # answer is out of validation range so should fail
+    answer_factory(question=question, document=document, value=4)
+
+    if question.is_hidden == "true":
+        assert DocumentValidator().validate(document, info) is None
+    else:
+        with pytest.raises(ValidationError):
+            DocumentValidator().validate(document, info)
```
anehx: I think we should add a test for validating subforms when hidden...

winged: This does exactly that. Or do you mean yet another level of subforms?

winged: Nevermind :man_facepalming:

PR #749, opened 2019-10-22
docs: explain license choice

winged:


README.md
```diff
@@ -54,7 +54,39 @@ an example.
...
+
+What it does is that it gives us the motivation to use the official interfaces
+that Caluma provides. We want to keep clean interfaces, and want to motivate
+other users to do the same, despite that it would sometimes be easier to hack
+into the code directly to make it do what you want.
+
+It also protects the project against a large-scale SaaS provider "taking"
+Caluma, modifying it and providing an enhanced, incompatible version of it,
+making money off of it without giving back to the community. This is also the
+main reason behind the license choice.
```
kaldras: Is it the main reason? I believe the main reason was because we as an Open Source company believe all the changes you do should be given back to the public as well. But maybe we should just not state "a main" reason and just a list of reasons which influenced us on the decision?

winged: Fair point, removed the sentence and added a link to "saas loophole" explanation instead

PR #750, opened 2019-10-22
fix(data_source): revert dynamic data source

JohnL17:


winged:
Why are we reverting this?

winged:
OK, so after offline discussion, here's the reasoning:

Doing the API this way breaks caching of the form client-side, as the form now depends on the document it is being used for. This is indeed not a good idea, so we should revert it.

There's a follow-up PR coming to modify validation of dynamic choice questions so a previously-saved answer will be accepted as valid, as long as it doesn't change.

Thus, I figure we can merge this (Note - backwards incompatible if you used 2.0.0, but not if you used 1.x)

PR #758, opened 2019-10-25
fix(ordering): fix ordering by question value

winged:
Ordering by answer value only worked directly on documents,
as the order-by value was joined naively to the outer "pk"
instead of the pk of the document table.

This fixes the problem so that everything that's ordered
by an answer value is now working correctly


fkm:
I can't judge the PR. But thank you! :sunglasses:

PR #762, opened 2019-10-28
feat(data_source): implement dynamic data source

JohnL17:
BREAKING_CHANGE: instead of the `validate` property the validation
of the answer value is done by the new implemented `validate_answer_value`.

PR #763, opened 2019-10-28
New major release: Lots of bug fixes, features, License: GPLv3

winged:


winged:
OK #762 seems to be almost ready, so let's wait for it to be merged

winged:
OK still blocking, #762 introduced some problems that we need to fix first

anehx:
Also, wait for `ember-caluma` :exclamation:

Issue #751, opened 2019-10-22
License Change: GPLv3


winged:
After the change of the license last week from MIT to AGPLv3, we received some valued feedback. For this reason, we're now opening the discussion to make a small step back and use the "normal" GPLv3 instead.

## Why we chose the AGPL

We as an organisation have a strong opinion towards open source and free software. We want to live the philosophy as much as we can.

The AGPL is a rather strict license. In addition to the GNU GPL, you are
required to provide your changes in the source code not only to the "direct"
user, but to users accessing the software via a network as well.

Or as is [explained here](https://www.synopsys.com/blogs/software-security/saas-companies-open-source-risk/):
> SaaS companies are more free to use GPL licensed-code because the obligations of the GPL are triggered upon software distribution and SaaS software by nature is not distributed. (This is often referred to as the GPL “SaaS loophole.”)

This allows such companies to "take" GPL code, modify it and provide it to their users without the actual users benefitting from the software's license, as is the GPL's intention.

Our goal is to build a strong community around Caluma, and we feel that we should do everything to help this goal come along. Part of it is having a good license that incentivizes people to contribute back.

## Feedback

The feedback we received after the change to AGPL was quite valuable. We learned of a few situations that we didn't consider before that maybe we should have.
The move to AGPL was also discussed "behind closed doors", opposing the open culture we want to nourish. We try to right this by documenting the reasons in #749.

I'll address the feedback we got in the following points

### Licensing of extension points

Due to the strictness of the AGPL, one could consider that extensions, such as visibility and permissions would be "infected" by the AGPL and would thus need to be shared to the end users. This would be problematic as it could expose internal workings of IT infrastructure, or logic that is definitely not intended for the public.

While it is true that it is python code called directly from Caluma, we still consider those modules to be configuration. A config file can easily be excluded from such license claims - you wouldn't share your `/etc/shadow` for example.

Other projects, such as Drupal, [do the same in practice](https://groups.drupal.org/node/17555):
> It's a little bit of a gray area, but the general guideline we worked out is that despite settings.php being technically code, what's actually stored there is data, not anything complex enough for copyright to even apply. So if someone tried to invoke GPL or AGPL to get access to a site's settings.php file, which has database info in it, we'd be justified in brushing them off.

### Straigh-out blacklisting of AGPL code

[...]

PR #767, opened 2019-10-30
chore(license): switch main license to GPL-3.0-or-later

ganwell:
Based on the discussion in issue #751, we reevaluated our goals. We want

-  diverse uses and integration of Caluma
-  Caluma to remain public as Free Software
-  an active community
-  to guarantee mutual benefit to our contributors

We found that the definitions in AGPL aren't as clearcut as we wished; therefore, we want to switch to GPL-3.0-or-later. GPL will support all of our goals, with a minor draw-back: Not every last modification is guaranteed to end up in public. Which isn't that bad and it resolves all discussion about configuration.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -31,4 +31,4 @@ License: CC0-1.0
 # Source code of caluma
 Files: caluma/*.py docs/*
 Copyright: 2019 Adfinis SyGroup AG <info@adfinis-sygroup.ch>
-License: AGPL-3.0-or-later
+License: GPL-3.0-or-later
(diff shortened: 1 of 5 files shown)


anehx:
Looks good, there's only one unanswered question: what exact license are we using for the frontend packages?

ganwell:
@anehx GPL-3.0-or-later is the only correct one. Its the official SPDX identifier.

ganwell:
> The notation is a bit confusing, we should decide which one of those we use:
> 
> * GPL-3.0-or-later
> * GPLv3+
> * GPL-3.0

@anehx GPL-3.0-or-later is the only correct one. Its the official SPDX identifier. I can't change that pypi doesn't use the correct one and that the license text itself isn't consistent. I changed the badge, though. Let me know if I missed an occurrence.

ganwell:
And GitHub does everything wrong, when it comes to licenses:

1. They use the wrong identifiers
2. They don't accept the SPDX license texts
3. They don't scan the right files

Many people opened issues, for some reason, they don't want to fix the problem.

PR #768, opened 2019-10-30
fix(questions): do not return questions multiple times

winged:
In the allQuestions query, do not return the same question multiple
times if it is used in more than one form.
This evaded noticing until now because we usually use the `slugs=[]..`
filter, which does the `distinct()` call for us.


fkm:
I'll check tomorrow. 

But thank you!

fkm:
Doesn't seem to help. But I'll test again this afternoon.

PR #774, opened 2019-11-01
fix(OIDC): raise a 401 if user is not logged in

JohnL17:


open-dynaMIX:
I was not asked for a review, but give it anyway :)

I think it would be cleaner to handle this before the `OIDCClient()` is instantiated. i.e. [here](https://github.com/projectcaluma/caluma/blob/master/caluma/user/views.py#L108).

PR #775, opened 2019-11-04
chore(reuse): Use GFDL license for documentation

ganwell:
GPL cannot apply to documentation since there are no runtime artifacts

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -11,9 +11,7 @@ Files:
     .pre-commit-config.yaml
     .flake8
     .travis.yml
-    CONTRIBUTING.md
     Dockerfile
-    README.md
     commit-msg
     commitlint.config.js
     docker-compose.override.yml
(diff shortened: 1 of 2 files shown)


.reuse/dep5
```diff
@@ -29,6 +29,11 @@ Copyright: 2019 Adfinis SyGroup AG <info@adfinis-sygroup.ch>
 License: CC0-1.0
 
 # Source code of caluma
-Files: caluma/*.py docs/*
+Files: caluma/*.py
 Copyright: 2019 Adfinis SyGroup AG <info@adfinis-sygroup.ch>
 License: GPL-3.0-or-later
+
+# Documentation of caluma
+Files: docs/*
```
winged: Needs to include the few `*.md` files in the toplevel folder (`CONTRIBUTING.md`, `README.md`)

ganwell: Added these two. Thanks for reviewing.

PR #784, opened 2019-11-07
Fix evaluate is_hidden

### LINKED ISSUE #781: Form validation: isHidden behaves differently than in the frontend
In the frontend implementation, we don't even evaluate a question's `isHidden` JEXL if all the fields it depends on are hidden themselves.

Example:

Question 1 is always visible
Question 2 is hidden when Q1 == "foo"
Question 3 is required and hidden when Q2 == null

This currently leads to the frontend hiding Q3 and not validating it's requiredness (because Q2, it's only dependant, is hidden), while the backend would think that it's displayed and validate it's requiredness.

winged:
fixes #781

### DIFF
--- a/caluma/form/jexl.py
+++ b/caluma/form/jexl.py
@@ -1,7 +1,9 @@
+from functools import partial
+
 from pyjexl.analysis import ValidatingAnalyzer
 from pyjexl.evaluator import Context
 
-from ..core.jexl import JEXL
+from ..core.jexl import JEXL, ExtractTransformSubjectAnalyzer
 
 
 class QuestionMissing(Exception):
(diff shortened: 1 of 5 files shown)


winged:
@open-dynaMIX I'd be glad if you could have another look. Needed quite a bit of refactoring to get right

caluma/form/jexl.py
```diff
@@ -17,11 +19,19 @@ def visit_Transform(self, transform):
...
-        answer_by_question = answer_by_question if answer_by_question else {}
+    def __init__(self, validation_context=None, **kwargs):
+        answer_by_question = validation_context["answers"] if validation_context else {}
+
         super().__init__(**kwargs)
 
-        self.context = Context({"form": form})
+        context_data = None
+        if validation_context:
+            context_data = {**validation_context}
```
open-dynaMIX: Isn't `validation_context` a dict already?

winged: Yes. But I'm modifying it a few lines further down, which I don't want to propagate to the caller. Thus, a copy is created

open-dynaMIX: :+1:

PR #791, opened 2019-11-15
feat(form): allow Answers with empty value

### LINKED ISSUE #790: Allow empty answers
We want to support having `Answers` without a set value.

This will be particularly usefull when working with historical records.

open-dynaMIX:
This commit adds the possibility to save/query Answers with `null`
values.

It also adds a new lookup mode `ISNULL` to the `HasAnswerFilter`. If
this lookup mode is used, the provided `value` (if any) is ignored.

Since graphql-core doesn't support `null` literals yet, the way for setting
the value of an Answer to `None` is to omit the `value` field
altogether.

Closes #790 and #800 

BREAKING CHANGE: `Answers` are now allowed to have `null` values.

### DIFF
--- a/caluma/form/filters.py
+++ b/caluma/form/filters.py
@@ -108,6 +108,7 @@ class AnswerLookupMode(Enum):
     CONTAINS = "contains"
     ICONTAINS = "icontains"
     INTERSECTS = "intersects"
+    ISNULL = "isnull"
 
     GTE = "gte"
     GT = "gt"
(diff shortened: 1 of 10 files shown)


czosel:
Why is this a breaking change? Aren’t we just being more permissive?

open-dynaMIX:
> Why is this a breaking change? Aren’t we just being more permissive?

I'm not 100% sure about that.

You could've relied on the backend to prevent `null` values in `Answers`. This won't work anymore.

To be safe, I'd consider all changes on the optionality of existing fields as potentially breaking.

czosel:
In general this sounds reasonable, in this specific case I'm not so sure, since most of the code that relies on this is in ember-caluma anyway. I wonder what @kaldras and @anehx say about it?

open-dynaMIX:
We claim to be webservice agnostic. So I don't think ember-caluma should have anything to do with the decision if this is a breaking change.

Of course I'm aware of the fact that noone is using caluma without ember-caluma atm ;)

winged:
> In general this sounds reasonable, in this specific case I'm not so sure, since most of the code that relies on this is in ember-caluma anyway. I wonder what @kaldras and @anehx say about it?

It's also (and mainly) about the type system. The value types changed from `String!` to `String` (where the missing `!` denotes nullability). This by itself implies a breaking change, as you need to change the types in the mutations/queries as well.

Issue #793, opened 2019-11-15
Rethink caching strategy for datasources


kaldras:
The language code is not within the cache-key. This could lead to problems when caching is activated and therefore should be looked at again. For example:

The datasource translates its own values. Fetch the values in "de", cached for 30 seconds and switch to "fr" within these 30 seconds. Caluma will then return the cached "de" values since the same key for this datasource is used.

open-dynaMIX:
I assume there is another problem, as the datasource caches the whole return value of `get_data()` and not just the value with it's translated label. So the translation should not be affected by the caching at all.

Translation must not happen in the DataSource. `get_data()` must return all available translations per option. Translation is done when fetching the data from it and follows the same logic as the `LocalizedField()`.

open-dynaMIX:
Resolved by #797

PR #797, opened 2019-11-18
fix(data_source): use LocalizedValue for tranlations

### LINKED ISSUE #792: Consistent translation logic in datasource
Currently the datasource handles multilingual values using custom logic (see `caluma/core/utils.py(5)translate_value()`) which can lead to unexpected results. It would probably be better to use the same translation logic used of by the `localizedFields`.

open-dynaMIX:
This commit gets rif of our own (faulty) translation logic and uses
`LocalizedValue` instead.

Closes #792 and #793


docs/extending.md
```diff
@@ -195,6 +195,9 @@ this value will also be used as label.
 
 For the label, it's possible to use a dict with translated values.
 
+Beware, translating the values inside of the `get_data`-method and returning just
```
winged:
```suggestion
Note: If the labels returned from`get_data()` depend on the current user's language, you need to return a `dict` with the language code as keys instead of translating the value yourself. Returning already 
```

caluma/form/format_validators.py
```diff
@@ -22,18 +21,24 @@ class BaseFormatValidator:
...
+
     def __init__(self):
         if not all(
             [self.slug, self.regex, self.name, self.error_msg]
         ):  # pragma: no cover
             raise NotImplementedError("Missing properties!")
+        if not isinstance(self.name, dict) and isinstance(
+            self.error, dict
+        ):  # pragma: no cover
+            raise ValueError("`name` and `error_msg` must be of type `dict`!")
```
winged: `LocalizedValue` can handle passing it a simple string, so I'm not sure we need to enforce this. Would simplify the API a bit for monolingual projects.

```python
In [1]: from localized_fields.value import LocalizedValue
In [2]: LocalizedValue("hello world")
Out[2]: LocalizedValue<{'de': 'hello world', 'fr': None, 'it': None, 'en': None}> 0x140401073378760
```

docs/extending.md
open-dynaMIX: Done!

caluma/form/format_validators.py
open-dynaMIX: Done!

PR #804, opened 2019-11-19
docs: explain how to use pipenv

ganwell:


ganwell:
Works but, not needed.

PR #805, opened 2019-11-19
docs: explain how to use pipenv

ganwell:
I first thought it makes no sense because we use docker, but I needed to install the packages for my editor. So its nice to have the commands in the documentation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,9 +71,20 @@ pre-commit install
 ```
 
 ### Setup commit-msg hook
+
 If you want to have your commit message automatically linted, execute below commands:
 
 ```bash
 npm install @commitlint/{config-conventional,cli}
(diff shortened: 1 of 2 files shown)


ganwell:
Turns out this is even more useful. pipenv has support for .env variables and for my current task I have to compare different dependencies. With pipenv I can easly switch between them. Running multiple containers is a pain.

CONTRIBUTING.md
```diff
@@ -71,9 +71,21 @@ pre-commit install
...
+
+### Using pipenv
+
+Docker is the way to run and test caluma, but you can install the requirements
+for your IDE/editor using pipenv:
+
+```bash
+pipenv --three install -r requirements.txt
+pipenv install -d -r requirements-dev.txt
+vim
```
open-dynaMIX: Starting vim is not really needed for using pipenv, is it? :)

PR #809, opened 2019-11-20
fix(history): optimize `historical_qs_as_of`

open-dynaMIX:
This commit optimizes `historical_qs_as_of()` so that the amount of
queries is significatly reduced and it returns a `Queryset()` instead of a
list.

### DIFF
--- a/caluma/form/historical_schema.py
+++ b/caluma/form/historical_schema.py
@@ -24,18 +24,18 @@ from .storage_clients import client
 def historical_qs_as_of(queryset, date, pk_attr):
     """Get history revision as of `date` for queryset.
 
-    :param queryset: history qs
+    :param queryset: history Queryset()
     :param date: aware datetime()
     :param pk_attr: str (pk field name)
+    :return: Queryset()
     """
-    # TODO: This could be optimised with some sql magic in order to return a queryset.
     # This could become unnecessary as soon as
     # https://github.com/treyhunner/django-simple-history/issues/397 is resolved.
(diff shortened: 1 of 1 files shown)


caluma/form/historical_schema.py
```diff
@@ -24,18 +24,25 @@
...
+    # TODO: This could be optimised with some sql magic.
     # This could become unnecessary as soon as
     # https://github.com/treyhunner/django-simple-history/issues/397 is resolved.
     queryset = queryset.filter(history_date__lte=date)
     for original_pk in set(queryset.values_list(pk_attr, flat=True)):
-        changes = queryset.filter(**{pk_attr: original_pk})
-        last_change = changes.latest("history_date")
-        yield last_change
+        # exclude all records referencing the same original_pk except the most recent
+        queryset = queryset.exclude(
```
winged: I don't like this. It creates a whole bunch of nested queries, inverting each other. For each `original_pk`, you will have an additional nesting level, as you use the (updated) queryset variable to filter for the next `pk`.

I think we'd be better off by firing off an intermediate query.. First, get the newest versions using an aggregate (`Max('history_date')`), group by `pk_attr`. The resulting pairs can then be used to build a query using `Q` objects.

open-dynaMIX:
@winged I've tried to stop you from reviewing this, the problems are obvious.

I found a more elegant solution for this. The reason I didn't add a test for this is, that `test_historical_table_answer()`  already tests this.

PR #813, opened 2019-11-22
feat(permissions): make permission decorators chainable

sbor23:
Before, the `has_permission` and `has_object_permission` decorators
would need a separate function for each Mutation.
Now, the same function can be decorated for multiple mutations, reducing
the amount of code if some mutations share the same permission logic.

Example:
```python
class CustomPermission(BasePermission):

    @has_permission(FirstMutation)
    @has_permission(SecondMutation)
    def has_permission_for_both(self, mutation, info):
        return False
```

### DIFF
--- a/caluma/core/permissions.py
+++ b/caluma/core/permissions.py
@@ -1,5 +1,4 @@
 import inspect
-from functools import wraps
 
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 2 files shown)


sbor23:
On a side-note, the same thing could be done for visibilities and validations, but that wasn't necessary in our use-case yet. And it's debatable if there's as much sharable code for those things as there is for permissions.

Issue #814, opened 2019-11-22
Tracking issue for graphql 3.0


ganwell:
- [x] !! Sync everything: there is #1413, which should be force-pushed to graphene_django_v3 #1438
- [x] Make pytests order-independent in master
- [x] Lower coverage (done in #1413)
- [x] Enable xfaillist #1413
- [x] Fix all pytests in graphene_django_v3 branch (pytest-xfaillist)
- [x] Upstream bugs (if found)
- [x] Disable xfaillist
- [x] Increase coverage
- [x] Fix coverage
- [x] merge master to graphene_django_v3 contantly 
- [x] Merge to master

sbor23:
@ganwell would you mind updating the checklist?

ganwell:
@sbor23 Ok?

PR #815, opened 2019-11-22
add pipenv workflow for local development

ganwell:
This kind of just happend:

1. I needed a way to install the environment for my editor, so that completion works, pipenv was very easy to use

2. Then managing the dependencies and running multiple databases for #814 was a pain, so I started to use pipenv for development

3. Then @sbor23 asked how to do local development and I defined a proper workflow

4. Then I realized that there are certain common mistakes, so I coded the setup of pipenv in setup.py

5. And after all this, it made sense to also test in the pipenv, so there are no surprises for developers deciding do local development

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+os: "linux"
+
 python:
   - "3.6"
 
@@ -25,20 +27,27 @@ before_install:
 
   # set UID to run docker service with
   - echo "UID=$(id --user)" > .env
+jobs:
+  include:
+    - env: BUILD_TYPE=standard
+      install:
+        - docker-compose up -d --build
+        - npm install @commitlint/{config-conventional,travis-cli}
 
-install:
-  - docker-compose up -d --build
-  - npm install @commitlint/{config-conventional,travis-cli}
-
-script:
-  - docker-compose exec caluma black --check .
-  - docker-compose exec caluma flake8
-  - ./node_modules/.bin/commitlint-travis
-  - docker-compose exec caluma ./manage.py makemigrations --check --dry-run --no-input
-  - docker-compose exec caluma pytest --no-cov-on-fail --cov --create-db -vv
-  - docker-compose exec caluma reuse lint
-after_success:
-  - git config --global user.name "semantic-release (via TravisCI)"
-  - git config --global user.email "semantic-release@travis"
-  - pip install python-semantic-release
-  - semantic-release publish
+      script:
+        - docker-compose exec caluma black --check .
+        - docker-compose exec caluma flake8
+        - ./node_modules/.bin/commitlint-travis
+        - docker-compose exec caluma ./manage.py makemigrations --check --dry-run --no-input
+        - docker-compose exec caluma pytest --no-cov-on-fail --cov --create-db -vv
+        - docker-compose exec caluma reuse lint
+      after_success:
+        - git config --global user.name "semantic-release (via TravisCI)"
+        - git config --global user.email "semantic-release@travis"
+        - pip install python-semantic-release
+        - semantic-release publish
+    - env:
+        BUILD_TYPE=pipenv
+        PIPENV_IGNORE_VIRTUALENVS=1
+      install: python3 setup.py pipenv
+      script: pipenv run pytest
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,46 @@
 """Setuptools package definition."""
 
+import distutils.cmd
+import os
+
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -25,20 +27,27 @@ before_install:
...
+      after_success:
+        - git config --global user.name "semantic-release (via TravisCI)"
+        - git config --global user.email "semantic-release@travis"
+        - pip install python-semantic-release
+        - semantic-release publish
+    - env:
+        BUILD_TYPE=pipenv
+        PIPENV_IGNORE_VIRTUALENVS=1
+      install: python3 setup.py pipenv
+      script: pipenv run pytest
```
sbor23: Not entirely sure what that does.
Why would we need another CI job for running pytest?
This is already done in the first job using docker-compose, no?

sbor23:
Very cool, didn't know you could so easily script the projects setup using the setup.py!
Does pyenv need any extension for handling python versions or does it all out of the box?

This looks like a very valuable addition for caluma noobs. :)

ganwell:
> Very cool, didn't know you could so easily script the projects setup using the setup.py!
> Does pyenv need any extension for handling python versions or does it all out of the box?

Well its the Kenneth Reitz Philosophy it does only one thing: creating reproducible virtualenvs. If it doesn't find the requested python version, it will just tell you that you have to install it. You can either install it from system ie `dnf install python36` or use another tool.

> This looks like a very valuable addition for caluma noobs. :)

.travis.yml
ganwell: * If the python version is not set right, the test might fail. 
* If the code in setup.py somehow breaks
** because pipenv had changed its interface, the test might fail
** because distribute or setuptools changed its interface, the test might fail

There are a lot of ways the changes in this PR might fail. If we test it, we can fix it. If we don't test it, a user or contributor might have an issue and he won't know if he has done something wrong or if it is broken. We don't want to deliver broken things to users!

setup.py
```diff
@@ -1,12 +1,46 @@
...
     exec(fp.read(), version)
 
+pipenv_setup = """
+echo UID=$(id --user) > .env
+echo ENV=dev >> .env
+docker-compose up -d db
+rm -f Pipfile*
+touch Pipfile
+pipenv install --python 3.6 -r requirements.txt
+pipenv install -d -r requirements-dev.txt
```
sbor23: Doing this will will flag all the deps in `requirements.txt` as dev-dependencies as well, because we have the line `-r requirements.txt` in there.
Why would you want to remove the Pipfile and Pipefile.lock during the setup?
If you want to check if the Pipfile.lock is up-to-date, there is the `--deploy` flag (or `--sync`?) that checks that, no?
https://pipenv.readthedocs.io/en/latest/advanced/#using-pipenv-for-deployments

ganwell: > Doing this will will flag all the deps in `requirements.txt` as dev-dependencies as well, because we have the line `-r requirements.txt` in there.

I am aware of how to use pipenv correctly, we just can't do it (yet?). For new project I use pipenv in its full glory.

It does't matter that the non-dev dependencies are added to the dev-dependencies, since you can't install only the dev-dependencies with pipenv anyways. This is the best way to bootstrap the Pipfile from requirements*.txt.

If you do

```bash
pipenv --rm
pipenv sync
```
you get only the non-dev-requirements.

```bash
pipenv --rm
pipenv sync -d
```
you get the dev- and the non-dev-requirement.

> Why would you want to remove the Pipfile and Pipfile.lock during the setup?

Well, I don't want to get rid of the requirements*.txt file (yet?), a lot of tooling expects the requirements*.txt.

* Because we still have the requirements.txt as primary definition, we can't use pipenv for deployments
* Because we use docker for deployment, we don't want to use pipenv for deployments

> If you want to check if the Pipfile.lock is up-to-date, there is the `--deploy` flag (or `--sync`?) that checks that, no?
> https://pipenv.readthedocs.io/en/latest/advanced/#using-pipenv-for-deployments

Well the idea is to call ./setup.py pipenv only once for your local repository. From then on you use only pipenv commands. Only if the requirements*.txt have changed you would call ./setup.py pipenv again. Since the Pipfile is meant to be the primary source, it doesn't update constaints from requirements.txt if Pipfile exists, so we have to delete the Pipfile*, in order to transfer the changes to the Pipfile*.

sbor23: Ah I see, makes sense now. :)
Just saw that Pipfile* is in the .gitignore, so it's only used as an alternative workflow.
I like!

PR #819, opened 2019-11-28
fix(validation): Validate dependencies correctly

winged:
When validating, we must only consider questions directly within a given
form, not all below.  This caused answers to be validated that should
not have been.


winged:
> Where's the test? scream

coming right up

> Also, your commit title should be lowercase to match the convention

Already fix'd

anehx:
It'd be good if @open-dynaMIX would take a look at it as well

PR #840, opened 2019-12-10
feat(pytest): run pytest in parallel with xdist

ganwell:
pytest-django creates a database per worker by default: https://git.io/Je9cn


ganwell:
If you think we should not run this be default, here are the arguments:

1. If we don't run it be default people might break it
2. It works very well, I am excited
3. We can always change it

winged:
Oh, just thought of this after merging: This might need some mention in the docs for when you need to update snapshots

PR #837, opened 2019-12-06
chore(deps): update to graphene-django v2.7.1

open-dynaMIX:
The update to graphene-django v2.7.1 needs some changes on some classes
we subclassed from it.

This commit implements them and updates the version in requirements.txt.

PR #841, opened 2019-12-10
chore(deps): update to graphene-django v2.7.1

ganwell:
* Remove insignificant performance hack. Performance profiling is known to have a variance of 20% - 40%, only improvements of %50+ should be considered, especially when hacking the internals
of another library.

* Update to graphene v2.0.1 since graphene-django v2.7.1 depends on that version.

* Adapt to refactoring of DjangoConnectionField. We inject our version of
  connection_from_list, connection_from_list_slice.

Replaces #837

### DIFF
--- a/caluma/core/filters.py
+++ b/caluma/core/filters.py
@@ -507,50 +507,6 @@ class DjangoFilterConnectionField(
         queryset.query.select_related = default_queryset.query.select_related
         return queryset
 
-    @classmethod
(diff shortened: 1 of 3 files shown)


ganwell:
@open-dynaMIX sorry of making an almost identical PR, but these changes affect graphen-django 3.0 too and I wanted to dig through it myself.

PR #842, opened 2019-12-11
chore(deps): rename fsfs-reuse to reuse

ganwell:


winged:
What's the difference? Why the change?

ganwell:
They renamed the package, reuse was not available, then they could clear that problem with the pypi managers and now its called reuse.

PR #846, opened 2019-12-13
chore: include minimal form fixture

sbor23:
The form fixture represents a minimal form as it's used by the
ember-caluma addon. It includes all question types apart from dynamic
questions.
This is intended to be development aid, reducing the burden to have setup some test data.
It can be loaded into the local project by running `python manage.py
loaddata caluma/fixtures/form-fixture.json`.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -23,6 +23,7 @@ Files:
     setup.cfg
     setup.py
     uwsgi.ini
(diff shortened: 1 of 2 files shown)


winged:
You should add the new JSON file to `.reuse/dep5` in the " Files below the threshold of originality" list. Otherwise, LGTM

PR #853, opened 2019-12-18
fix(release): version bump

winged:
Try to get semantic-release back to working state


winged:
I hope this fixes the mess that somehow happened in the release branch

PR #871, opened 2020-01-07
fix(validation): correctly handle indirectly hidden questions

winged:
Questions can not only be hidden by evaluating their is_hidden expression
but also when they're part of a subform, where the containing form question
is hidden. We need to check for that case as well, which didn't happen till now.

### DIFF
--- a/caluma/caluma_form/jexl.py
+++ b/caluma/caluma_form/jexl.py
@@ -1,9 +1,12 @@
+from collections import defaultdict
 from functools import partial
 
+from django.db.models import Q
 from pyjexl.analysis import ValidatingAnalyzer
 from pyjexl.evaluator import Context
 
 from ..caluma_core.jexl import JEXL, ExtractTransformSubjectAnalyzer
+from .models import Question
 
 
 class QuestionMissing(Exception):
@@ -21,6 +24,13 @@ class QuestionValidatingAnalyzer(ValidatingAnalyzer):
 class QuestionJexl(JEXL):
     def __init__(self, validation_context=None, **kwargs):
         answer_by_question = validation_context["answers"] if validation_context else {}
+
+        if validation_context:
+            if "jexl_cache" not in validation_context:
+                validation_context["jexl_cache"] = defaultdict(dict)
+            self._cache = validation_context["jexl_cache"]
+        else:
+            self._cache = defaultdict(dict)
 
         super().__init__(**kwargs)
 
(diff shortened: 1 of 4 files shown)


caluma/form/jexl.py
```diff
@@ -85,8 +125,20 @@ def is_hidden(self, question):
         all_deps_hidden = bool(deps) and all(
             self.is_hidden(self._question(dep)) for dep in deps
         )
-
-        return all_deps_hidden or self.evaluate(question.is_hidden)
+        if all_deps_hidden:
+            self._cache_is_hidden[question.pk] = True
+            return True
+
+        # Also check if the quesiton is hidden indirectly,
```
sbor23: Typo: question

caluma/form/tests/test_jexl.py
```diff
@@ -69,3 +72,85 @@ def test_all_deps_hidden(db, form, document_factory, form_question_factory):
...
+    document_factory,
+    answer_factory,
+    info,
+    fq_is_hidden,
+):
+    # Questions can not only be hidden by evaluating their is_hidden expression
+    # but also if they're part of a subform, where the containing form question
+    # is hidden.
+    #
+    # Showcase form to demonstrate this (f: form, q:question, d:document,a:answer):
```
sbor23:
```suggestion
    # Showcase form to demonstrate this (f: form, q:question, d:document, a:answer):
```

caluma/caluma_form/jexl.py
```diff
@@ -67,12 +70,49 @@ def _question(self, slug):
...
+        to the given question, and checks if they're hidden.
+
+        If all subforms are hidden where the question shows up,
+        the question shall be hidden as well.
+        """
+
+        doc_form = self.context.get("document").form
+
+        # first, build a list of paths to the question
+        def find_path(fq, acc):
```
czosel: nitpick: I don't know what's meant by `fq` - maybe rather use a longer variable name?

caluma/caluma_form/jexl.py
```diff
@@ -67,12 +70,49 @@ def _question(self, slug):
...
+            parent_fqs = Question.objects.filter(
+                Q(type=Question.TYPE_FORM) | Q(type=Question.TYPE_TABLE),
+                Q(sub_form__questions__pk=fq) | Q(row_form__questions__pk=fq),
+            )
+            res = []
+            if doc_form in fq.forms.all():
+                # found a path - add it to our result list
+                res.append(acc + [fq])
+            for pfq in parent_fqs:
```
czosel: same here

caluma/caluma_form/validators.py
```diff
@@ -195,19 +197,88 @@ def validate(self, document, info, validation_context=None, **kwargs):
...
+
+        q_jexl = jexl.QuestionJexl(validation_context)
+        for question in validation_context["questions"].values():
+            try:
+                is_hidden = q_jexl.is_hidden(question)
+
+                if not is_hidden:
+                    visible_questions.append(question.slug)
+                    if question.type == Question.TYPE_FORM:
+                        # form questions's answers are still in the top level document
```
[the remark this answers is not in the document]
czosel: :+1:

caluma/caluma_form/jexl.py
winged: `question` would be the obvious choice, but that's already present in the outer context. I'll think of something

caluma/caluma_form/jexl.py
```diff
@@ -67,12 +77,59 @@ def _question(self, slug):
             )
         return question
 
+    def _paths_to_question(self, question, doc_form):
+        """Return all paths from the doc form to the given question."""
+        cache = self._cache["containers_hidden_paths"]
+        if question.slug in cache:
+            return cache[question.slug]
```
czosel: The return value of the function seems to depend on `doc_form` but the caching is only based on `question`. Doesn't this cause issues?

winged: There can only be one answer per question in the whole document structure. The only place where that's not true is in table rows. But the "containing question" thing falls down there anyway, as you cannot refer to a specific row anyhow, so the expression cannot be correctly evaluated in that case.

czosel: Hmm... isn't `doc_form` always the main form? if so, then adding it to the caching keys should not change the cache hit ratio, right?
Or are you calling this with for table row documents as well?

winged: Changed the code to include `doc_form` in the cache key

caluma/caluma_form/validators.py
```diff
@@ -195,19 +198,89 @@ def validate(self, document, info, validation_context=None, **kwargs):
...
         if not validation_context:
             validation_context = self._validation_context(document)
+        visible_questions = []
+
+        q_jexl = jexl.QuestionJexl(validation_context)
+        for question in validation_context["questions"].values():
+            try:
+                is_hidden = q_jexl.is_hidden(question)
+
+                if not is_hidden:
```
sbor23: You could remove nesting if you do this instead:
```python
if is_hidden:
    continue
# do the validation now ...
```

caluma/caluma_form/validators.py
```diff
@@ -285,67 +358,55 @@ def _validate_required(self, validation_context):  # noqa: C901
...
+                if question.type == Question.TYPE_FORM:
+                    # form questions's answers are still in the top level document
+                    sub_context = {
+                        **validation_context,
+                        "questions": {
+                            q.slug: q for q in question.sub_form.questions.all()
+                        },
+                    }
+                    self._validate_required(sub_context)
+                elif question.type == Question.TYPE_TABLE and is_required:
```
sbor23: Just a question: Why do we validate TableQuestions only if `is_required`, but FormQuestions are validates anyway?

czosel: I think FormQuestions don't have a real required state on their own - it's given by the requiredness of the questions the form contains.

czosel:
Coverage is still lacking: https://travis-ci.com/projectcaluma/caluma/jobs/277914328

caluma/caluma_form/validators.py
sbor23: Ah makes sense, thanks

winged:
> Coverage is still lacking: https://travis-ci.com/projectcaluma/caluma/jobs/277914328

Nice, missing coverage pointed out a refactoring/cleanup bug! (fixed)

czosel:
Awesome! Now let's make sure @open-dynaMIX 's feedback is addressed and then we can get this merged :rocket:

winged:
As discussed, let's merge this first, then do the 5.0 release

PR #885, opened 2020-01-22
chore(release): remove semantic-release from pipeline

open-dynaMIX:
We had too much trouble with automatic releases not detecting breaking
changes. That's why we decided to go back to manual releasing and using
`python-semantic-release` just as support tool for writing release
notes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -68,8 +68,3 @@ omit=
     caluma/wsgi.py
     caluma/caluma_metadata.py
 show_missing = True
-
-[semantic_release]
-version_variable=caluma/caluma_metadata.py:__version__
-branch=release
-upload_to_pypi=False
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -68,8 +68,3 @@ omit=
     caluma/wsgi.py
     caluma/caluma_metadata.py
 show_missing = True
-
-[semantic_release]
```
winged: Shouldn't we keep the settings, so we can use semantic-release to at least help in the process?

open-dynaMIX: In a personal project, I use semantic-release without a config:

```shell
semantic-release changelog --noop --unreleased -D version_source=tag
```
With this, `semantic-release` will just lookup the last tag and go from there. The version variable needs to be changed manually.

winged: Fair enough. Nice work, let's merge it

PR #895, opened 2020-01-29
Validation performance fixes

winged:
This should speed up validation significantly :)

### DIFF
--- a/caluma/caluma_form/structure.py
+++ b/caluma/caluma_form/structure.py
@@ -1,0 +1,192 @@
+"""Hierarchical representation of a document / form."""
+import weakref
+from functools import singledispatch
+from typing import Optional
+
+from .models import Question
+
+
+def object_local_memoise(meth):
+    def new_meth(self, *args, **kwargs):
+        if not hasattr(self, "_memoise"):
+            self._memoise = {}
+
+        key = str([args, kwargs, meth])
+        if key in self._memoise:
+            return self._memoise[key]
+        ret = meth(self, *args, **kwargs)
+        self._memoise[key] = ret
+        return ret
+
+    return new_meth
+
+
+class Element:
+    def __init__(self, parent=None):
+        self._parent = weakref.ref(parent) if parent else None
+
+    def parent(self):
+        return self._parent() if self._parent else None
+
+    def children(self):  # pragma: no cover
+        return []
+
+
+class AnsQuestion(Element):
+    def __init__(self, document, form, question, answer=None, parent=None):
+        super().__init__(parent)
+        self.document = document
+        self.form = form
+        self.question = question
+        self.answer = answer
+
+    @classmethod
+    def factory(cls, document, form, question, answer=None, parent=None):
+        if question.type == Question.TYPE_FORM:
+            return DocForm(
+                document, form=question.sub_form, question=question, parent=parent
+            )
+        elif question.type == Question.TYPE_TABLE:
+            return RowAnsQuestion(
+                document,
+                form=question.row_form,
+                question=question,
+                answer=answer,
+                parent=parent,
+            )
+
+        return AnsQuestion(document, form, question, answer, parent=parent)
+
+    def ans_value(self):
+        if self.answer is None:
+            # no answer object at all - return empty in every case
+            return self.question.empty_value()
+
+        if self.answer.value is not None:
+            return self.answer.value
+
+        elif self.question.type == Question.TYPE_TABLE:
+            return [
+                {cell.question.slug: cell.ans_value() for cell in row.children()}
+                for row in self.children()
+            ]
+        elif self.question.type == Question.TYPE_FORM:
+            # forms never have values themselves
+            return None
+            return [
+                {cell.question.slug: cell.ans_value() for cell in row.children()}
+                for row in self.children()
+            ]
+
+    def __repr__(self):
+        return f"<AnsQuestion question={self.question.slug}, value={self.ans_value()} hidden=({self.question.is_hidden}) req=({self.question.is_required})>"
+
+
+class RowAnsQuestion(AnsQuestion):
+    @object_local_memoise
+    def children(self):
+        if not self.answer:
+            return []
+
+        return [
+            DocForm(
+                row_doc,
+                self.question.row_form,
+                question=self.question,
+                parent=self.parent(),
+            )
+            for row_doc in self.answer.documents.all()
+        ]
+
+
+class DocForm(Element):
+    def __init__(self, document, form, question=None, parent=None):
+        super().__init__(parent)
+        self.document = document
+        self.form = form
+        self.question = question
+        self._ans_questions = None
+        self._sub_forms = None
+
+    def ans_value(self):
+        return None
+
+    def _build_ans_questions(self):
+        if self._ans_questions is None:
+            self._ans_questions = {aq.question.slug: aq for aq in self.children()}
+        if self._sub_forms is None:
+            self._sub_forms = {
+                aq.question.slug: aq
+                for aq in self.children()
+                if aq.question.type == Question.TYPE_FORM
+            }
+
+    def get_ans_question(
+        self, question_slug: str, allow_check_parent: bool = True
+    ) -> Optional[AnsQuestion]:
+
+        self._build_ans_questions()
+
+        ans_question = self._ans_questions.get(question_slug)
+
+        if not ans_question and allow_check_parent:
+            ans_question = (
+                self.parent().get_ans_question(question_slug) if self.parent() else None
+            )
+        if ans_question:
+            return ans_question
+
+        # OK start looking in subforms / row forms below our level.
+        # Since we're looking down, we're disallowing recursing to outer context
+        # to avoid recursing back to where we are
+        for subform in self._sub_forms.values():
+            sub_aq = subform.get_ans_question(question_slug, allow_check_parent=False)
+            if sub_aq:
+                return sub_aq
+        # if we reach this line, we didn't find the question
+        return None
+
+    @object_local_memoise
+    def children(self):
+        answers = {ans.question_id: ans for ans in self.document.answers.all()}
+        return [
+            AnsQuestion.factory(
+                document=self.document,
+                form=self.form,
+                question=question,
+                answer=answers.get(question.slug),
+                parent=self,
+            )
+            for question in self.form.questions.all()
+        ]
+
+    def __repr__(self):
+        q_slug = self.question.slug if self.question else None
+        if q_slug:
+            return f"<DocForm fq={q_slug}, doc={self.document.pk} hidden=({self.question.is_hidden}) req=({self.question.is_required})>"
+
+        return f"<DocForm form={self.form.slug}, doc={self.document.pk}>"
+
+
+def print_document_structure(document):  # pragma: no cover
+    """Print a document's structure.
+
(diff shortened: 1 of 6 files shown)


caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+            self._memoise = {}
+
+        key = str([args, kwargs, meth])
+        if key in self._memoise:
+            return self._memoise[key]
+        ret = meth(self, *args, **kwargs)
+        self._memoise[key] = ret
+        return ret
+
+    return new_meth
```
czosel: I always prefer longer variable names if the abbreviation is not completely obvious. This is an extreme example where the abbreviation even has a (very different) meaning :wink:

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+            ans_question = (
+                self.parent().get_ans_question(question_slug) if self.parent() else None
+            )
+        if ans_question:
+            return ans_question
+
+        # OK start looking in subforms / row forms below our level.
+        # Since we're looking down, we're disallowing recursing to outer context
+        # to avoid recursing back to where we are
+        for subform in self._sub_forms.values():
```
sbor23: See above, this would be:
```python
for subform in self.sub_forms:
    ...
```

caluma/caluma_form/structure.py
czosel: Also, doesn't this break the method depends on `self`?

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+        self._parent = weakref.ref(parent) if parent else None
+
+    def parent(self):
+        return self._parent() if self._parent else None
+
+    def children(self):  # pragma: no cover
+        return []
+
+
+class AnsQuestion(Element):
```
czosel: In `ember-caluma` we're using the same idea, but we're calling it `field` - might be a better name than `AnsQuestion`?

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+    def visit(vis):
+        raise Exception(f"generic visit(): {vis}")
+
+    @visit.register(Element)
+    def _(vis):
+        print("   " * ind["i"], vis)
+        ind["i"] += 1
+        for c in vis.children():
+            visit(c)
+        ind["i"] -= 1
```
sbor23: Nice application of visitor here. :)

caluma/caluma_form/validators.py
```diff
@@ -196,19 +196,16 @@ def validate(self, document, info, validation_context=None, **kwargs):
...
         # we need to build the context in two steps (for now), as
         # `self.visible_questions()` already needs a context to evaluate
         # `is_hidden` expressions
         intermediate_context = {
             "form": document.form,
             "document": document,
-            "all_questions": all_questions,
-            "questions": questions,
-            "answers": self.get_document_answers(document),
+            # "answers": self.get_document_answers(document),
```
sbor23: Leftovers

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+        return ret
+
+    return new_meth
+
+
+class Element:
+    def __init__(self, parent=None):
+        self._parent = weakref.ref(parent) if parent else None
+
+    def parent(self):
```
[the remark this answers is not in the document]
winged: I had this as a property before, but that made more problems than it solved. Better be explicit in such cases :)

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,218 @@
...
+    def sub_forms(self):
+        if self._sub_forms is None:
+            self._sub_forms = {
+                aq.question.slug: aq
+                for aq in self.children()
+                if aq.question.type == Question.TYPE_FORM
+            }
+        return self._sub_forms
+
+    def get_ans_question(
```
czosel:
```suggestion
    def get_field(
```

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,218 @@
...
+        return f"<FieldSet form={self.form.slug}, doc={self.document.pk}>"
+
+
+def print_document_structure(document):  # pragma: no cover
+    """Print a document's structure.
+
+    Intended halfway as an example on how to use the structure
+    classes, and a debugging utility.
+    """
+    ind = {"i": 0}
```
czosel: Possibily stupid question, but I don't see it: Why is this an object instead of just an integer?

caluma/caluma_form/tests/test_validators.py
```diff
@@ -126,60 +126,6 @@ def test_validate_dynamic_option_exists(
...
-    )
-
-    if should_throw:
-        error_msg = f"Questions {sub_form_question_2.question.slug} are required but not provided"
-        with pytest.raises(ValidationError, match=error_msg):
-            DocumentValidator().validate(main_document, info)
-    else:
-        DocumentValidator().validate(main_document, info)
-
-
```
czosel: Why delete the test?

caluma/caluma_form/validators.py
```diff
@@ -228,7 +224,8 @@ def visible_questions(self, document, validation_context=None):
         visible_questions = []
 
         q_jexl = jexl.QuestionJexl(validation_context)
-        for question in validation_context["questions"].values():
+        for ans_question in validation_context["structure"].children():
```
czosel:
```suggestion
        for field in validation_context["structure"].children():
```

caluma/caluma_form/structure.py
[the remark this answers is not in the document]
anehx: Maybe name it `get_parent` to really be explicit..

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class Element:
+    def __init__(self, parent=None):
+        self._parent = weakref.ref(parent) if parent else None
+
+    def parent(self):
+        return self._parent() if self._parent else None
+
+    def children(self):  # pragma: no cover
```
anehx: `get_children`

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+                document,
+                form=question.row_form,
+                question=question,
+                answer=answer,
+                parent=parent,
+            )
+
+        return AnsQuestion(document, form, question, answer, parent=parent)
+
+    def ans_value(self):
```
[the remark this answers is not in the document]
anehx: `get_field`

caluma/caluma_form/tests/test_validators.py
winged: The test is completely covered by `test_validate_hidden_subform`, so no need to have two tests doing the same thing

caluma/caluma_form/structure.py
winged: I'm not a fan of naming things `get_XXX`. It's kind of java-think IMO

anehx: Then make it a property - this doesn't even do anything or am I wrong?

caluma/caluma_form/structure.py
winged: Fixed the naming. And no, it is explicitly meant to be used on methods (vs functions), so it's fine to depend on `self`

caluma/caluma_form/structure.py
winged: Updated the naming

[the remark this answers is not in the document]
winged: Yep, updated the naming

caluma/caluma_form/structure.py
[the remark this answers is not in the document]
winged: `get_field` is wrong, as it's explicitly not the field, but the value to be returned. But updated the naming nonetheless

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+                {cell.question.slug: cell.ans_value() for cell in row.children()}
+                for row in self.children()
+            ]
+        elif self.question.type == Question.TYPE_FORM:
+            # forms never have values themselves
+            return None
+            return [
+                {cell.question.slug: cell.ans_value() for cell in row.children()}
+                for row in self.children()
+            ]
```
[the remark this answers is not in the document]
winged: Fix'd

caluma/caluma_form/structure.py
winged: Python scoping rules. Can't update variables in the "outer" scope by assignment. You could mess with the `global` keyword, but that's even worse

caluma/caluma_form/structure.py
winged: This is overloaded in the subclasses where the implementation actually needs to do things

anehx: Ok

caluma/caluma_form/structure.py
czosel: Ah right, thanks!

caluma/caluma_form/structure.py
czosel: I meant more if that state of the object changes, the memoisation can return incorrect results because the cache key doesn't depend on the state of the object.

winged: Ah. In that case you'd be right, but in this situation, the objects will be virtually immutable, so it should be fine

czosel: :+1:

anehx: Walter-Whites-Smooth-Wink-On-Breaking-Bad

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+            self._ans_questions = {aq.question.slug: aq for aq in self.children()}
+        if self._sub_forms is None:
+            self._sub_forms = {
+                aq.question.slug: aq
+                for aq in self.children()
+                if aq.question.type == Question.TYPE_FORM
+            }
+
+    def get_ans_question(
+        self, question_slug: str, allow_check_parent: bool = True
```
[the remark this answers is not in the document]
winged: Done

caluma/caluma_form/validators.py
winged: Done

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def _build_ans_questions(self):
+        if self._ans_questions is None:
+            self._ans_questions = {aq.question.slug: aq for aq in self.children()}
+        if self._sub_forms is None:
+            self._sub_forms = {
+                aq.question.slug: aq
+                for aq in self.children()
+                if aq.question.type == Question.TYPE_FORM
+            }
```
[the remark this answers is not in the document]
winged: Doen

caluma/caluma_form/structure.py
```diff
@@ -0,0 +1,212 @@
...
+                row_doc,
+                self.question.row_form,
+                question=self.question,
+                parent=self.parent(),
+            )
+            for row_doc in self.answer.documents.all()
+        ]
+
+
+class DocForm(Element):
```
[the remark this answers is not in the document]
winged: Done

caluma/caluma_form/structure.py
winged: done

PR #898, opened 2020-01-30
feat(form): implement CopyDocumentMutation

sbor23:
The `CopyDocument` mutation creates a copy of an existing `Document` and
all it's `Answer`s. A reference to the original instance is stored in the new `source`
field.


caluma/caluma_form/models.py
```diff
@@ -208,8 +208,44 @@ class Document(UUIDModel):
...
+
+            # TableAnswer: copy AnswerDocument too
+            for answer_doc in AnswerDocument.objects.filter(answer=source_answer):
+                new_doc = answer_doc.document.copy(family=family)
+
+                AnswerDocument.objects.create(
+                    answer=new_answer, document=new_doc, sort=answer_doc.sort
+                )
+
+        return new_document
```
open-dynaMIX: What about file answers, their containing file objects and the files in Minio?

caluma/caluma_form/tests/test_document.py
```diff
@@ -1060,3 +1060,89 @@ def test_remove_document_without_case_table(
...
+    )
+    other_q_1 = form_question_factory(
+        form=main_form, question__type=Question.TYPE_TEXT, question__slug="other_q_1"
+    )
+
+    # MD
+    #   - TA
+    #       - RD1
+    #           sqa:"foo"
+    #   - oqa:"something
```
open-dynaMIX: Please make this less cryptic.

caluma/caluma_form/models.py
```diff
@@ -208,8 +208,44 @@ class Document(UUIDModel):
     form = models.ForeignKey(
         "caluma_form.Form", on_delete=models.DO_NOTHING, related_name="documents"
     )
+    source = models.ForeignKey(
+        "self",
+        blank=True,
+        null=True,
+        help_text="Reference this document has been copied from",
+        related_name="+",
```
open-dynaMIX: Up for discussion: Should there be a way to query copies of a document? `related_name="copies"`?

caluma/caluma_form/models.py
sbor23: They are intentionally ignored for the time being...

caluma/caluma_form/models.py
sbor23: Very good question @winged @czosel

czosel: I could imagine that we’re gonna need this somewhen, yeah. If there are no downsides I’d say let’s add it already now.

open-dynaMIX: I don't see any downsides. Of course, there is a bit more needed than setting the `related_name` in order to make it accessible with graphql.

czosel: Right - I’d wait with extending the API though until we have an actual use-case.

caluma/caluma_form/models.py
sbor23: So we should probably not copy them at all to prevent accidentally changing/deleting file answers from the original?

sbor23: Excluded FileAnswers for now

caluma/caluma_form/tests/test_document.py
sbor23: Done

caluma/caluma_form/models.py
sbor23: Added `related_name="copies"`

caluma/caluma_form/models.py
winged: Isn't this some really unexpected/surprising behaviour? The whole document is copied, except for files?

sbor23: Well I added this as a TODO, but your point stands of course

czosel: Can we add this to the graphiQL docs? Something like "Warning: Copying of file answers is currently not supported"

open-dynaMIX: Sorry, I don't really understand this discussion. Why don't we just also copy the files? Doesn't seem that difficult...

czosel: This decision was made purely because we don't need it for our current use case. If it's really that simple, it would be best to implement it of course!

open-dynaMIX: IMO it's ok to not implement something we don't need. Filters come to mind.

But I don't think we should partially implement features, just because we don't need all of it right now.

czosel: I guess copying the caluma FileAnswer is indeed not very hard. What we've been wondering earlier was if we also need to copy the file in minio - which sounds like more effort.

sbor23: Copying the FileAnswer is trivial. However it's very dangerous if we just copy the Answer but not the File. So we end up with two Answers referencing the same file...

open-dynaMIX: Yes, the file in minio has to be duplicated too. Removal of a fileanswer also removes the corresponding `File` object. Removing a `File` object also removes the file in minio (in fact it moves the file in minio, so it stays available for historical records.

So if multiple `File` objects reference the same minio file and one of those `File` objects gets deleted, the other `File` objects reference a file that doesn't exist anymore.

czosel: Right - do you think copying the file inside minio is difficult?

sbor23: Alright we're doing the file copy!

sbor23: Done

sbor23:
@czosel @winged care to resolve and merge?

caluma/caluma_form/models.py
czosel: Looks simple enough 😅

open-dynaMIX:
Please wait with merging. I've got still some questions.

caluma/caluma_form/models.py
```diff
@@ -245,12 +285,20 @@ class Meta:
...
+
     def _move_blob(self):
         # move the file on update
         # this makes sure it stays available when querying the history
-        old_file = self.history.first()
-        new_name = f"{old_file.pk}_{old_file.name}"
-        client.move_object(self.object_name, new_name)
+        client.move_object(*self._get_file_names())
+
+    def copy(self, *args, **kwargs):
```
open-dynaMIX: This does copy the file in minio, not the `File` object. So for consistency, it should be called `copy_blob()`.

caluma/caluma_form/models.py
```diff
@@ -245,12 +285,20 @@ class Meta:
...
+        old_name, new_name = self._get_file_names()
```
open-dynaMIX: Correct me if I'm wrong: This constructs the new file name from the pk of the newest historical entry. Why?

sbor23: Addressing the top question as well:
You're right, something is off.
I think we need to create and copy of `File`, because we need to know it's pk to copy the blob to the right name. And return the copy ofc, so the name `copy()` would still make sense.

open-dynaMIX: The logic should be: Create a new `File` object with the same name as the original one. And then call

```python
client.copy_object(original.object_name, new_file.object_name)
```

sbor23: Yes that's what I just did. :)
Though the name of the function would stay `copy()`.

open-dynaMIX: Now the name `copy()` makes sense of course.

caluma/caluma_form/models.py
```diff
@@ -208,8 +208,48 @@ class Document(UUIDModel):
...
+        help_text="Reference this document has been copied from",
+        related_name="copies",
+        on_delete=models.SET_NULL,
+    )
     meta = JSONField(default=dict)
 
+    def copy(self, family=None):
+        """Create a copy including all its answers."""
+
+        new_document = self.__class__.objects.create(
```
winged: Nitpick: A cleaner way would be to use `type(self)` instead of `self.__class__`.

But feel free to change or merge nonetheless

caluma/caluma_form/tests/test_document.py
```diff
@@ -1060,3 +1060,114 @@ def test_remove_document_without_case_table(
...
+    assert file_question.question.slug in new_document.answers.all().values_list(
+        "question__slug", flat=True
+    )
+    new_file_answer = new_document.answers.get(
+        question__slug=file_question.question.slug
+    )
+    assert new_file_answer.value == file_answer.value
+    # file is copied
+    minio_mock.copy_object.assert_called()
+    assert new_file_answer.file.name != file_answer.file.name
```
open-dynaMIX:
```suggestion
    assert new_file_answer.file.object_name != file_answer.file.object_name
```

sbor23: Done

caluma/caluma_form/models.py
sbor23: Done

winged:
Sorry @sbor23 I think #935 introduced some changes you need to rebase onto

sbor23:
> Sorry @sbor23 I think #935 introduced some changes you need to rebase onto

Done

PR #913, opened 2020-02-07
feat(document): add source field referencing parent

sbor23:


czosel:
Hmm, feels a bit strange when copying is not even possible within caluma, doesn't it?
cc @open-dynaMIX @anehx

sbor23:
Migration will clash with https://github.com/projectcaluma/caluma/pull/913

open-dynaMIX:
I agree with @czosel . I think we have two options:

1. Implement copying of documents in caluma and add this field
2. When copying is done outside of caluma, the `meta` field should be used for storing this kind of information.

> Migration will clash with #913

Did you mean to link to something else?

sbor23:
Closing in favour of original PR https://github.com/projectcaluma/caluma/pull/898

PR #919, opened 2020-02-12
fix(auth): fix user object creation

### LINKED ISSUE #912: Fix client token auth
The logic now assumes that for client tokens, a 401 or 403 is returned by the `userinfo` endpoint. If that happens we query the `introspection` endpoint. This seems to have changed in Keycloak and the `userinfo` endpoint does return some data and hence the `introspect` endpoint is never called.

The distinction between `OIDCUser` and `OICDClient` is not always possible, so we should avoid trying to make it. Instead we should always create `OIDCUser` instances and handle claims there if they come from `userinfo` or `introspect`.

The distinction can then still be made by the users based on the claims they can access on the `OIDCUser`.

open-dynaMIX:
Some OIDC implementations allow to call the `userinfo` endpoint with a
client token. This breaks the distinction between `OIDCUser` and
`OIDCClient` and always an `OIDCUser` is created.

This commit implements a generic solution for that problem:

 - There is no `OIDCClient` class anymore
 - `OIDCUser` is instantiated with either a `userinfo`- or a
   `introspection`-dict.
 - The provided claims are stored in the new `OIDCUser.claims` property.
 -  The new `OIDCUser.claims_source` property indicates the source of the claims
    (`userinfo` or `introspection`).
 -  A `DeprecationWarning` will be given, when accessing the now
    deprecated properties `userinfo` or `introspection`.

Closes #912


open-dynaMIX:
I'm not sure if this should be considered a breaking change. Implementations that rely on the type of the user object will break. But I'm not aware of any.

Accessing the `userinfo` or `inspection` property on the user object is deprecated, but still works.

PR #923, opened 2020-02-14
fix(data_source): add unique_together to DynamicOption

### LINKED ISSUE #922: Add unique_together to DynamicOption
`DynamicOptions` are not unique right now. In order to have only one record per question, document and slug, we should add a `unique_together` contraint. This also allows us to make reliable `get` lookups instead of `filter().first()`.

open-dynaMIX:
DynamicOptions are not unique right now. In order to have only one record per
question, document and slug, this commit adds a `unique_together` contraint.
This also allows us to make reliable get lookups instead of filter().first().

Closes #922


anehx:
Hm why aren't they? There should only be one dynamic option per document per question - or am I wrong?

open-dynaMIX:
> There should only be one dynamic option per document per question - or am I wrong?

No, there are situations where e.g. only the `label` changes for an entry in the data source. In that case a new record will be created. In order to get the latest label the next time this entry is selected, my proposed change is needed.

While explaining this, I've just realised, that even more records are created, because the code for creating them looks like that:

```python
        DynamicOption.objects.get_or_create(
            document=document,
            question=question,
            slug=option,
            label=valid_label,
            created_by_user=info.context.user.username,
            created_by_group=info.context.user.group,
        )
```

No idea why they also get created per user and group. I'm gonna change this as well.

open-dynaMIX:
As discussed, we now use a `unique_together` contraint on the `DynamicOption` model.

anehx:
What happens with data that fail the new constraint?

open-dynaMIX:
> What happens with data that fail the new constraint?

Great question :). I'm on it.

open-dynaMIX:
Ready for review.

PR #932, opened 2020-02-17
feat(validation): add support for minlength validation

czosel:


czosel:
ready for review

czosel:
When integrating this with ember-caluma, a question came up: If implemented like this, it's not possible to save answers when their length is shorter than the required length. Should we consider skipping the minlength validation on `SaveAnswer` just like we do with `required`? But we'd have to show the error in the frontend for the user to notice that something is wrong ... 
@anehx

caluma/caluma_form/serializers.py
```diff
@@ -197,6 +197,7 @@ class Meta:
 
 
 class SaveTextQuestionSerializer(SaveQuestionSerializer):
+    min_length = IntegerField(min_value=1, required=False, allow_null=True)
```
winged: Shouldn't `min_value` be 0? Otherwise, you'd need to explicitly set it to `None`/`null` when you don't want to enforce minimum length, leading to additional check required in the frontend

czosel: Sure? Shouldn't `required=False, allow_null=True` allow for that case? See also how `max_length` is defined on the line below.

winged: Yes, it would allow setting it to `null` but not to 0, which in my opinion should be allowed (and cause the same effect)

czosel: Ah, now I understand what you mean. The additional check in the frontend is already implemented, you can try it with `maxLength` here: https://demo.caluma.io/form-builder/rst/questions/test

winged: Hmm, `maxLength` has a bit other semantics, as a max length of 0 doesn't make sense. Setting `minLength` to 0 however could make sense from a user's perspective, even if it's the same as the value being absent.

However I don't have strong enough feelings about this to be pedantic about it, so feel free to merge :)

PR #942, opened 2020-02-19
feat(models): increase max-length of slugs to 150

fkm:


winged:
Looks good, but you need to run `black` on the changed files

fkm:
Sorry!

Flake8 and Black are done and pytest is running. I'll push as soon as the tests are done.

Issue #940, opened 2020-02-19
Add caluma to PyPI


open-dynaMIX:
We've decided to add caluma to the python package index (PyPI).

I've already uploaded a package to https://test.pypi.org. It can be found [here](https://test.pypi.org/project/caluma/).

Here's how to install it with pip:

```shell
pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple caluma
```

TODOs:

- [x] Verify everything works as expected.
- [ ] Verify all necessary files are added to the source dist (`MANIFEST.in`).
- [ ] Use absolute URIs in the documentation, so that the links are also working in PyPI. Or use a dedicated markdown file for PyPI.
- [x] Create a CI job for automatically building and uploading caluma when a new tag is created.
- [x] Add documentation about usage of caluma apps outside of the caluma service

Related issues and PRs:

* #938 
* #939 
* #941
* #946 
* #947 
* #950
* #951 
* #953 
* #954 
* #959

winged:
Closing, this has been completed for a while now :)

PR #947, opened 2020-02-20
docs(interface): add docs about usage of django apps and interfaces

### LINKED ISSUE #941: Document API available from extension points
This discussion came up during PR #935. I agree we should address this in the documentation to let people know exactly what interfaces should be stable from within extension points, and which ones are likely to change.

---

> Discussion: The family was kind of intended to be non-public from Caluma's perspective, but is still used in customer project's permissions / visibility layers.

Related idea to this, I think it would help to define the API that is officially supported to be used in extensions. This should be documented, so that it's obvious for all users. 

Usage of modules outside of the defined API should still be possible, but they might be subject to change.

Right now it's not really possible to write extensions without reading the caluma source. Documentation of those modules would help new users to get started.

> Due to this change, this PR would possibly need to be escalated to "breaking" - what do you guys think?

I agree.

_Originally posted by @open-dynaMIX in https://github.com/projectcaluma/caluma/pull/935#issuecomment-588061140_

open-dynaMIX:
This commit adds some documentation about

* using the caluma django apps inside an existing django project
* supported interfaces.

Related to #940
Closes #941

### DIFF
--- a/docs/django-apps.md
+++ b/docs/django-apps.md
@@ -1,0 +1,87 @@
+# Using Caluma as django apps
+
+## Disclaimer
+
+It is highly recommended to use Caluma as a dedicated service. However, there are usecases, where
+it might make sense to integrate Caluma into another django project.
+
+If you just want to get Caluma up and running, please see the documentation about [setting up
+the Caluma service](configuration.md).
+
+Please beware that Caluma only works with PostgreSQL, and requires the `psqlextra`
+backend to use the advanced features (such as JSON fields etc).
+
+## Installation
+
+Caluma is on PyPI, so you can just
+
+```shell
+pip install caluma
+```
+
+
+## Configuration
+
+### Installed apps
+
+Add the Caluma apps you want to use to your `INSTALLED_APPS`.
+
+Some notes about Caluma-internal dependencies:
+
+* `caluma_workflow` needs `caluma_form` to work correctly (as cases and work items point to documents)
+* `caluma_core` should always be installed when using Caluma (though, technically,
+  it would work without it, as core doesn't contain any models
+* `caluma_user` should always be added (same caveat as with core)
+
+```python
+INSTALLED_APPS = [
+    # ...
+    # Caluma and it's dependencies:
+    "caluma.caluma_core.apps.DefaultConfig",
+    "caluma.caluma_user.apps.DefaultConfig",
+    "caluma.caluma_form.apps.DefaultConfig",
+    "caluma.caluma_workflow.apps.DefaultConfig",
+    "caluma.caluma_data_source.apps.DefaultConfig",
+    "graphene_django",
+    "localized_fields",
+    "psqlextra",  # Caluma needs a PostgreSQL database using the psqlextra backend
+    "simple_history",
+    # ...
+]
+```
+
+### Settings
+
+Import the Caluma settings at the top of your `DJANGO_SETTINGS_MODULE`:
+
+```python
+from caluma.settings.caluma import *  # noqa
+```
+
+This will only load Caluma specific settings and no django specific ones (db, cache, etc.).
+
+Then you can configure caluma normally [using environment variables](configuration.md).
+
+
+### Database
+
+Caluma needs a PostgreSQL database using the `psqlextra` backend to use the advanced
+features (such as JSON fields etc).
+
+This is mandatory, that's why we tell you thrice.
+
+### Debug middleware
+
+It's recommended to add following lines to your settings file (after importing the caluma settings):
+
+```python
+# GraphQL
+if DEBUG:
+    GRAPHENE["MIDDLEWARE"].append("graphene_django.debug.DjangoDebugMiddleware")
+```
+
+
+## Supported interfaces
+
+See [interfaces](interfaces.md) for information about the officially supported API.
+
--- a/docs/interfaces.md
+++ b/docs/interfaces.md
@@ -1,0 +1,303 @@
+# Interfaces
+
+Here you find a list of interfaces we support and won't change without further notice.
+
+
+## Models
+
+We can't guarantee stability for whole models. However certain fields are considered stable.
+Changes in non-guaranteed fields won't receive prominent mention in release notes, for example,
+and won't trigger a major version bump (according to semver).
+
+For more information about a model and it's fields, please consult the source code.
+
+### General
+
+All models share a common set of fields:
+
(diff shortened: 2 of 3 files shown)


docs/django-apps.md
```diff
@@ -0,0 +1,67 @@
...
+pip install caluma
+```
+
+
+## Configuration
+
+### Installed apps
+
+Add the caluma apps you want to use to your `INSTALLED_APPS`:
+
```
winged: Some note on the caluma-internal dependencies would be useful:

* caluma_workflow needs caluma_form to work correctly (as cases and work items point to documents)
* core should always be installed when using caluma (though, *technically*, it would work without it, as core doesn't contain any models
* caluma_user should always be added (same caveat as with core)

I would also highlight using `psqlextra`, as Caluma won't work without it (eg. with the standard postgresql backend)

docs/django-apps.md
```diff
@@ -0,0 +1,67 @@
+# Using Caluma as django apps
+
+## Disclaimer
+
+It is highly recommended to use Caluma as a dedicated service. However, there are usecases, where
+it might make sense to integrate Caluma into another django project.
+
+If you just want to get Caluma up and running, please see the documentation about [setting up
+the Caluma service](configuration.md).
```
winged: Add a disclaimer that Caluma only works with PostgreSQL. Direct integration won't work on other databases.

docs/django-apps.md
```diff
@@ -0,0 +1,67 @@
...
+    "psqlextra",
+    "simple_history",
+    # ...
+]
+```
+
+### Settings
+
+Import the caluma settings at the top of your `DJANGO_SETTINGS_MODULE`:
+
```
winged: Add a note that this only contains the settings that are required in addition to the regular django settings.

(Possibly also mention again to configure the database to use the `psqlextra` backend)

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
+# Interfaces
+
+Here you find a list of interfaces we support and won't change without further notice.
+
+
+## Models
+
+We can't guarantee stability for whole models. However certain fields are considered stable.
```
winged:
```suggestion
We can't guarantee stability for whole models. However certain fields are considered stable.
Changes in non-guaranteed fields won't receive prominent mention in release notes, for example,
and won't trigger a major version bump (according to semver)
```

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
...
+* `modified_at`
+* `created_by_user`
+* `created_by_group`
+* `history` --> manager for accessing historical records
+
+
+### caluma_form.models.Form
+
+#### Fields
+
```
winged: I'd add a (very) short description on what this model represents

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
...
+* `name`
+* `description`
+* `meta`
+* `is_published`
+* `is_archived`
+* `questions`
+* `source`
+
+### caluma_form.models.FormQuestion
+
```
winged: I'd add a (very) short description on what this model represents

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
...
+#### Properties
+
+* `min_length`
+* `max_length`
+* `max_value`
+* `min_value`
+
+#### Methods
+
+* `empty_values()`
```
winged: I'd remove this - it's only used by the validation internally and I don't assume it has a good reason to be used by extensions

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
...
+#### Fields
+
+* `id` / `pk`
+* `question`
+* `option`
+* `sort`
+
+#### Methods
+
+* `natural_key()`
```
winged: I think this is only used to tell django how to generate the primary key. Shouldn't be part of the public API

docs/interfaces.md
```diff
@@ -0,0 +1,315 @@
...
+
+#### Properties
+
+* `object_name`
+* `upload_url`
+* `download_url`
+* `metadata`
+
+
+### caluma_form.models.DynamicOption
```
winged: I wonder if we want to leave out DynamicOption alltogether from the public API. This is only used to keep track of once-valid options to dynamic questions and shouldn't be messed with by extension writers, IMHO

open-dynaMIX:
@winged Thanks for your suggestions. The requested changes are done.

docs/django-apps.md
winged:
```suggestion
the Caluma service](configuration.md).

Please beware that Caluma only works with PostgreSQL, and requires the `psqlextra` backend to use the advanced features (such as JSON fields etc)
```

docs/django-apps.md
```diff
@@ -0,0 +1,87 @@
...
+### Installed apps
+
+Add the Caluma apps you want to use to your `INSTALLED_APPS`.
+
+Some notes about Caluma-internal dependencies:
+
+* `caluma_workflow` needs `caluma_form` to work correctly (as cases and work items point to documents)
+* `caluma_core` should always be installed when using Caluma (though, technically,
+  it would work without it, as core doesn't contain any models
```
czosel: I'd leave the note in brackets out because it doesn't really matter to someone who just wants to get started with caluma in their django app (same below)

docs/interfaces.md
```diff
@@ -0,0 +1,303 @@
...
+* `caluma_core.visibilities.BaseVisibility`
+* `caluma_core.visibilities.filter_queryset_for`
+* `caluma_core.visibilities.Union`
+* `caluma_data_source.data_source.BaseDataSource`
+* `caluma_data_source.utils.data_source_cache`
+* `caluma_user.permissions.CreatedByGroup`
+* `caluma_user.permissions.IsAuthenticated`
+* `caluma_user.visibilities.Authenticated`
+* `caluma_user.visibilities.CreatedByGroup`
+* `caluma_workflow.visibilities.AddressedGroups`
```
czosel: Do we have a strategy to keep this file up to date?

czosel:
Awesome work! :100:

docs/django-apps.md
open-dynaMIX: See #954.

docs/interfaces.md
open-dynaMIX: No, not yet. Do you have a suggestion?

PR #958, opened 2020-02-24
fix(models): reduce slug max-length to 120

fkm:
This doesn't work. I still get the error message that the slug I'm trying to save is too long for a varying 50 field.


czosel:
What didn’t work before?

fkm:
> What didn’t work before?

The slugs got too long after I started adding a timestamp suffix. As we have fields in the database that combine two slugs, we need to keep them below `255 / 2`.

But there's some field somewhere that doesn't get updated by the migrations.

czosel:
This sounds a little bit too application-specific to me to warrant another change in caluma, but maybe I'm just missing some context.

fkm:
@kaldras @anehx
Could you explain this?

I'm this close to go with my initial idea and hash all the slugs.

caluma/caluma_form/migrations/0033_auto_20200221_1346.py
```diff
@@ -0,0 +1,69 @@
...
+            field=models.ForeignKey('Question', on_delete=models.CASCADE),
+        ),
+        migrations.AlterField(
+            model_name='questionoption',
+            name='option',
+            field=models.ForeignKey('Option', on_delete=models.CASCADE),
+        ),
+
+        # The previous migrations fail to update the length of question_id.
+        migrations.RunSQL("ALTER TABLE caluma_form_formquestion ALTER COLUMN question_id TYPE character varying(120);"),
```
anehx: Shouldn't this be done for `caluma_form_questionoption` too?

fkm: This was a hack that shouldn't be done anywhere.

The reason I did this was because the this field didn't update its length with the 'normal' migration.

czosel: Maybe best to mark this PR as a "draft"? Then everyone knows that reviews are not really wanted yet.

fkm: How about it is assigned to a specific developer?

czosel: To me that carries a different meaning in this repo. Since caluma is used in many projects, reviews are often done by multiple people, even when not assigned.

PR #959, opened 2020-02-24
chore(urls): move view and add urls.py to core

open-dynaMIX:
This commit moves the graphql view to the core app and renames it to
`CalumaGraphQLView`. Additionally it adds a `caluma_core.urls` that gets
included in the main `urls.py`.

This means third-party projects can include the caluma urls like this:

```python
from django.conf.urls import include, url

urlpatterns = [
    # ...
    url(r"", include("caluma.caluma_core.urls")),
    # ...
]
```

### DIFF
--- a/caluma/caluma_core/urls.py
+++ b/caluma/caluma_core/urls.py
@@ -1,0 +1,8 @@
+from django.conf import settings
+from django.conf.urls import url
+
+from caluma.caluma_user.views import AuthenticationGraphQLView
+
+urlpatterns = [
+    url(r"", AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG), name="graphql")
+]
--- a/caluma/urls.py
+++ b/caluma/urls.py
@@ -1,12 +1,3 @@
-from django.conf import settings
-from django.conf.urls import url
+from django.conf.urls import include, url
 
-from .caluma_user.views import AuthenticationGraphQLView
-
-urlpatterns = [
-    url(
-        r"^graphql",
-        AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG),
-        name="graphql",
-    )
-]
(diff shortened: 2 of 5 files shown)


caluma/caluma_core/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.conf import settings
+from django.conf.urls import url
+
+from caluma.caluma_user.views import AuthenticationGraphQLView
+
+urlpatterns = [
+    url(
+        r"^graphql",
```
winged: I would remove the prefix here so we can add it in the "main" URLs file

caluma/urls.py
```diff
@@ -1,12 +1,3 @@
...
-from .caluma_user.views import AuthenticationGraphQLView
-
-urlpatterns = [
-    url(
-        r"^graphql",
-        AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG),
-        name="graphql",
-    )
-]
+urlpatterns = [url(r"", include("caluma.caluma_core.urls"))]
```
winged: Add the "/graphql" prefix here

docs/django-apps.md
```diff
@@ -80,6 +80,21 @@ if DEBUG:
...
+### urls.py
+
+Include the Caluma urls:
+
+```python
+from django.conf.urls import include, url
+
+urlpatterns = [
+    # ...
+    url(r"", include("caluma.caluma_core.urls")),
```
winged: Add a suggestion for  `/graphql` as URL prefix for the include here

caluma/caluma_core/urls.py
open-dynaMIX: Done.

caluma/urls.py
open-dynaMIX: Done.

docs/django-apps.md
open-dynaMIX: Done.

Issue #773, opened 2019-11-01
Update python version


sbor23:
Now that 3.8 is out: Are there any particular reasons for staying on the `python:3.6` base image?
There are some goodies for developers especially like my precious `breakpoint()` which catches me off guard every time I try to use it in a <3.7 project.

So my proposal would be to bump the image to `python:3.7`, it will have bug fixes until 2023.
https://www.python.org/dev/peps/pep-0537/#and-beyond-schedule

czosel:
We'd have to check if python 3.7 is available for our projects with non-docker deployments.

open-dynaMIX:
One of the reasons we didn't switch to 3.7 already is, that not all dependencies had support for it. Before making this change, we have to make sure that this is not the case anymore.

ganwell:
I accidentally tried 3.7: we have a lot of snapshot-tests, many of them break because in 3.7 the output is ordered different. I guess we have to check all that output and update the snapshots. Maybe we should check if jumping to 3.8 causes less problems. In theory dicts have a fixed order since Raymond Hettingers dicts improvements in 3.6, maybe something else is causing the problem.

PR #960, opened 2020-02-25
chore: update to python 3.7

sbor23:
We used caluma already for some time in a downstream project with python 3.7, so pretty sure we can just increase the version string and should be done with it.

Closes https://github.com/projectcaluma/caluma/issues/773


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6@sha256:764b2c6beae96a8c8c6046392d5b1b691773337c5a7fd29347935ee0a3223146
+FROM python:3.7.6
```
ganwell: The implications of this have been discussed? I am stickler for hash chains.

sbor23: Not really. My thought process was that maybe renovate bot picks it up this way when the next point release is available, not sure tho.
Would something advice against going `python:3.7.6@sha256:xxx` or is this redundant?

ganwell: Tags can change, digests can not. If you only use the tag the image you are using can change if upstream updates the tag. In 3.6@hash, the tag should become decoration. At least that is what the documentation says.

I guess I could test that with a registry, but I trusted docker in that matter.

ganwell: I tested that.

sbor23: Alrighty did that :)

PR #962, opened 2020-02-26
fix(models): slug fields should be 127 chars

winged:
Slug field length is limited to 120 chars, so we can use combined
slugs in many-to-many rel tables as a natural primary key (for example
form-slug.question-slug in the FormQuestion model). The NaturalKeyModel
which we built for this has a CharField primary key with max length of
255 chars.  We use 120 instead of the technical possible 122 chars so
it can better be remembered.


winged:
@open-dynaMIX This contains some weird mess that I've not been able to track down yet: Sometimes, the foreign keys don't get updated to the length of the SlugField they're pointing to. I've created a test to verify this correctness, which works properly, but I could not figure out why this happens in the first place.

I'd be glad if we could track this down in the coming days, as it's really a weird issue

caluma/caluma_form/tests/test_migration.py
```diff
@@ -273,3 +275,95 @@ def test_migrate_to_family_as_pk(transactional_db, caplog):
     assert new_unrelated_doc.family == new_unrelated_doc
     assert new_root_doc.family == new_root_doc
     assert expected_msg in caplog.messages
+
+
+def _col_type_from_db(field):
```
open-dynaMIX: This could be imported from `caluma.utils`. Only difference is `_field_name()` is inside `_col_type_from_db` here, but outside in `utils`.

czosel:
This might sound a bit weird, but I think `127` chars would be possible - which is actually quite easy to remember since it's `2^7 - 1` :wink:

winged:
Ready for re-review. Fixed all your feedback  - even the 127chars @czosel :)

caluma/caluma_form/tests/test_migration.py
winged: Hmm this is a leftover from me refactoring it.. fixed!

PR #971, opened 2020-03-05
feat(workflow): add case families

### LINKED ISSUE #235: Keyword: Case families
(just like document families)

This will be used to simplify visibility to a tree of cases.

open-dynaMIX:
Analogous to `Document` families, this PR implements `Case`
families.

These commits

 - add the field `family` to the case model
 - add a migration that sets the family for existing cases (historical cases referencing non-existent cases are treated as their own family)
 - implement the automatic setting of the case family when starting new cases
 - add a `rootCase` filter, in order to fetch cases belonging to the same family
 - add a `family_work_items` field to the `Case` type

Closes #235


open-dynaMIX:
It's quite a big PR. I suggest reviewing the commits one-by-one.

caluma/caluma_workflow/migrations/0020_case_families.py
```diff
@@ -0,0 +1,60 @@
...
+    HistoricalCase = apps.get_model("caluma_workflow", "HistoricalCase")
+    db_alias = schema_editor.connection.alias
+
+    orphan_cases = Case.objects.using(db_alias).filter(family__isnull=True)
+    orphan_historical_cases = HistoricalCase.objects.using(db_alias).filter(
+        family__isnull=True
+    )
+
+    for case in orphan_cases:
+        case.family = case
```
winged: This doesn't set the case's family correctly IMHO, as it doesn't take the family structure into account. I think we should actually do that, so even "old" cases will have the correct structure.

caluma/caluma_workflow/models.py
```diff
@@ -111,6 +113,14 @@ class Case(UUIDModel):
         (STATUS_CANCELED, "Case is cancelled."),
     )
 
+    family = models.ForeignKey(
+        "self",
+        help_text="Family id which case belongs too.",
```
winged: Typo (s/too/to/)

open-dynaMIX: Fixed.

caluma/caluma_workflow/migrations/0020_case_families.py
open-dynaMIX: You're absolutely right. It's fixed. Historical cases referencing non-existent cases, are treated as their own family.

caluma/caluma_workflow/migrations/0020_case_families.py
```diff
@@ -0,0 +1,71 @@
...
+    db_alias = schema_editor.connection.alias
+
+    orphan_cases = Case.objects.using(db_alias).filter(family__isnull=True)
+    orphan_historical_cases = HistoricalCase.objects.using(db_alias).filter(
+        family__isnull=True
+    )
+
+    for case in orphan_cases:
+        family = case
+        while hasattr(family, "parent_work_item"):
```
sbor23: Out of interest, why `while` and not `if`?

winged: @sbor23 Because you want the family to point to the top-level case, not just the parent, so you need to go as far up the tree as you can

sbor23: Ah gotcha!

PR #972, opened 2020-03-05
fix(validation): the word "form" should always be the root form

winged:
When evaluating JEXL expressions, you can use the word `form` to get
the name of the top form. This wasn't the case ror row forms until now
(but only in is_hidden evaluation).  Now, `form` always points to the
root form of the document.


anehx:
@open-dynaMIX It should not evaluate to a different value in `is_hidden` and `is_required` the `form` context variable should **always** be the root form in our question JEXL! This was definied like that at the beginning of caluma but was wrongly implemented in some of our countless reimplementations of the validation..

open-dynaMIX:
@anehx Thanks for the clarification! I misread the PR description.

PR #976, opened 2020-03-09
feat(events): implement caluma events

### LINKED ISSUE #970: [workflow] Caluma "events" (signals)
## Rationale

We want Caluma to emit signals at certain stages of workflows.

Right now we solve this in various projects with the (mis-)use of the `validations` extension. This has various shortcomings:

 - It handles work that can't be considered "validation"
 - We want to trigger actions after making sure the validation will be successful. Hence duplicating the actual validation. This can lead to dangerous situations: What if our duplicated validation let's something slip through the actual validation prohibits at a later point?
 - It is tied to requests and exclusively takes place in the context of them.
 - It clutters the actual validation extension --> actual validation vs. tasks we want to run at the successful end of a mutation

Signals will provide us with a simple and transparent way to react on certain events.

In order for users of the Caluma service to make use of those signals, we propose an additional extension `events.py`.

## Signal receivers are blocking

For this kind of "side-effects" it is desirable, that receivers are non-blocking. This is not possible with plain signals. In order to provide us with a base that will allow us to extend Caluma events in a way that enables non-blocking receivers of events, we propose to build a simple wrapper (or just an alias) around the `django.dispatch.receiver` decorator called `@on`.

So in a first step, we will implement Caluma "events" in a blocking manner, using the `@on` decorator. At a later point, we will be able to delegate the execution of the receiver function to celery or another task queue. As we don't see a valid usecase for blocking receivers vs. non-blocking, this is considered an implementation detail that does not affect the public API.

## Bigger context

This proposed change has to be viewed in the context of the higher goal to make it as easy as possible for third-party applications to react on events in Caluma. Having this signals infrastructure in place will provide us with a good starting point to further pursue this goal.

Some related discussion can be found in #468.

## Example

Caluma emits an event when a `WorkItem` is completed. We want to send a notification when this happens. This could be implemented as follows in the `events` extension:

``` python
from caluma_core.events import on
from caluma_workflow.events import complete_work_item


@on(complete_work_item)
def send_mail_on_complete_work_item(sender, work_item, **kwargs):
    if work_item.task.slug == "slug-we-are-interested-in":
        # send notification
        pass
```

## TODOs

 - [ ] Implement a sub-set of the needed signals we anticipate
 - [ ] Implement a wrapper around (or an alias for)  `django.dispatch.receiver` called `@on`
 - [ ] Implement an additional extension `events`
 - [ ] Document the events and their usage

open-dynaMIX:
This commit implements Caluma Events. It provides a starting set of
signals, that can be used in event receivers.

In order to provide us with a base that will allow us to extend Caluma events
in a way that enables non-blocking receivers of events, this commit aliases
the `django.dispatch.receiver` decorator as `@on`.

Example event receiver:

```python
from caluma.caluma_core.events import on
from caluma.caluma_workflow.events import completed_work_item
from caluma.caluma_workflow.schema import CompleteWorkItem


@on(completed_work_item)
def send_mail_on_complete_work_item(sender, work_item, **kwargs):
    if work_item.task.slug == "slug-we-are-interested-in":
        # send notification
        pass


# It's also possible to specify a sender, which refers to a mutation class,
# analog to the `permission_for` decorators
@on(completed_work_item, sender=CompleteWorkItem)
def send_mail_on_complete_work_item_2(sender, work_item, **kwargs):
    if work_item.task.slug == "slug-we-are-interested-in":
        # send notification
        pass
```

Closes #970


open-dynaMIX:
Ready for review.

I'd like to discuss the naming/selection of the events. Right now they are (with one exception) derived from the serializer that emits the event. Maybe we come up with a better scheme.

caluma/caluma_workflow/events.py
```diff
@@ -0,0 +1,12 @@
+from django.dispatch import Signal
+
+new_work_item = Signal(providing_args=["work_item", "info"])
```
winged: It might be useful to name the events using a past-tense verb to imply more strongly that the events are triggered after the completion of the change. For example `saved_work_item`, `created_work_item` and so on.

BTW What's the difference between `new_work_item` and `create_work_item`? Seeing from the code, one seems to be when created implicitly via starting a new case, the other when creating one explicitly. Could this be encoded into a parameter instead?

caluma/caluma_workflow/events.py
```diff
@@ -0,0 +1,12 @@
...
+save_work_item = Signal(providing_args=["work_item", "info"])
+create_work_item = Signal(providing_args=["work_item", "info"])
+cancel_work_item = Signal(providing_args=["work_item", "info"])
+complete_work_item = Signal(providing_args=["work_item", "info"])
+skip_work_item = Signal(providing_args=["work_item", "info"])
+complete_case = Signal(providing_args=["case", "info"])
+save_case = Signal(providing_args=["case", "info"])
+start_case = Signal(providing_args=["case", "info"])
```
winged: Another thing: I'd remove the `info` bit from the event. We explicitly want to de-couple the events from the request, so this should be omitted in my opinion.

caluma/caluma_workflow/serializers.py
```diff
@@ -240,6 +242,21 @@ def create(self, validated_data):
...
         bulk_create_with_history(work_items, models.WorkItem)
+
+        send_event(
+            events.start_case,
+            sender=self.__class__,
+            case=instance,
+            info=self.context["info"],
+        )
+        for work_item in work_items:  # pragma: no cover
+            send_event(
```
winged: Those calls take up a lot of space. I think if we could put them into a method of the serializers (mixin?), we could simplify the calls a lot, leading to less clutter in the process (especially regarding code readability). This could imply removing the `self.__class__` parameter as it could be implied, and if we remove the `info` parameter as noted above, an event would in most cases use a single line instead of six as it is now

caluma/caluma_workflow/serializers.py
```diff
@@ -274,16 +311,30 @@ def validate(self, data):
...
-            closed_at=timezone.now(),
-            closed_by_user=user.username,
-            closed_by_group=user.group,
+        ):
+            work_item.status = models.WorkItem.STATUS_CANCELED
+            work_item.closed_at = timezone.now()
+            work_item.closed_by_user = user.username
+            work_item.closed_by_group = user.group
+            work_item.save()
+            send_event(
```
winged: I'm wondering if we should first update all work items, and only after the update send out the events. Otherwise, if there's some inter-relationship between the work items, the event handler may see some items in the old state, and some in the new

docs/events.md
```diff
@@ -0,0 +1,91 @@
...
+@on(complete_work_item, sender=CompleteWorkItemSerializer)
+def send_mail_on_complete_work_item_2(sender, work_item, **kwargs):
+    if work_item.task.slug == "slug-we-are-interested-in":
+        # send notification
+        pass
+```
+
+## @on
+
+The `@on` decorator takes two arguments:
```
winged: Please add a note on why we're doing the `@on` decorator instead of using Django's `@receiver`

caluma/caluma_workflow/events.py
open-dynaMIX: > BTW What's the difference between new_work_item and create_work_item? Seeing from the code, one seems to be when created implicitly via starting a new case, the other when creating one explicitly. Could this be encoded into a parameter instead?

I would go one step further: Let's have just one `created_work_item` signal. The sender makes it clear why the event was emitted.

winged: Indeed, that's even better!

open-dynaMIX: Done.

caluma/caluma_workflow/events.py
open-dynaMIX: Implemented as discussed on the phone.

caluma/caluma_workflow/serializers.py
open-dynaMIX: Done as discussed.

caluma/caluma_workflow/serializers.py
open-dynaMIX: Nice catch! Done.

docs/events.md
open-dynaMIX: Done.

open-dynaMIX:
@winged Ready for re-review.

caluma/caluma_core/events.py
```diff
@@ -0,0 +1,28 @@
...
+    Send events and handle exceptions in receiver functions.
+
+    This wrapper handles the sending of signals as well as handling of any Exceptions that
+    occur. Exceptions are logged with `ERROR` level, but not raised, in order to prevent
+    rollback of the transaction.
+    """
+    result = signal.send_robust(**kwargs)
+    for func, exc in result:
+        if exc is not None:
+            logger.error(
```
winged: You could use `logger.exception()` instead, which produces some more output

caluma/caluma_workflow/tests/test_events.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_events(db, work_item_factory, schema_executor):
+    work_item = work_item_factory(status=models.WorkItem.STATUS_READY, child_case=None)
+
+    @on(completed_work_item)
+    def complete_work_item_event_receiver(sender, work_item, **kwargs):
+        work_item.meta = {"been-there": "done that"}
+        work_item.save()
+        triggering_name_error_should_not_raise  # noqa
```
winged: I'd rather raise an exception manually than triggering one implicitly like this. Just use `raise Exception("this should not cause an error")` or something similar. Would also be kind of self-documenting this way :)

docs/events.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+@on(completed_work_item)
+def send_mail_on_complete_work_item(sender, work_item, **kwargs):
+    if work_item.task.slug == "slug-we-are-interested-in":
+        # send notification
+        pass
+
+
+# It's also possible to specify a sender:
```
winged:
```suggestion
# It's also possible to specify a sender, which refers to a mutation class,
# analog to the `permission_for` decorators
```

caluma/caluma_core/events.py
open-dynaMIX: From the [docs](https://docs.python.org/2/library/logging.html#logging.Logger.exception):

```
This method should only be called from an exception handler.
```

So for using this we would have to raise the exception, then catch it and then log it.

caluma/caluma_workflow/tests/test_events.py
open-dynaMIX: Right! Done.

docs/events.md
open-dynaMIX: Done.

caluma/caluma_core/events.py
winged: Right! Let's keep it as is then

PR #978, opened 2020-03-10
feat: Extend JEXL

winged:
Extend and cleanup JEXL evaluation, so more information can be accessed.

This is considered a proposal. Implementation details, especially regarding the "new" JEXL syntax now introduced, are open for discussion.

New with this PR, you'll have access to an `info` object in question JEXL expressions that provides some useful information:

  - `info.root.form` - Same as `form`, gives the root form's slug
  - `info.form` - The direct form where the question resides. Could be a row
    form if the question is within a table, or a FormQuestion's form, etc.
  - `info.parent.form` - The form above the current form. May be useful in
    deeply nested structures.

While we're at it, I've also started documenting the other things we have, such as the `form` variable and our custom transforms and operators. The idea is to build up a comprehensive JEXL guide that both frontend and backend may refer to for consistency.

I've also added a fixture to build a full document/form structure. This will be used to clean up some existing tests in a follow-up PR.

### DIFF
--- a/caluma/conftest.py
+++ b/caluma/conftest.py
@@ -185,3 +185,98 @@ def permission_classes(settings):
     yield set_permission_classes
     settings.PERMISSION_CLASSES = old_permissions
     config.ready()
+
+
+@pytest.fixture
+def form_and_document(
+    db,
+    form_factory,
+    question_factory,
+    form_question_factory,
+    document_factory,
+    answer_factory,
+):
+    """
+    Return a factory for a form and document.
+
+    The return value is a 4-tuple with the following values:
+        (form, document, questions_dict, answers_dict)
+
+    Optionally add a table question (and a row in the document),
+    and a subform for added complexity.
+
+    The slugs are named as follows:
+
+    * form: top_form
+       * question: top_question
+       * question: table
+           * row_form: row_form
+               * question: column
+       * question: form_question
+           * sub_form: sub_form
+               * question: sub_question
+    """
+
+    def factory(use_table=False, use_subform=False):
+        form = form_factory(slug="top_form")
+        document = document_factory(form=form)
+
+        questions = {}
+        answers = {}
+
+        questions["top_question"] = question_factory(
+            slug="top_question", is_required="true", is_hidden="false"
+        )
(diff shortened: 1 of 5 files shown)


fkm:
Frontend adjustment: https://github.com/projectcaluma/ember-caluma/pull/805

caluma/conftest.py
```diff
@@ -185,3 +185,98 @@ def set_permission_classes(class_names):
     yield set_permission_classes
     settings.PERMISSION_CLASSES = old_permissions
     config.ready()
+
+
+@pytest.fixture
+def form_and_document(
```
open-dynaMIX: This is very handy! I like!

docs/validation.md
```diff
@@ -10,8 +10,72 @@ in the backend and frontend as of its defined small scope. This allows
 frontends in the browser to validate data in exactly the same way as
 the server will, while not limiting validation to simple regexes.
 
-This speeds up user interaction a good deal
+This speeds up user interaction a good deal. JEXL expressions are used in
+the following contexts:
 
+* Deciding whether a field should be visible or not (`is_hidden` property)
+* Deciding whether a field is required (`is_required` property)
+* ...
```
open-dynaMIX: Also:
 - `Task.address_groups`
 - `Flow.next`

But maybe not somethinf to put into `validation.md`.

docs/validation.md
```diff
@@ -10,8 +10,72 @@ in the backend and frontend as of its defined small scope. This allows
...
+* `info` is a data structure containing information about the document and it's
+  form. It's typology is roughly as outlined below. Common access patterns are:
+
+  - `info.root.form` - Same as `form`, gives the root form's slug
+  - `info.form` - The direct form where the question resides. Could be a row
+    form if the question is within a table, or a FormQuestion's form, etc.
+  - `info.parent.form` - The form above the current form. May be useful in
+    deeply nested structures.
+
+Note that the `info` object also contains further information about othe
```
open-dynaMIX:
```suggestion
Note that the `info` object also contains further information about other
```

PR #989, opened 2020-03-19
feat(WorkItem): name and description

### LINKED ISSUE #988: Add name and description to WorkItems
# Rationale

Our goal is to be able to dynamically start `WorkItems` based on a generic `Task`, that can be manually assigned to groups.

For this we need the possibility to set our own `name` and `description`.

## Proposal

We propose to add two new fields to the `WorkItem` model:

1. `name`
2. `description`

If not explicitly set, those should inherit their values from the `Task` instance.

Additionally we need the same two fields on the `saveWorkItem` and `createWorkItem` mutations.

open-dynaMIX:
This PR adds `name` and `description` fields to the `WorkItem`
model and the serializers.

If not explicitly set, those fields inherit their values from their siblings on
the `Task` model.

Additionally it adds `name` to the `WorkItemFilterSet` fields.

Closes #988


caluma/caluma_workflow/models.py
```diff
@@ -236,3 +249,16 @@ class Meta:
...
+
+
+@receiver(post_init, sender=WorkItem)
+def set_name_and_description(sender, instance, **kwargs):
+    """
+    Ensure WorkItem has a name and description set.
+
+    Default to values from Task.
+    """
+    if not any((value for _, value in instance.name.items())):
```
winged: If you're not interested in the key of a dict, you can use `.values()`, turning this line into `if not any(instance.name.values())`

winged: Also you coud *probably* just evaluate `instance.name` in boolean context, which will return `False` if the dict is empty. Not sure if LocalizedField populates the keys by default right now, so we'd have to test this idea

caluma/caluma_workflow/models.py
```diff
@@ -236,3 +249,16 @@ class Meta:
...
+        instance.name = instance.task.name
+    if not any((value for _, value in instance.description.items())):
+        instance.description = instance.task.description
```
winged: Is it guaranteed that a work item always has a task, even at `post_init`?

caluma/caluma_workflow/models.py
open-dynaMIX: Great catch. Now using `.values()`. `bool(instance.name)` doesn't work.

caluma/caluma_workflow/models.py
open-dynaMIX: Yes, you can't save a `WorkItem` without a `Task`.

winged: Right, but a WorkItem can be created step-by-step, which would lead to a crash:
```python
>>> from caluma.caluma_workflow.models import *
>>> wi = WorkItem()
Traceback (most recent call last):
...
caluma.caluma_workflow.models.WorkItem.task.RelatedObjectDoesNotExist: WorkItem has no task.
```
The question is if people using extension point will use this to create work items, and if we want to support this. But for simplicity's sake I think we can keep the code this way.

caluma/caluma_workflow/models.py
```diff
@@ -236,3 +249,16 @@ class Meta:
...
+
+
+@receiver(post_init, sender=WorkItem)
+def set_name_and_description(sender, instance, **kwargs):
+    """
+    Ensure WorkItem has a name and description set.
+
+    Default to values from Task.
+    """
+    if not any((value for value in instance.name.values())):
```
winged: You don't need the `value for value in` bit. `any(instance.name.values())` works just the same :)

caluma/caluma_workflow/models.py
open-dynaMIX: Of course! Using `pre_save` now.

caluma/caluma_workflow/models.py
open-dynaMIX: :man_facepalming:  Done.

PR #1009, opened 2020-04-02
fix(graph*-3.0): only accepts integer or string as id

ganwell:
#814 creates a lot of chicken and egg problems. I want to at least push simple fixes to caluma.

The graph*-3.0 projects try to validate all inputs and prevent all implicit conversion.


ganwell:
@winged There are many places, but I fix bugs in order. Of the 20 bugs I have in my branch, only this one had not connections to other problems and could be cleanly pushed.

PR #1020, opened 2020-04-14
feat: make created_by_group settable

### LINKED ISSUE #614: Make "created_by_group" settable
# Rationale

For chosing the group used for the `created_by_group` field, Caluma just uses the first group in the list it receives from the OIDC provider.

We propose, that it should be possible to manually choose the group.

## Implementation

It is already possible to modify the `OIDCUser` object in the validations. Right now, to manually set the group, one needs to make sure the desired group is at the start of the list `OIDCUser.groups`.

We propose to provide a method `OIDCUser.set_group()`  to manually set the group.

open-dynaMIX:
This commit simplifies manual setting of the `created_by_group` field
from within validations.

Before, in order to override the chosen group, one had to prepend it to
the list `OIDCUser.groups`. Now, one can just set the property:
`user.group = "foo"`.

Closes #614

### DIFF
--- a/caluma/caluma_core/tests/test_validations.py
+++ b/caluma/caluma_core/tests/test_validations.py
@@ -1,7 +1,7 @@
 import pytest
 from django.core.exceptions import ImproperlyConfigured
 
-from .. import models, serializers
+from .. import models, mutation, serializers, types, validations
 from ..mutation import Mutation
 from ..validations import BaseValidation, validation_for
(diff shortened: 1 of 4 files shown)


czosel:
Looks good! Would there be some place where we could add documentation for this?

open-dynaMIX:
@czosel Done.

czosel:
Docs look good as well, thanks! :+1:

PR #1026, opened 2020-04-16
fix(jexl): evaluate boolean expressions lazily

winged:
In most programming languages, boolean expressions (AND, &&; OR, ||) are
evaluated lazily. We also want this here!

The common use case is checking if an optional question has a value
before using it in a comparison, for example.

### DIFF
--- a/caluma/caluma_core/jexl.py
+++ b/caluma/caluma_core/jexl.py
@@ -2,6 +2,7 @@ from itertools import count
 
 import pyjexl
 from pyjexl.analysis import ValidatingAnalyzer
+from pyjexl.evaluator import Context, Evaluator
 from pyjexl.exceptions import ParseError
 from rest_framework import exceptions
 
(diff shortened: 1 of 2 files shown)


winged:
Discussion point: It seems that the JS implementation (https://github.com/TomFrost/Jexl/blob/master/lib/evaluator/handlers.js#L27) also does eager evaluation for logical expressions. So I'm wondering whether the correct way would instead be to emulate the JS behaviour to avoid those TypeErrors.

czosel:
I think this should be implemented inside JEXL, not overwritten in caluma. We can always pull in a fork of JEXL until it is merged.

winged:
> I think this should be implemented inside JEXL, not overwritten in caluma. We can always pull in a fork of JEXL until it is merged. And if JEXL for JS also does eager evaluation, the change would have to land there first. Since this seems like quite a big step, we might also want to consider other solutions like the `default` transform we discussed earlier.

Agree, but then again, the JS implementation doesn't do shortcuts either, so not sure what the standard should be, and how easy it'd be to get upstream.

I intend to put up an issue with pyjexl to see what they think - just didn't get around to it just now

czosel:
The stabdard is defined by the JS version - so we’d have to ask there if this (breaking) change is something they’d consider merging. If not, then is is also not a way to move forward for us, imo.

winged:
Issue in upstream JEXL: https://github.com/TomFrost/Jexl/issues/79

czosel:
Just another thought: We would also get away with 
```
"foo"|answer !== null && 15 > ("foo"|answer || 0)
```
right? (default transform is not even needed)

winged:
> Just another thought: We would also get away with
> 
> ```
> "foo"|answer !== null && 15 > ("foo"|answer || 0)
> ```
> 
> right? (default transform is not even needed)

This would work, yes (apart from the `!==` operator not being supported)

czosel:
Right, seems that comparisons are strongly typed in JEXL by default. I'll update the example.

czosel:
Let's close this for now, since we probably won't move forward with it.

winged:
Good news! TomFrost just merged this in his JEXL library, and following this I've sent a PR to the PyJEXL implementation as well:

* https://github.com/TomFrost/Jexl/pull/84
* https://github.com/mozilla/pyjexl/pull/23

PR #1038, opened 2020-04-28
feat(filters): add filters for object creation dates

winged:
Introduce filters `createdBefore` and `createdAfter`. Those can be used to limit
results to objects created within a given time range when used in combination.

Note that The "After" filter is inclusive, while the "Before" filter is exclusive.
This is done so that clients don't need to fiddle with filter values when they
want to filter for one specific day, for example.

/cc @fkm


winged:
> Awesome! Please make gitlint happy (or make it ignore B1 rofl ), then we can merge this.

Fixed :)

PR #1046, opened 2020-05-05
feat(arch): move business logic for case creation to custom manager

czosel:
When caluma is used as a django app, it is often desireable to trigger specific mutations by calling python APIs directly (instead of the GraphQL API). This shows how we could go about offering a Python API for starting cases that mirrors what the GraphQL API offers via `saveCase`: We extract the business logic out of the serializer into a custom manager.

### DIFF
--- a/caluma/caluma_workflow/managers.py
+++ b/caluma/caluma_workflow/managers.py
@@ -1,0 +1,92 @@
+from django.db.models import Manager
+
+from caluma.caluma_core.events import send_event
+from caluma.caluma_user.models import BaseUser
+
+from ..caluma_form.models import Document
+from . import events, models, utils
+
+
+class CaseManager(Manager):
+    def _validate(self, data):
+        form = data.get("form")
+        workflow = data.get("workflow")
+
+        if form:
+            if (
+                not workflow.allow_all_forms
+                and not workflow.allow_forms.filter(pk=form.pk).exists()
+            ):
+                raise Exception(
+                    f"Workflow {workflow.pk} does not allow to start case with form {form.pk}"
+                )
+
+        return data
+
+    def _pre_create(self, validated_data, user):
+        parent_work_item = validated_data.get("parent_work_item")
+        validated_data["status"] = models.Case.STATUS_RUNNING
+
+        form = validated_data.pop("form", None)
+        if form:
+            validated_data["document"] = Document.objects.create(
+                form=form, created_by_user=user.username, created_by_group=user.group
+            )
+
+        if parent_work_item:
+            case = parent_work_item.case
+            while hasattr(case, "parent_work_item"):
+                case = case.parent_work_item.case
+            validated_data["family"] = case
+
+        return validated_data
+
+    def _post_create(self, case, user, parent_work_item):
+        # Django doesn't save reverse one-to-one relationships automatically:
+        # https://code.djangoproject.com/ticket/18638
+        if parent_work_item:
+            parent_work_item.child_case = case
+            parent_work_item.save()
+
+        workflow = case.workflow
+        tasks = workflow.start_tasks.all()
+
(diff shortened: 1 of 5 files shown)


czosel:
@winged @open-dynaMIX ready for review

caluma/caluma_workflow/serializers.py
```diff
@@ -229,56 +171,24 @@ class CaseSerializer(SendEventSerializerMixin, serializers.ModelSerializer):
...
-            validated_data["document"] = Document.objects.create(
-                form=form, created_by_user=user.username, created_by_group=user.group
-            )
 
-        if parent_work_item:
-            case = parent_work_item.case
-            while hasattr(case, "parent_work_item"):
-                case = case.parent_work_item.case
-            validated_data["family"] = case
+        validated_data = models.Case.objects._pre_create(validated_data, user)
```
winged, line 185: I don't like this 3-step creation procedure. Users *may* forget to do it exactly this way (granted, much better than before, but still)

Also, underscore-prefixed names in python denote private methods, so this looks very much like a violation to me. Especially since we want to simplify the procedure for people using our code directly, this is a bad idea IMHO

caluma/caluma_workflow/serializers.py
```diff
@@ -229,56 +171,24 @@ class CaseSerializer(SendEventSerializerMixin, serializers.ModelSerializer):
...
-
-        workflow = instance.workflow
-        tasks = workflow.start_tasks.all()
-
-        work_items = bulk_create_work_items(tasks, instance, user)
-
-        self.send_event(events.created_case, case=instance)
-        for work_item in work_items:  # pragma: no cover
-            self.send_event(events.created_work_item, work_item=work_item)
+        case = super().create(validated_data)
```
winged, line 187: Instead of doing the "pre_create", "create", "post_create" dance, we should call 
`case = models.Case.objects.start(...)` instead, cleaning up the parameters so the method can properly work with it.

caluma/caluma_workflow/serializers.py
```diff
@@ -229,56 +171,24 @@ class CaseSerializer(SendEventSerializerMixin, serializers.ModelSerializer):
...
-        if form:
-            if (
-                not workflow.allow_all_forms
-                and not workflow.allow_forms.filter(pk=form.pk).exists()
-            ):
-                raise exceptions.ValidationError(
-                    f"Workflow {workflow.pk} does not allow to start case with form {form.pk}"
-                )
+        try:
+            data = models.Case.objects._validate(data)
```
winged, line 175: This is a private method call from "outside", which looks dangerous. See my other comment regarding validation

caluma/caluma_workflow/managers.py
```diff
@@ -0,0 +1,92 @@
...
+
+from caluma.caluma_core.events import send_event
+from caluma.caluma_user.models import BaseUser
+
+from ..caluma_form.models import Document
+from . import events, models, utils
+
+
+class CaseManager(Manager):
+    def _validate(self, data):
```
winged, line 11: Django has infrastructure for model-level validation: https://docs.djangoproject.com/en/3.0/ref/models/instances/#django.db.models.Model.clean_fields . 
I suggest using this instead of using a custom method

caluma/caluma_workflow/managers.py
```diff
@@ -0,0 +1,92 @@
...
+        form = data.get("form")
+        workflow = data.get("workflow")
+
+        if form:
+            if (
+                not workflow.allow_all_forms
+                and not workflow.allow_forms.filter(pk=form.pk).exists()
+            ):
+                raise Exception(
```
winged, line 20: Use a `ValidationError` here (`from django.core.exceptions import ValidationError`)

czosel:
Closing in favor of #1051

PR #1051, opened 2020-05-08
feat(architecture): expose start case logic in public python api module

anehx:
When caluma is used as a django app, it is often desireable to trigger
specific mutations by calling python APIs directly (instead of the
GraphQL API). This shows how we could go about offering a Python API for
starting cases that mirrors what the GraphQL API offers via saveCase: We
extract the business logic out of the serializer into a domain logic
class which is then used in a public method exposed in an api module.

Closes #1046

### DIFF
--- a/caluma/caluma_workflow/api.py
+++ b/caluma/caluma_workflow/api.py
@@ -1,0 +1,41 @@
+from typing import Optional
+
+from caluma.caluma_form.models import Form
+from caluma.caluma_user.models import BaseUser
+
+from . import domain_logic, models
+
+
+def start_case(
+    workflow: models.Workflow,
+    user: BaseUser,
+    form: Optional[Form] = None,
+    parent_work_item: Optional[models.WorkItem] = None,
+    **kwargs
+) -> models.Case:
+    """
+    Start a case of a given workflow (just like `saveCase`).
+
+    Similar to Case.objects.create(), but with the same business logic as
+    used in the `saveCase` mutation.
+
+    Usage example:
+    ```py
+    start_case(
+        workflow=Workflow.objects.get(pk="my-wf"),
+        form=Form.objects.get(pk="my-form")),
+        user=AnonymousUser()
+    )
+    ```
+    """
+    data = {"workflow": workflow, "form": form, "parent_work_item": parent_work_item}
+
+    data.update(kwargs)
+
+    validated_data = domain_logic.StartCaseLogic.pre_start(
+        domain_logic.StartCaseLogic.validate_for_start(data), user
+    )
+
+    case = models.Case.objects.create(**validated_data)
+
+    return domain_logic.StartCaseLogic.post_start(case, user, parent_work_item)
--- a/caluma/caluma_workflow/domain_logic.py
+++ b/caluma/caluma_workflow/domain_logic.py
@@ -1,0 +1,74 @@
+from django.core.exceptions import ValidationError
+
+from caluma.caluma_core.events import send_event
+
+from ..caluma_form.models import Document
+from . import events, models, utils
+
+
+class StartCaseLogic:
+    """
+    Shared domain logic for starting cases.
+
+    Used in the `saveCase` mutation and in the `start_case` API. The logic
+    for case creation is split in three parts (`validate_for_start`,
+    `pre_start` and `post_start`) so that in between the appropriate create
+    method can be called (`super().create(...)` for the serializer and
+    `Case.objects.create(...) for the python API`).
+    """
+
+    @staticmethod
+    def validate_for_start(data):
+        form = data.get("form")
+        workflow = data.get("workflow")
+
+        if form:
+            if (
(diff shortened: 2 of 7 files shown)


caluma/caluma_workflow/api.py
```diff
@@ -0,0 +1,24 @@
+from . import domain_logic
+
+
+def start_case(*args, **kwargs):
```
winged, line 4: Please use `*args` and `**kwargs` only for optional or unknown parameters. Known and required parameters should be named (optional ones should have a default value).

In this case, `workflow` is required, `form` is known-but-optional, as is `user`. Any others shold be documented

caluma/caluma_workflow/domain_logic.py
```diff
@@ -0,0 +1,96 @@
...
+                and not workflow.allow_forms.filter(pk=form.pk).exists()
+            ):
+                raise ValidationError(
+                    f"Workflow {workflow.pk} does not allow to start case with form {form.pk}"
+                )
+
+        return data
+
+    @staticmethod
+    def pre_start(validated_data, user):
```
winged, line 28: I'd like to see better naming: What does this method *do*?

winged: Also, some short docstring would be nice, even if this is not "public API"

caluma/caluma_workflow/domain_logic.py
```diff
@@ -0,0 +1,96 @@
...
+        if parent_work_item:
+            case = parent_work_item.case
+            while hasattr(case, "parent_work_item"):
+                case = case.parent_work_item.case
+            validated_data["family"] = case
+
+        return validated_data
+
+    @staticmethod
+    def post_start(case, user, parent_work_item):
```
winged, line 47: Same here: naming, docstring

caluma/caluma_workflow/domain_logic.py
```diff
@@ -0,0 +1,96 @@
...
+        Start a case of a given workflow (just like `saveCase`).
+
+        Similar to Case.objects.create(), but with the same business logic as used in
+        the `saveCase` mutation.
+
+        All model relationships must be passed as resolved instances instead of primary keys.
+
+        Usage example:
+        ```py
+        Case.objects.start(
```
winged, line 79: Example is not correct anymore, this assumes this method still resides on the case manager

caluma/caluma_workflow/domain_logic.py
winged: Maybe it'd make sense to explain *why* we're doing this 3-step dance. Could be a docstring on the `StartCaseLogic` class, for example. Just so future readers will know

caluma/caluma_workflow/api.py
anehx: Fair point. However, `user` is required as well. But I won't document all `**kwargs` since that's just what will be passed to `Case.objects.create`.. Or do I missunderstand you?

caluma/caluma_workflow/domain_logic.py
anehx: What? I thought we discussed this naming through yesterday! :disappointed:

caluma/caluma_workflow/domain_logic.py
anehx: That docstring should not be here at all. :grinning:  We agreed to only document the public API, which I did.

caluma/caluma_workflow/domain_logic.py
anehx: This is exactly as discussed and agreed yesterday. Also I think I'm the wrong person to write that docstring (which I think is unnecessary) since I just moved the code from the serializer (where it was undocumented) to here.

caluma/caluma_workflow/domain_logic.py
czosel: I actually expected the entire `start` method to be moved to `api.py`, because it's only called there.

anehx: Ok

anehx: Done

caluma/caluma_workflow/api.py
anehx: Done

caluma/caluma_workflow/domain_logic.py
czosel: I think `pre_start` and `post_start` are acceptable method names. I agree though that a docstring for `StartCaseLogic` might help, maybe something along the lines of

> Shared domain logic for starting cases.
>
> Used in the `SaveCase` mutation and in the `start_case` api. The logic for case creation is split in two parts (`pre_start` and `post_start`) so that in between the appropriate `create` method can be called (the serializer calls `super.create`).

anehx: Done

PR #1052, opened 2020-05-08
feat(architecture): expose complete and skip work item to public API

anehx:


caluma/caluma_workflow/api.py
```diff
@@ -39,3 +39,47 @@ def start_case(
...
 
     return domain_logic.StartCaseLogic.post_start(case, user, parent_work_item)
+
+
+def complete_work_item(work_item: models.WorkItem, user: BaseUser) -> models.WorkItem:
+    """
+    Complete a work item (just like `completeWorkItem`).
+
+    Usage example:
+    ```py
```
winged, line 49: In python docstrings, examples are usually denoted not in markdown syntax, but using this:
```
>>> complete_work_item(
...     work_item=models.WorkItem.first(),
...     user=AnonymousUser()
... )
WorkItem object xyz
```
(Whereas the last line represents the return value). This is mostly how it would look when typed into the python interpreter.

caluma/caluma_form/validators.py
```diff
@@ -184,7 +184,7 @@ def validate(self, *, question, document, info, validation_context=None, **kwarg
 
 
 class DocumentValidator:
-    def validate(self, document, info, validation_context=None, **kwargs):
+    def validate(self, document, user, validation_context=None, **kwargs):
```
winged, line 187: I love that you refactored that dreaded `info` out of the validator! I wanted to get rid of it for a while now :)

caluma/caluma_workflow/api.py
```diff
@@ -39,3 +39,47 @@ def start_case(
...
+
+    return work_item
+
+
+def skip_work_item(work_item: models.WorkItem, user: BaseUser) -> models.WorkItem:
+    """
+    Skip a work item (just like `SkipWorkItem`).
+
+    Usage example:
+    ```py
```
winged, line 71: Same applies here regarding docstring syntax

caluma/caluma_workflow/serializers.py
```diff
@@ -256,105 +256,61 @@ def update(self, instance, validated_data):
...
-            case=self.instance.case,
-            document=self.instance.document,
-            info=self.context["info"],
-        )
-        data["status"] = models.WorkItem.STATUS_COMPLETED
-        data["closed_at"] = timezone.now()
-        data["closed_by_user"] = user.username
-        data["closed_by_group"] = user.group
-        return super().validate(data)
+        try:
```
winged, line 263: I think the try/except isn't really needed here - or why do you catch a validation error just to raise one again (as far as I know, they're the same type anyway, despite coming from different modules)

caluma/caluma_workflow/api.py
anehx: Could you attach a link to the documentation of that standard?

caluma/caluma_workflow/serializers.py
anehx: Not sure about that: https://github.com/encode/django-rest-framework/blob/f0dbf0a264677f2a53faab402ff49f442fc4383a/docs/community/3.0-announcement.md#using-serializersvalidationerror

anehx: "whenever raising validation errors you should prefer using the `serializers.ValidationError` exception class, and not Django's built-in exception."

winged: Fair enough! Keep it as is then

caluma/caluma_workflow/api.py
winged: https://docs.python-guide.org/writing/documentation/#docstrings-and-magic

It's more convention, as the `doctest` module uses the same syntax - even though we don't actually use `doctest`

anehx: Perfect thanks, I'll change it!

caluma/caluma_workflow/api.py
anehx: Fixed

caluma/caluma_workflow/api.py
anehx: Fixed

czosel:
Any idea why the test is failing?

winged:
> Any idea why the test is failing?

I think it could be some implicit ordering in the query of the test: There's 3 documents, but only two of them are inspected in the response. We could retry and hope, or try to explicitly order the documents in that test

czosel:
The retry seems to have helped, even though it's not displayed correctly in travis status in this MR.

winged:
I've changed the failing test so it won't rely on result ordering anymore. Semantics of the test stayed the same.

PR #1059, opened 2020-05-18
feat(allCases): add documentForms filter for allCases graph

anehx:
Since a workflow can have multiple allowed forms it makes sense to allow
filtering for a set of document forms.

### DIFF
--- a/caluma/caluma_workflow/filters.py
+++ b/caluma/caluma_workflow/filters.py
@@ -9,6 +9,7 @@ from ..caluma_core.filters import (
     MetaFilterSet,
     OrderingFilter,
     SearchFilter,
+    SlugMultipleChoiceFilter,
     StringListFilter,
     generate_list_filter_class,
 )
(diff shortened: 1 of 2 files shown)


caluma/caluma_workflow/filters.py
```diff
@@ -79,6 +80,7 @@ class CaseFilterSet(MetaFilterSet):
     order_by = OrderingFilter(label="CaseOrdering", fields=("status",))
 
     document_form = CharFilter(field_name="document__form_id")
+    document_forms = GlobalIDMultipleChoiceFilter(field_name="document__form_id")
```
winged, line 83: The `GlobalIDMultipleChoiceFilter` is not the correct filter type: Forms have slugs, so the filter parameter should be string. The global id filters apply additional magic to parse the global ID - and in this case it also has the wrong type:

`documentForm: String` vs `documentForms: [ID]`

There's a `SlugMultipleChoiceFilter` that would be suitable for the given purpose, I suggest using that one instead

anehx: Done

PR #1066, opened 2020-05-29
chore: remove version pinning for subdependencies

sbor23:
graphene-django does a good job pinning the version for graphql-core,
graphene and graphql-relay.
So we remove them from our requirements and let graphene-django handle
these.

### DIFF
--- a/caluma/caluma_core/tests/test_inversible_filters.py
+++ b/caluma/caluma_core/tests/test_inversible_filters.py
@@ -27,7 +27,7 @@ def test_inversible_with_vars(db, schema_executor, document_factory):
         }
     """
 
-    result = schema_executor(query, variables=variables)
+    result = schema_executor(query, variable_values=variables)
     assert not result.errors
 
     result_ids = [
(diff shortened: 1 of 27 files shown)


sbor23:
Ooops the CI proves me wrong.
Seems like this doesn't work even with `graphene-django==2.10.1`

sbor23:
@ganwell I figured out you are working on the whole *3 upgrade thingy which should be just around the corner, right?
So I won't dive deeper into the issues here, even though I can't understand why we still have these DeprecationWarnings as documented [here](https://github.com/graphql-python/graphene-django/issues/893) (but they should be fixed upstream since 2.9.0).

I just leave this open so we can maybe remove the pinning (or at least discuss it) when the upgrade is complete. :)

ganwell:
@sbor23 Actually, it is the reverse. I think it is a good thing to remove the pinning and fix issues in master. It would help me a lot, because these bugs pile-up together with the 300 failing tests, when we switch to 3.0.

ganwell:
@sbor23 I've never seen this warning in 3.0, still it would be nice to update in master. I guess it is a single fix, that will restore most of the tests.

PR #1067, opened 2020-05-29
fix(historical): use matching question for historicalanswer

sbor23:
Use correct historical question for finding answer type.

### DIFF
--- a/caluma/caluma_form/schema.py
+++ b/caluma/caluma_form/schema.py
@@ -25,43 +25,13 @@ from .validators import get_document_validity
 
 
 def resolve_answer(answer):
-    question_answer_types = {
-        models.Question.TYPE_MULTIPLE_CHOICE: ListAnswer,
-        models.Question.TYPE_INTEGER: IntegerAnswer,
-        models.Question.TYPE_FLOAT: FloatAnswer,
-        models.Question.TYPE_DATE: DateAnswer,
-        models.Question.TYPE_CHOICE: StringAnswer,
-        models.Question.TYPE_TEXTAREA: StringAnswer,
-        models.Question.TYPE_TEXT: StringAnswer,
-        models.Question.TYPE_TABLE: TableAnswer,
-        models.Question.TYPE_FILE: FileAnswer,
-        models.Question.TYPE_DYNAMIC_CHOICE: StringAnswer,
-        models.Question.TYPE_DYNAMIC_MULTIPLE_CHOICE: ListAnswer,
-    }
     if answer.question.type == models.Question.TYPE_STATIC:
         raise Exception('Questions of type "static" should never have an answer!')
-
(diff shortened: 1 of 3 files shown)


sbor23:
I'm just leaving this here...
https://github.com/treyhunner/django-simple-history/issues/397

open-dynaMIX:
> I'm just leaving this here...
> [treyhunner/django-simple-history#397](https://github.com/treyhunner/django-simple-history/issues/397)

https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/historical_schema.py#L33 ;)

open-dynaMIX:
Will have a closer look next week.

sbor23:
> > I'm just leaving this here...
> > [treyhunner/django-simple-history#397](https://github.com/treyhunner/django-simple-history/issues/397)
> 
> https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/historical_schema.py#L33 ;)

Hahaha how did I not realize this :D

caluma/caluma_form/tests/test_history.py
```diff
@@ -372,3 +372,131 @@ def test_historical_table_answer(
     result = schema_executor(historical_query, variables=variables)
     assert not result.errors
     snapshot.assert_match(result.data)
+
+
+def test_history_question_answer(
```
winged, line 377: Can you add a comment on what aspect is being tested here?

caluma/caluma_form/schema.py
```diff
@@ -963,3 +933,34 @@ def resolve_all_format_validators(self, info):
...
+    models.Question.TYPE_CHOICE: StringAnswer,
+    models.Question.TYPE_TEXTAREA: StringAnswer,
+    models.Question.TYPE_TEXT: StringAnswer,
+    models.Question.TYPE_TABLE: TableAnswer,
+    models.Question.TYPE_FILE: FileAnswer,
+    models.Question.TYPE_DYNAMIC_CHOICE: StringAnswer,
+    models.Question.TYPE_DYNAMIC_MULTIPLE_CHOICE: ListAnswer,
+}
+
+QUESTION_OBJECT_TYPE = {
```
winged, line 952: Good work on refactoring out those lists! However I'd prefer this being plural named (`QUESTION_OBJECT_TYPES`)

caluma/caluma_form/tests/test_history.py
sbor23: Gave it a better name instead

winged: I'd still like something along the lines of "test that historical question types retain the right type even when the question type for a given slug changes" or something similar. So you know what the intention is behind the test (I know I should do that more as well...)

sbor23: Did that

PR #1071, opened 2020-06-02
fix(form): make HistoricalDocument inherit from Document

### LINKED ISSUE #1070: HistoricalDocument not subclassing Document type
Is there a specific reason why `HistoricalDocument` isn't a subclass of `Document`, like `HistoricalAnswer` subclasses `Answer` and every other type?

https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/historical_schema.py#L145

There are cases where this would be convenient, such as visibilities and other extensions would need only one definition for handling both.

sbor23:
HistoricalDocument is of type Document.
This way, we can reuse the Document type for visibilities and such,
applying the same rules to HistoricalDocument if we have a rule for
Document.

fixes #1070

Caution: This is probably a breaking change, if an extension (eg. visibility) expects that a `@filter_queryset_for` rule for Document doesn't apply for HistoricalDocument (which would use the rule of a fallback, probably Node or no filter).

### DIFF
--- a/caluma/tests/snapshots/snap_test_schema.py
+++ b/caluma/tests/snapshots/snap_test_schema.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 
 from snapshottest import Snapshot
 
-
 snapshots = Snapshot()
 
 snapshots[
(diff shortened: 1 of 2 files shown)


caluma/tests/snapshots/snap_test_schema.py
```diff
@@ -847,6 +846,7 @@
   source: Document
   historyDate: DateTime!
   historyType: String
+  answers(before: String, after: String, first: Int, last: Int, metaValue: [JSONValueFilterType], orderBy: [AnswerOrdering], filter: [AnswerFilterSetType], order: [AnswerOrderSetType], question: ID, search: String, createdByUser: String, createdByGroup: String, createdBefore: DateTime, createdAfter: DateTime, metaHasKey: String, questions: [ID], visibleInContext: Boolean): AnswerConnection
```
winged, line 849: I think this could be problematic - It will show the "current" answers, not the historical ones. Thus I'd like to see the `answers` connection being excluded in the GQL schema. (The taxonomy change was only to allow visibility reuse, right?)

open-dynaMIX: That's exactly the reason I didn't inherit from `Document`. Tthis change should not result in any changes to the schema.

sbor23: I was fearing this. So this isn't a viable change I guess.

open-dynaMIX: Yes. And this doesn't only apply to `HistoricalDocument`, but also to `HistoricalFile`.

PR #1080, opened 2020-06-05
chore: create mailmap

ganwell:
* only people who already commited with @adfinis.com are mapped to @adfinis.com

* also fixing wrong names, so there are a few mappings with the same e-mail


open-dynaMIX:
I don't have a very good feeling having a file with our email addresses in the repo. I'm not really affected by this, because there are "just" work emails addresses from me. One could say, as soon as you make a commit with an email address, this becomes public information. Still, I think it would be good to gather some more opinions on this.

ganwell:
@open-dynaMIX ,well our mail addresses are all over the place, I took address directly from the repo using git  `shortlog` -e -s -n.`

*I added no new information to the repository.*

@sbor23 I thought we fix these, once people start using @adfinis.con. I for example not will start @dfinis.com using in a year or so. I put the ganwell@fangorn.ch -> jean-louis.fuchs@adfinis.sch rules in to all open source projects recententy, so I will not switch in a awile.

PR #1094, opened 2020-06-20
perf(caluma_form): add history_question_type to HistoricalAnswer

sbor23:
Save the current question type to the historical answer upon save.
This resolves a performance issue, figuring out the question type for
each HistoricalAnswer causing a query.


czosel:
Can you quantify the impact this has?

sbor23:
Addressed the signal registration, as well as moving history to an ABC.

sbor23:
This is verified to resolve the performance issue with our internal project.

However the migration is now having the same performance problems as the feature used to have, making it reeaally ressource intensive.
On a project with 366496 answers this is running for more than 30 minutes now...
I don't see how we can speed it up though...

czosel:
Oh, damn - this would require extra coordination when this would be deployed.

caluma/caluma_form/models.py
```diff
@@ -291,7 +294,26 @@ class Meta:
...
+    )
+
+    class Meta:
+        abstract = True
+
+
+class Answer(core_models.BaseModel):
+    """Records an answer to a question of arbitrary type.
+
+    We need to replicate the UUIDModel in order to register it as historical model.
```
winged, line 311: I think this is better put in a comment rather than the docstring. No need to put implementation detail into "public" docstrings.

caluma/caluma_workflow/tests/snapshots/snap_test_work_item.py
```diff
@@ -78,162 +78,6 @@
     }
 }
 
-snapshots[
```
winged, line 81: Why did those snapshots disapper? They don't seem to be related to the change of this PR

sbor23: Huh that's odd. I would assume that the CI would complain about this?
No idea tbh

caluma/caluma_form/migrations/0035_historicalanswer_history_question_type.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations, models
+
+
+def add_initial_historical_question_type(apps, schema_editor):
+    HistoricalAnswer = apps.get_model("caluma_form", "HistoricalAnswer")
+    HistoricalQuestion = apps.get_model("caluma_form", "HistoricalQuestion")
+
+    for answer in HistoricalAnswer.objects.all():
```
winged, line 11: This *might* be possible to optimize by doing a raw SQL `UPDATE` statement. It would need some ugly `JOIN` magic but should speed things up considerably.

Something like
```sql
update caluma_historical_answer ha
left join (
    select * from caluma_historical_question hq on (
        hq.slug = ha.question_id and history_date <= ha.history_date
    )
    order by history_date desc
    limit 1
) hq_last
set ha.history_question_type = hq_last.type
where hq_last.type is not null
```
followed up by a simplified `update` to fill in the remaining answers where no historical question was found:
```sql
update caluma_historical_answer ha
left join caluma_form_question q on (q.slug = ha.question_id)
set ha.history_question_type = q.type
where ha.history_question_type is null
```

sbor23: Huh that's clever, I was thinking of an update but wouldn't figure out the ordering / limit 1 part. I will give it a try!

Actually the fallback should theoretically not be necessary, but it might be that downstream projects did not setup the history properly (like we did...).

Thanks!

caluma/caluma_form/models.py
sbor23: Done

caluma/caluma_workflow/tests/snapshots/snap_test_work_item.py
winged: Maybe someone didn't clean them up before? But I'd still investigate this

caluma/caluma_form/migrations/0035_historicalanswer_history_question_type.py
sbor23: Ok just measured the whole thing, brings it down to 30 seconds!

winged: Awesome :)

PR #1097, opened 2020-06-25
feat: add --dangling option for cleanup_history command, drop --keep option

sbor23:
Add option -d, --dangling, cleaning up instances with missing relation.
For example, if a question was deleted, the historical answers
referencing this question might have a broken reference.
Passing -d will include theses instances for deletion.

We drop --keep, this functionality is now provided by upstream simple_history's `clean_old_history`

### DIFF
--- a/caluma/caluma_core/management/commands/cleanup_history.py
+++ b/caluma/caluma_core/management/commands/cleanup_history.py
@@ -1,7 +1,16 @@
 from dateparser import parse
 from django.core.management.base import BaseCommand
+from django.db.models import Q
 from django.utils import timezone
 from simple_history.models import registered_models
+
+# Instances which are created at run-time, referencing config-time models
+# eg. Answers are created at run-time by users and referencing Question
+RELATED_HISTORICAL_MODELS = {
+    "Answer": {"question__type__isnull": True},
+    "Document": {"form__name__isnull": True},
+    "DynamicOption": {"question__type__isnull": True},
+}
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


sbor23:
Hmm just realized this CLI option doesn't make too much sense the way it is now.
I think if the --dangling option is set, ALL dangling instances should be remove, regardless if the a date was given in --keep or not.
There's no case where one would want to keep those instances.

sbor23:
> Hmm just realized this CLI option doesn't make too much sense the way it is now.
> I think if the --dangling option is set, ALL dangling instances should be remove, regardless if the a date was given in --keep or not.
> There's no case where one would want to keep those instances.

Made the --dandling option independent from the --keep option, so it's all or nothing.

sbor23:
As just discussed with @open-dynaMIX I will remove the "--keep" part of the command since this is now supported by simple_history itself: https://github.com/jazzband/django-simple-history/pull/675

winged:
> As just discussed with @open-dynaMIX I will remove the "--keep" part of the command since this is now supported by simple_history itself: [jazzband/django-simple-history#675](https://github.com/jazzband/django-simple-history/pull/675)

Nice! However in that case it'd be a good idea to update the docs to reflect this (and extend it a little, while we're at it):
`caluma/docs/historical-records.md`

sbor23:
> > As just discussed with @open-dynaMIX I will remove the "--keep" part of the command since this is now supported by simple_history itself: [jazzband/django-simple-history#675](https://github.com/jazzband/django-simple-history/pull/675)
> 
> Nice! However in that case it'd be a good idea to update the docs to reflect this (and extend it a little, while we're at it):
> `caluma/docs/historical-records.md`

Good point! Did that.

czosel:
@winged wanna take another look?

winged:
You may want to rebase - the diff looks horrible on Github

winged:
Seems like one of the snapshots failed. Can you update the PR and push again? Then I think it's ready to merge

PR #1098, opened 2020-06-25
feat(workflow): allow dynamic groups for group JEXL transform

anehx:
This commits adds the possibility to create custom dynamic groups which
can be used in the `control_groups` and `address_groups` of a task. The
transform will be evaluated when creating the work item.


docs/dynamic-groups.md
```diff
@@ -0,0 +1,37 @@
+# Dynamic Groups
+
+Caluma allows you to define custom dynamic groups to use in the
+`address_groups` and `control_groups` properties of a `Task`. Those dynamic
+groups can be used to e.g. define role based groups.
```
czosel, line 5:
```suggestion
Caluma allows you to write code that returns groups for use in the
`address_groups` and `control_groups` properties of a `Task`. A typical use-case for such "dynamic groups" would be to define role based groups.
```

docs/dynamic-groups.md
```diff
@@ -0,0 +1,37 @@
...
+)
+
+
+class CustomDynamicGroups(BaseDynamicGroups):
+    @register_dynamic_group("client_administrator")
+    def resolve_client_administrator(self, task, case, user, prev_work_item):
+        if case.workflow.slug == "workflow-1":
+            return "some-administrator-group"
+
+        return "some-other-administrator-group"
```
czosel, line 30:
```suggestion
    @register_dynamic_group("legal_dept")
    def resolve_legal_dept(self, task, case, user, prev_work_item):
        # custom business logic ...
        return find_legal_dept(case)
```

docs/dynamic-groups.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+You can then use that defined dynamic group `"client_administrator"` in the
+`Task` properties by postfixing them with a transform:
+`["client_administrator"|dynamic_group]`. The created `WorkItem` would then
+have an evaluated value `["some-administrator-group"]` or
+`["some-other-administrator-group"]` for that property.
```
czosel, line 37:
```suggestion
You can then use that defined dynamic group `"legal_dept"` in the
`Task` properties by postfixing them with a transform:
`"legal_dept"|groups`. The created `WorkItem` would then
have an evaluated value like `["legal-dept-boston"]` or
`["legal-dept-nyc"]` for that property.
```

caluma/caluma_workflow/jexl.py
```diff
@@ -12,6 +12,11 @@ def visit_Transform(self, transform):
         if transform.name == "groups" and not isinstance(transform.subject.value, list):
             yield f"{transform.subject.value} is not a valid list of groups."
 
+        if transform.name == "dynamic_group" and not isinstance(
+            transform.subject.value, str
+        ):
+            yield f"{transform.subject.value} is not a valid dynamic group name."
```
czosel, line 18: As discussed: let's use the existing `groups` transform instead (check if it is dynamic group, otherwise treat as regular group).

docs/dynamic-groups.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Configuration
+
+To configure a custom dynamic groups class, you can set the `DYNAMIC_GROUPS_CLASSES` environment variable.
+
+When using the Caluma service, you can place your class in
+`caluma.extensions.dynamic_groups` and set `DYNAMIC_GROUPS_CLASSES` to
+`["caluma.extensions.dynamic_groups.CustomDynamicGroups"]`.
+
+## Example dynamic groups class
```
czosel, line 15: Reminder: Can return list of groups or single group.

caluma/caluma_workflow/jexl.py
```diff
@@ -54,7 +62,25 @@ def __init__(self, validation_context=None, **kwargs):
             context_data = {"info": validation_context}
 
         self.context = Context(context_data)
-        self.add_transform("groups", lambda spec: spec)
+        self.add_transform(
+            "groups",
+            lambda names: list(
+                set(
+                    chain(
```
czosel, line 69: why do we need chain here?

docs/extending.md
```diff
@@ -258,6 +269,34 @@ conflicts.
 [['my-option'], ...]
 ```
 
+## Dynamic Groups
+
+Caluma allows you to write code that returns groups for use in the `address_groups` and `control_groups` properties of a `Task`.
```
open-dynaMIX, line 274: The groups are used in `addressed_groups` / `controlling_groups`. The code is used in `address_groups` / `control_groups`.

anehx: True, but the transforms can only be used in the task. Maybe a completely different wording would be better, do you have a proposal?

open-dynaMIX:
```suggestion
Caluma allows you to write code to be executed as part of the JEXL evaluation in the `address_groups` and `control_groups` properties of a `Task`.
```

PR #1113, opened 2020-07-21
chore: test flaky

sbor23:


caluma/caluma_workflow/tests/test_work_item.py
```diff
@@ -336,6 +336,7 @@ def test_complete_multiple_instance_task_form_work_item_next(
     snapshot.assert_match(result.data)
 
 
+@pytest.mark.flaky(10, 9)
```
czosel, line 339: Does this mean "at least 9 successes in 10 runs"?

czosel:
Is fixing the actual flaky behavior not feasible?

caluma/caluma_workflow/tests/test_work_item.py
sbor23: Yeah, I just wanted to try out if flaky helps us in any way... But doesn't seem to be the case, we don't want any flakiness.

sbor23:
> Is fixing the actual flaky behavior not feasible?

I don't think the code is flaky itself, because the tested result data seems fine to me.
Rather snapshottest is at fault.
It doesn't save to an `OrderedDict` but only to a dictionary.
And then the equality check doesn't like if the test result has a different ordering, even though it's shouldn't matter.

PR #1114, opened 2020-07-21
chore(test): use syrupy instead of snapshottest

sbor23:


czosel:
What's the advantage of syrupy?

sbor23:
> What's the advantage of syrupy?

It seems to handle dict ordering better!
It supports more pytest-idiomatic syntax so you can write `assert value == snapshot` instead of `snapshot.assert_match(value)`.
It doesn't write to a python file but has its own serialized data format, so `black` and `isort` won't care about it.

Also it's maintained compared to snapshottest which didn't have a release in a year.

sbor23:
We might wanna move the `.ambr` snapshot files to commons license...

winged:
Good to merge from my perspective, but I'd like @open-dynaMIX to have a quick glance as well :)

PR #1115, opened 2020-07-22
fix(test): sort the groups jexl

sbor23:


sbor23:
This should fix the non-deterministic snapshot tests failing.

PR #1116, opened 2020-07-22
refactor: fields list should be list, not tuple

winged:
The DRF documentation shows the fields list to be a list, not a
tuple. When analyzing the code using MyPy, it assumes that an N-tuple
is sort of a type, and changing the type to an N+X-tuple in a subclass
is therefore a type validation (which I sort of agree with).

Therefore, let's use lists instead.

### DIFF
--- a/caluma/caluma_form/serializers.py
+++ b/caluma/caluma_form/serializers.py
@@ -22,7 +22,7 @@ class QuestionJexlField(serializers.JexlField):
 class SaveFormSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Form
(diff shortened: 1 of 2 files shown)


ganwell:
@winged it seems you have to update snapshottests.

winged:
> @winged it seems you have to update snapshottests.

Flaky tests :( Works fine on my machine. I'll look into it

sbor23:
@winged would you mind rebasing and pushing this again?

winged:
Rebased, and removed a flaky-test-fix, shouldn't be needed anymore

sbor23 (approved):
Nice, thanks!

PR #1117, opened 2020-07-23
chore: upgrade dependencies, conformity to flake8 rules

sbor23:
Drop isort as explicit dependency, it's installed by flake8-isort.


sbor23:
I like to have as few pinned versions as necessary, otherwise we always get the automated PRs from upgrades which are not supported by the other package. But then again maintainers don't always do a good job pinning themselves so we end up with a borked install.

Goes both sides really.

sbor23:
I left the isort since it's also used as a standalone tool.

PR #1127, opened 2020-07-30
feat(workflow): add mutation and api function for canceling work items

### LINKED ISSUE #1119: Mutation for cancelling workitems
We need a mutation `cancelWorkItem`, otherwise we can't trigger that state manually

anehx:
Resolves #1119

### DIFF
--- a/caluma/caluma_workflow/api.py
+++ b/caluma/caluma_workflow/api.py
@@ -109,3 +109,24 @@ def cancel_case(case: models.Case, user: BaseUser) -> models.Case:
     domain_logic.CancelCaseLogic.post_cancel(case, user)
 
     return case
+
+
+def cancel_work_item(work_item: models.WorkItem, user: BaseUser) -> models.WorkItem:
+    """
+    Cancel a work item (just like `CancelWorkItem`).
+
+    >>> cancel_work_item(
+    ...     work_item=models.WorkItem.first(),
+    ...     user=AnonymousUser()
+    ... )
+    <WorkItem: WorkItem object (some-uuid)>
+    """
+    domain_logic.CancelWorkItemLogic.validate_for_cancel(work_item)
+
+    validated_data = domain_logic.CancelWorkItemLogic.pre_cancel({}, user)
+
+    update_model(models.WorkItem.objects.get(pk=work_item.pk), validated_data)
+
+    domain_logic.CancelWorkItemLogic.post_cancel(work_item, user)
+
+    return work_item
--- a/caluma/caluma_workflow/domain_logic.py
+++ b/caluma/caluma_workflow/domain_logic.py
@@ -266,3 +266,30 @@ class CancelCaseLogic:
         )
 
         return case
+
+
+class CancelWorkItemLogic:
+    @staticmethod
+    def validate_for_cancel(work_item):
+        if work_item.status != models.WorkItem.STATUS_READY:
+            raise ValidationError("Only READY work items can be cancelled")
+
+    @staticmethod
+    def pre_cancel(validated_data, user):
+        validated_data["status"] = models.WorkItem.STATUS_CANCELED
+        validated_data["closed_at"] = timezone.now()
+        validated_data["closed_by_user"] = user.username
+        validated_data["closed_by_group"] = user.group
+
+        return validated_data
+
+    @staticmethod
+    def post_cancel(work_item, user):
+        send_event(
+            events.cancelled_work_item,
+            sender="post_cancel_work_item",
+            work_item=work_item,
+            user=user,
+        )
+
+        return work_item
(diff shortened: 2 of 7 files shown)


caluma/caluma_workflow/serializers.py
```diff
@@ -297,6 +297,34 @@ class Meta:
...
 
 
+class CancelWorkItemSerializer(serializers.ModelSerializer):
+    id = serializers.GlobalIDField()
+
+    def validate(self, data):
+        try:
+            domain_logic.CancelWorkItemLogic.validate_for_cancel(self.instance)
+        except ValidationError as e:
+            raise exceptions.ValidationError(str(e))
```
open-dynaMIX, line 307: No need to catch the `ValidationError` and then raising the same Exception.

caluma/caluma_workflow/domain_logic.py
```diff
@@ -266,3 +266,32 @@ def post_cancel(case, user):
...
+            raise ValidationError("Only READY work items can be cancelled")
+
+    @staticmethod
+    def pre_cancel(validated_data, user):
+        validated_data["status"] = models.WorkItem.STATUS_CANCELED
+        validated_data["closed_at"] = timezone.now()
+        validated_data["closed_by_user"] = user.username
+        validated_data["closed_by_group"] = user.group
+
+        return validated_data
```
open-dynaMIX, line 284: This could be put into the `validate_for_cancel()` method. `pre_cancel()` should be used for logic that actually changes data or needs to run somewhere in the middle of the `update()`-method of the serializer.

caluma/caluma_workflow/domain_logic.py
```diff
@@ -266,3 +266,32 @@ def post_cancel(case, user):
...
+
+    @staticmethod
+    def post_cancel(work_item, user):
+        # TODO: reopen previous work item(s)?
```
open-dynaMIX, line 288: We should do this, right? Otherwise the `Case` would be stalled.

caluma/caluma_workflow/api.py
```diff
@@ -100,3 +100,24 @@ def cancel_case(case: models.Case, user: BaseUser) -> models.Case:
...
+    ...     work_item=models.WorkItem.first(),
+    ...     user=AnonymousUser()
+    ... )
+    <WorkItem: WorkItem object (some-uuid)>
+    """
+    domain_logic.CancelWorkItemLogic.validate_for_cancel(work_item)
+
+    validated_data = domain_logic.CancelWorkItemLogic.pre_cancel({}, user)
+
+    models.WorkItem.objects.filter(pk=work_item.pk).update(**validated_data)
```
open-dynaMIX, line 119: As described [here](https://django-simple-history.readthedocs.io/en/latest/common_issues.html#queryset-updates-with-history-updated-in-django-2-2), this doesn't work with `django-simple-history`. So you should iterate over the queryset.

caluma/caluma_workflow/domain_logic.py
open-dynaMIX: Or is the idea, that the client reopens the previous WorkItem?

caluma/caluma_workflow/serializers.py
anehx: It's not the same.. We catch the django exception and throw the proper exception from DRF.

This is the same in all other mutations that use the domain logic and AFAIR we already discussed this in detail..

caluma/caluma_workflow/domain_logic.py
anehx: This uses exactly the same patterns as the other functions which were already discussed enough. Please check the context of this file and reconsider.

anehx: Also, this obviously changes data :wink:

caluma/caluma_workflow/domain_logic.py
anehx: Right now we have no way to know what the previous work item is. I think for now the client needs to decide what happens to the workflow..

caluma/caluma_workflow/api.py
anehx: :astonished: We use this alot in this file. I'll open a PR to fix the existing usage of this..

caluma/caluma_workflow/serializers.py
open-dynaMIX: Ah, missed that. Nevermind.

caluma/caluma_workflow/domain_logic.py
anehx: As discussed: We'll leave this as it is. The `validate_for` methods are for validation only unlike the validation methods on the serializer which alter data..

caluma/caluma_workflow/api.py
anehx: Fixed!

PR #1128, opened 2020-07-30
fix(history): fix history by not using djangos update function

anehx:
Voilà!

https://github.com/projectcaluma/caluma/pull/1127#discussion_r462837424

### DIFF
--- a/caluma/caluma_workflow/api.py
+++ b/caluma/caluma_workflow/api.py
@@ -1,9 +1,18 @@
 from typing import Optional
 
+from django.db.models import Model
+
 from caluma.caluma_form.models import Form
 from caluma.caluma_user.models import BaseUser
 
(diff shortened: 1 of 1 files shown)


caluma/caluma_workflow/api.py
```diff
@@ -1,11 +1,20 @@
...
 from caluma.caluma_user.models import BaseUser
 
 from . import domain_logic, models
 
 
+def update_model(model: Model, data: dict) -> Model:
+    for key, value in data.items():
+        setattr(model, key, value)
+
+    model.save()
```
open-dynaMIX, line 15: We could also add this as a method to the `BaseModel`:

```python
def update(self, data):
    for key, value in data.items():
        setattr(self, key, value)
    self.save()
```

PR #1129, opened 2020-07-30
Dynamic groups context

sbor23:


caluma/caluma_workflow/tests/test_case.py
```diff
@@ -89,16 +94,18 @@ def test_start_case(
     inp = {"input": {"workflow": workflow.slug, "form": form.slug}}
 
     if mutation == "saveCase":
+        inp["context"] = (json.dumps({"additional_data": "foo"}),)
```
winged, line 97: Why is this a 1-tuple?

sbor23: Ooops that was overlooked.

sbor23:
> I like where this is going! Much better than the previous attempt to directly extend the result that the JEXL evaluation returned.
> 
> However I feel like we should do a few things more, while we still can:
> 
>     * rename `context` to `_context` to give a hint to clients that this field does not, in fact, map to the database. And make this explicit in the docs as well
Not sure if this is a common pattern in GraphQL but seems sane to me.
However my experiment showed that the graphene turns this into a camel-cased `Context` and I'm not sure how to prevent that.
So I'm leaving this as `context`.

>     * Allow passing this on all mutations (and possibly even queries, but this can be added later) - this could also help in cases where we need to pass in data to visibility and permission layers for example, where we used to mess with HTTP headers in the past
I'd like to refrain from that, as long as we don't pass the context around to the relevant places.

>     * Extend the docs with a concrete example
Done.

docs/extending.md
```diff
@@ -289,14 +291,34 @@ from caluma.caluma_workflow.dynamic_groups import (
...
+    @register_dynamic_group("legal_dept")
+    def resolve_legal_dept(self, task, case, user, prev_work_item, _context):
+        # _context might contain additional filters...
+        legal_depts = find_legal_dept(case, coast=_context.get("coast"))
+
+        # ... or plain groups to include
+        return legal_depts + _context.get("additional_groups", [])
+```
+
+Now if the `_context` looks like `{"coast": "west", "additional_groups": ["board", "administration"]}`, the resulting groups are `["legal-dept-san-francisco", "legal-dept-seattle", "board", "administration"]`.
```
winged, line 319: I like the example! However here's some leftover underscore

docs/extending.md
```diff
@@ -289,14 +291,34 @@ from caluma.caluma_workflow.dynamic_groups import (
...
+Use it for passing additional info and implement logic on it, for example run-time dependent `"addressed_groups"` that should be added to newly generated `WorkItems`.
+
+
+We can extend our example above:
+
+```python
+class CustomDynamicGroups(BaseDynamicGroups):
+    @register_dynamic_group("legal_dept")
+    def resolve_legal_dept(self, task, case, user, prev_work_item, _context):
+        # _context might contain additional filters...
```
winged, line 312: I'm not sure this comment wouldn't introduce more confusion, as filters may be confused with the *actual* filters.

```suggestion
        # `context` may contain arbitrary data from the client to affect behaviour here
```

sbor23: Thanks for the suggestion!

docs/extending.md
sbor23: Done

PR #1130, opened 2020-07-30
feat(workflow): save the previous work item on the model

anehx:
Making this information persistent enables us to reconstruct the exact
course of the workflow. This might be helpful in the future if we want
to implement a rollback or similar functionality.


anehx:
@winged Fair enough, I updated the commit description..

PR #1131, opened 2020-07-31
fix(tests): fix flakiness by sorting the list during snapshotting

sbor23:
Fixes flakiness like found here:
https://travis-ci.com/github/projectcaluma/caluma/jobs/366493032

### DIFF
--- a/caluma/caluma_workflow/tests/test_workflow.py
+++ b/caluma/caluma_workflow/tests/test_workflow.py
@@ -1,7 +1,8 @@
 import pytest
 from graphql_relay import to_global_id
 
-from ...caluma_core.tests import extract_serializer_input_fields
+from caluma.caluma_core.tests import extract_serializer_input_fields
+
 from .. import models, serializers
 
 
@@ -113,7 +114,7 @@ def test_add_workflow_flow(
     workflow,
     workflow_start_tasks_factory,
     task,
-    snapshot,
+    sorted_snapshot,
     success,
     next,
     admin_schema_executor,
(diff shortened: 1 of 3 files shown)


caluma/conftest.py
```diff
@@ -282,3 +282,12 @@ def factory(use_table=False, use_subform=False):
         return (form, document, questions, answers)
 
     return factory
+
+
+def sorted_snapshot(name, key):
```
winged, line 287: Nice! But couldn't you make this into a "proper" fixture that uses the `snapshot`, so we won't have to call it with an explicit matcher param? Could be interesting. Also this seems rather specific (handles `tasks` as a special case) so probably shouldn't be in the top-level conftest

sbor23: Ahh returning a `snapshot` instance is even nicer!

But I made it reusable because this problem probably shows up in other places as well.

sbor23: Made it a fixture that accepts a name (eg. `"tasks"`) and a `key` passed to the `sorted` builtin function (eg. `lambda x: x["slug"])

caluma/caluma_workflow/tests/test_workflow.py
```diff
@@ -157,7 +158,7 @@ def test_add_workflow_flow(
     )
     assert not bool(result.errors) == success
     if success:
-        snapshot.assert_match(result.data)
+        result.data == sorted_snapshot("tasks", lambda el: el["slug"])
```
winged, line 161: missing `assert`

sbor23: Done

PR #1133, opened 2020-08-03
feat(workflow): add support for dynamic tasks

anehx:
Please start your review by reading the documentation part :)


caluma/caluma_workflow/domain_logic.py
```diff
@@ -127,7 +126,13 @@ def post_complete(work_item, user, context=None):
...
         ).first()
 
-        if flow:
+        next_tasks = models.Task.objects.filter(
+            pk__in=utils.get_jexl_tasks(
+                flow.next if flow else None, work_item.case, user, work_item, context
+            )
+        )
+
+        if next_tasks.count():
```
open-dynaMIX, line 135:
```suggestion
        if next_tasks.exists():
```

caluma/caluma_workflow/jexl.py
```diff
@@ -17,11 +37,21 @@ def visit_Transform(self, transform):
...
+
+        if transform.name == "tasks":
+            if not isinstance(value, list):
+                yield f"`{value}` is not a valid list of tasks."
+            else:
+                for slug in value:
+                    if not task_exists(slug):
+                        yield f"The task `{slug}` does not exist or is not registered as dynamic task"
+
+        if transform.name == "task":
```
open-dynaMIX, line 50:
```suggestion
        elif transform.name == "task":
```

caluma/caluma_workflow/dynamic_tasks.py
```diff
@@ -0,0 +1,34 @@
...
+        @wraps(fn)
+        def wrapped(self, *args, **kwargs):
+            return fn(self, *args, **kwargs)
+
+        return wrapped
+
+    return wrapper
+
+
+class BaseDynamicTasks:
```
open-dynaMIX, line 19: Please add a docstring with an example, as we have it for the other extension base classes.

caluma/caluma_workflow/tests/test_work_item.py
```diff
@@ -499,7 +499,7 @@ def test_complete_work_item_with_next_multiple_instance_task(
         is_multiple_instance=True, address_groups=["group1", "group2", "group3"]
     )
     task_flow = task_flow_factory(task=task, workflow=workflow)
-    task_flow.flow.next = f"['{task_next.slug}']|task"
+    task_flow.flow.next = f"['{task_next.slug}']|tasks"
```
open-dynaMIX, line 502: Does this indicate a breaking change? Maybe it's too early in the morning, but to my understanding, this should never have worked :thinking:

anehx: This task flow couldn't exist since we validate this on creation. The test was wrong and it shouldn't have worked. However, since it would have failed in the validation, I don't recognize this as breaking..

caluma/caluma_workflow/dynamic_tasks.py
anehx: Fair, will do. I copied this from the dynamic groups where it's also missing so it might be clever to add a comment there as well

caluma/caluma_workflow/tests/test_work_item.py
open-dynaMIX: Thanks for clarifying.

caluma/caluma_workflow/jexl.py
```diff
@@ -17,11 +37,21 @@ def visit_Transform(self, transform):
...
 class TaskValidatingAnalyzer(ValidatingAnalyzer):
     def visit_Transform(self, transform):
-        if transform.name == "tasks" and not isinstance(transform.subject.value, list):
-            yield f"{transform.subject.value} is not a valid list of tasks."
-
-        if transform.name == "task" and not isinstance(transform.subject.value, str):
-            yield f"{transform.subject.value} is not a valid task slug."
+        value = parse_literal(transform.subject.value)
+
+        if transform.name == "tasks":
```
winged, line 42: I think the validation should be moved to the transform methods themselves, not in the generic visit method

caluma/caluma_workflow/jexl.py
```diff
@@ -123,10 +153,23 @@ def _get_dynamic_group(
...
+        self, case=None, user=None, prev_work_item=None, dynamic_context=None, **kwargs
+    ):
         super().__init__(**kwargs)
-        self.add_transform("task", lambda spec: spec)
-        self.add_transform("tasks", lambda spec: spec)
+
+        self.add_transform(
+            "task",
+            lambda slug: self.task_transform(
+                slug, case, user, prev_work_item, dynamic_context
```
winged, line 164: Instead of (ab)using closures here, I'd prefer going the OO path and putting the `__init__()` parameters into instance properties (`self.case = case; self.user = user` etc). Then the `add_transform()` won't need a lambda expression anymore:
```python
self.add_transform("task", self.task_transform)
```
which in my opinion is a lot easier to debug and understand

caluma/caluma_workflow/jexl.py
```diff
@@ -145,3 +188,48 @@ def extract_tasks(self, expr):
...
+        )
+
+    def task_transform(self, task_name, case, user, prev_work_item, dynamic_context):
+        all_tasks = self.tasks_transform(
+            [task_name], case, user, prev_work_item, dynamic_context
+        )
+
+        if len(all_tasks) > 1:
+            raise ValidationError(
+                f"The dynamic task `{task_name}` used by the `task` transform should not return more than one task -- it returned {len(all_tasks)}"
```
winged, line 208: Long line - should be split for readability/pep8:
```suggestion
                f"The dynamic task `{task_name}` used by the `task` transform should "
                f"not return more than one task -- it returned {len(all_tasks)}"
```

docs/extending.md
```diff
@@ -318,6 +317,45 @@ class CustomDynamicGroups(BaseDynamicGroups):
...
+from caluma.caluma_workflow.dynamic_tasks import (
+    BaseDynamicTasks,
+    register_dynamic_task,
+)
+
+
+class CustomDynamicTasks(BaseDynamicTasks):
+    @register_dynamic_task("audit-forms")
+    def resolve_audit_forms(self, case, user, prev_work_item, context):
+        needs_extra_audit = case.document.answers.filter(question_id="is-extra-complicated", value="is-extra-complicated-yes").exists()
```
winged, line 345:
```suggestion
        needs_extra_audit = case.document.answers.filter(
            question_id="is-extra-complicated",
            value="is-extra-complicated-yes"
        ).exists()
```

caluma/caluma_workflow/jexl.py
anehx: Could you give an example on how you'd imagine that? The transform method is for evaluation and should not include validation since we can't evaluate at the time we validate..

docs/extending.md
anehx: Black doesn't format python inside markdown snippets :-1: I'm used to prettier doing my formatting inside markdown files as well :wink:

caluma/caluma_workflow/jexl.py
winged: Yeh you're right - confused analyzer vs evaluation here.

caluma/caluma_workflow/jexl.py
anehx: Fixed

docs/extending.md
anehx: Fixed

caluma/caluma_workflow/jexl.py
anehx: Fixed

caluma/caluma_workflow/domain_logic.py
anehx: Fixed

caluma/caluma_workflow/dynamic_tasks.py
anehx: Done

anehx: I'll add the docstring for the dynamic groups in another PR

caluma/caluma_workflow/jexl.py
anehx: Good input, I changed it.

caluma/caluma_workflow/dynamic_tasks.py
anehx: #1138

caluma/caluma_workflow/jexl.py
anehx: Changed it for the groups transform as well: #1138

Issue #930, opened 2020-02-17
Deprecations


open-dynaMIX:
In this issue we collect deprecations for the next major version bump.

## Deprecations

 * [`userinfo` and `introspection` properties on `OIDCUser`](https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_user/models.py#L20)
 * `slug` field on [`FormFilterSet`](https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/filters.py#L33) and [`QuestionFilterSet`](https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/filters.py#L77)
* [`StartCase` mutation](https://github.com/projectcaluma/caluma/blob/db44614f1e6c20e5fdbb383e14688ef45d04730f/caluma/caluma_workflow/schema.py#L242)
* [Old-style filters](https://github.com/projectcaluma/caluma/blob/master/docs/graphql.md#old-style-filtering-deprecated)

sbor23:
What's up with this, there have been multiple releases in the mean time.
I guess this should be updated regularly for each release but is outdated at the moment, right?

open-dynaMIX:
Only for major version bumps.

PR #1140, opened 2020-08-04
chore(deps): update graphene-django

sbor23:
Update to graphene-django 2.12.1 and fix connection field, schema.

fixes problem found in https://github.com/projectcaluma/caluma/pull/985

### DIFF
--- a/caluma/caluma_core/types.py
+++ b/caluma/caluma_core/types.py
@@ -65,10 +65,13 @@ class DjangoConnectionField(DjangoConnectionField):
     This can be removed, when (or better if)
     https://github.com/graphql-python/graphql-relay-py/issues/12
     is resolved.
+
+    TODO: properly implement max_limit, see
+    https://github.com/graphql-python/graphene-django/blob/b552dcac24364d3ef824f865ba419c74605942b2/graphene_django/fields.py#L133
     """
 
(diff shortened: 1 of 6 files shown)


winged:
LGTM, but note that this is a breaking change due to the schema changing types (`String` to `ID`; `[String]!` to `[String!]!`). Might wanna batch it and do it later when we have some other breaking changes as well (see #930 for example)

sbor23:
> LGTM, but note that this is a breaking change due to the schema changing types (`String` to `ID`; `[String]!` to `[String!]!`). Might wanna batch it and do it later when we have some other breaking changes as well (see #930 for example)

True. However I made the id input variable a `graphene.String()` to stay as close as possible to the existing schema.

winged:
> > LGTM, but note that this is a breaking change due to the schema changing types (`String` to `ID`; `[String]!` to `[String!]!`). Might wanna batch it and do it later when we have some other breaking changes as well (see #930 for example)
> 
> True. However I made the id input variable a `graphene.String()` to stay as close as possible to the existing schema.

It's still gonna break clients if they reference types

sbor23:
Closing in favor of https://github.com/projectcaluma/caluma/pull/1413 which incorporates these fixes.

PR #1146, opened 2020-08-06
fix(workflow): enable context for skipping work items

anehx:
Skipping a work item causes the same side effects for the workflow as
completing it. Therefore it also needs the new context to function
properly.


sbor23:
Oops that one slipped through. Thanks!

PR #1159, opened 2020-08-14
fix(api): fix the return object of api methods

anehx:
The useless refetching of the models before updating them caused a
difference between the value in the DB and the return value since the
return value was not altered or refreshed from the DB.


sbor23:
Haha sorry I noticed this before and actually did a `work_item.refresh_from_db()` on our downstream project.
Forgot to upstream it. Thanks!

PR #1161, opened 2020-08-14
fix(test): import DjangoModelFactory from actual module path

sbor23:
For some reason this was never a problem, but now that we import the module in downstream project it's unable to import `DjangoModelFactory`.

I have several questions...


winged:
I'm wondering why this didn't blow up sooner... The change that caused it is from march: https://github.com/FactoryBoy/factory_boy/commit/a8da29e21b42544fe208b8f641836ca2f2b222c2#diff-dfa6f903347011a16e364051dcae1465

Issue #1156, opened 2020-08-13
RFC: Suspend / resume work items


anehx:
To handle use cases where a work item (_or even a case_) needs to be suspended and resumed at another point we need two new mutations / api methods:

`suspendWorkItem` / `suspend_work_item`

- Takes a work item as parameter and sets the status of said work item to `SUSPENDED`
- Requires the work item to have the status `READY`
- Fires an event `suspended_work_item`

`resumeWorkItem` / `resume_work_item`

- Takes a work item as parameter and sets the status of said work item to `READY`
- Requires the work item to have the status `SUSPENDED`
- Fires an event `resumed_work_item`

Those new methods would follow an official pattern of the standard we try to follow: http://www.workflowpatterns.com/patterns/resource/detour/wrp32.php

![](http://www.workflowpatterns.com/patterns/resource/images/fig2.png)

czosel:
:100: great idea!

The use-case in our project: If the case is "rejected" the rejection can be reverted afterwards. With this feature, we could suspend all open work items when the case is rejected, and

- resume all work items if the rejection is reverted
- cancel all work items when a new case is submitted that replaces the old one

fkm-adfinis:
Sounds good!

winged:
I like this! What I'd like to see is how this affects child cases of the suspended work item - do their work items get suspended as well? Should we mark those cases / work items somehow, or would we do this at query time to see if a parent work item is suspended?

anehx:
> I like this! What I'd like to see is how this affects child cases of the suspended work item - do their work items get suspended as well? Should we mark those cases / work items somehow, or would we do this at query time to see if a parent work item is suspended?

Very good point. This is an issue we should handle in general. Right now, canceling, skipping or completing a work item doesn't affect the child case at all..

anehx:
IMO we need to provide a way to cancel the whole case. If a work item with child case is processed we need to process the child case as well:

- Skipping a work item cancels the child case
- Canceling a work item cancels the child case as well
- Completing a work item only works if the child case is completed (?)

czosel:
> Completing a work item only works if the child case is completed (?)

This is already the case.

open-dynaMIX:
I really like the general idea!

I agree, that child cases need to be suspended too. But there's also question about resuming. I doubt, that it's always desirable to resume all child cases. Especially since we have no way to determine if a suspended child case was suspended on it's own, or as a part of suspending the parent case.

[...]

PR #1180, opened 2020-08-27
feat(workflow): add suspend/resume for work items

sbor23:
Addresses the base implementation of #1156


sbor23:
This is lacking suspend/resume case and recursion

PR #1186, opened 2020-09-01
feat(workflow): suspend and resume cases and work items

### LINKED ISSUE #1156: RFC: Suspend / resume work items

anehx:
This adds functionality (Python API as well as GraphQL API) for
suspending and resuming a work item or even a whole case.

Closes #1156


czosel:
Black complains that it would reformat `0024_auto_20200901_1026.py`.

czosel (approved):
:100:

caluma/caluma_workflow/serializers.py
```diff
@@ -448,3 +510,63 @@ class Meta:
...
+
+    def validate(self, data):
+        try:
+            domain_logic.SuspendWorkItemLogic.validate_for_suspend(self.instance)
+        except ValidationError as e:
+            raise exceptions.ValidationError(str(e))
+
+        return super().validate(data)
+
+    def update(self, work_item, validated_data):
```
winged, line 526: Missing `@transaction.atomic`

caluma/caluma_workflow/serializers.py
```diff
@@ -448,3 +510,63 @@ class Meta:
...
+
+    def validate(self, data):
+        try:
+            domain_logic.ResumeWorkItemLogic.validate_for_resume(self.instance)
+        except ValidationError as e:
+            raise exceptions.ValidationError(str(e))
+
+        return super().validate(data)
+
+    def update(self, work_item, validated_data):
```
winged, line 556: Missing `@transaction.atomic`

caluma/caluma_workflow/models.py
```diff
@@ -167,13 +168,14 @@ class WorkItem(UUIDModel):
...
     STATUS_SKIPPED = "skipped"
+    STATUS_SUSPENDED = "suspended"
 
-    STATUS_CHOICES = (STATUS_READY, STATUS_COMPLETED, STATUS_CANCELED)
     STATUS_CHOICE_TUPLE = (
         (STATUS_READY, "Task is ready to be processed."),
         (STATUS_COMPLETED, "Task is done."),
         (STATUS_CANCELED, "Task is cancelled."),
         (STATUS_SKIPPED, "Task is skipped."),
+        (STATUS_SUSPENDED, "Task is suspended."),
```
winged, line 178: Just noticed due to this change's effect in the migrations: This is on the `WorkItem` model, not `Task`, thus should have work item in the description. Maybe add a drive-by fix commit?

winged (changes requested):
Awesome work, I like it! :tada: 

Just some minor fixes required.

caluma/caluma_workflow/serializers.py
anehx: True! I [fixed the missing transactions](c7963005fbfe1fabeded03ae8304a9cda6bc1239) in the whole file. Since we always trigger an event that could potentially do DB changes it should be used in every `update` or `create`.

caluma/caluma_workflow/serializers.py
anehx: Fixed

caluma/caluma_workflow/models.py
anehx: Absolutely, that bothered me as well. Fixed in bfed4981b73acfbe49bb1ab4644b5742d63058b3

winged (approved):

PR #1187, opened 2020-09-01
chore: deprecate cancel events in favor of the american notation

anehx:
The following events were deprecated in favor of their british notation:

- `cancelled_case` => `canceled_case`
- `cancelled_work_item` => `canceled_work_item`


czosel:
Nitpick: The one-L version is american (PR title is incorrect)

caluma/caluma_workflow/domain_logic.py
```diff
@@ -268,6 +268,8 @@ def pre_cancel(case, validated_data, user, context=None):
 
     @staticmethod
     def post_cancel(case, user, context=None):
+        # TODO: remove in the next major release since `events.cancelled_case`
```
czosel, line 271: Is there a convention for marking deprecations in the source code, something like `# DEPRECATED`? That way we could make it easier to find everything that needs cleaning up when we make a new major release.

docs/events.md
```diff
@@ -60,16 +60,26 @@ in case the event sends additional arguments in the future.
...
 | `completed_case`      | `CompleteWorkItem`                                                | `case`, `user`, `context`      |
-| `cancelled_case`      | `CancelCase`                                                      | `case`, `user`, `context`      |
+| `canceled_case`       | `CancelCase`                                                      | `case`, `user`, `context`      |
 | `suspended_case`      | `SuspendCase`                                                     | `case`, `user`, `context`      |
 | `resumed_case`        | `ResumeCase`                                                      | `case`, `user`, `context`      |
 
+### Deprecated events
+
+Those events will be removed in the next major release since they were
+replaced by their equivalent in the british notation.
```
czosel, line 76:
```suggestion
replaced by their equivalent spelling in american english.
```

caluma/caluma_workflow/domain_logic.py
czosel: cc @winged @open-dynaMIX

open-dynaMIX: We usually emit `DeprecationWarning`s as can be seen [here](https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_user/models.py#L17).

The process is also mentioned [here](https://github.com/projectcaluma/caluma/blob/master/docs/maintainers.md#release-management).

It's not so easy to implement here I guess, as we would need to emit the warning in the event receiver. Maybe it could be done [here](https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_core/events.py#L8).

winged: Or we could try to emit the warning when registering a listener. We already use our own function for this, which currently is an alias for django's event register function.

If done this way, we could issue the warning on startup-time instead of runtime, which in my opinion would be cleaner. However it could still be the case that users go directly through django's `register()` instead, but currently that risk is rather low as we're the only users I'm aware of

open-dynaMIX: Great idea! The documented way of using caluma events is by using our `on()` decorator (currently just aliased), so we should be save even if we'd not be the only users. [This section](https://github.com/projectcaluma/caluma/blob/08879ce497ceef3b25ba54587360d5e8921e8d7a/docs/events.md#on) of the docs should be updated, if we do this.

anehx: Ok done

winged: Nice work, exactly what I had in mind!

docs/events.md
```diff
@@ -37,8 +37,8 @@ def send_mail_on_complete_work_item_2(sender, work_item, user, context, **kwargs
 
 ## @on
 
-Right now `@on` is just an alias for `django.dispatch.receiver`. This very likely will change
-in the future, but the public API should stay the same.
+`@on` is just a wrapper for `django.dispatch.receiver` that handles
+deprecations of Calumas custom events as well.
```
open-dynaMIX, line 41: I'd consider not stating the exact differences between `@on` and `@receiver`.

czosel: Fixed

czosel (approved):

docs/events.md
anehx: Done

caluma/caluma_workflow/migrations/0026_auto_20200901_1435.py
```diff
@@ -0,0 +1,40 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="case",
+            name="status",
+            field=caluma.caluma_core.models.ChoicesCharField(
+                choices=[
+                    ("running", "Case is running and work items need to be completed."),
+                    ("completed", "Case is done."),
+                    ("canceled", "Case is canceled."),
```
sbor23, line 19: Am I missing something or should we also provide an migration for moving existing status from "cancelled" to "canceled"?
The current migration just changes the value of the choices.

anehx: The value did not change but the description to said value. Check out the changes in the models.

PR #1191, opened 2020-09-04
fix: always send post_complete_work_item event

czosel:
This event should always be sent when a work item has been completed,
even if the workflow doesn't continue (e.g. for multiple instance tasks
that remain open).


anehx (changes requested):
This will break the logic of the events, since the `created_work_item` events will be fired after the `complete_work_item` event..

czosel:
As discussed: This order is actually more logical, and it doesn't break our current implementations. I added some documentation that specifies the order of emitted events.

anehx (approved):

PR #1195, opened 2020-09-08
fix: don't send complete_work_item event when skip

czosel:
Based on #1191

### DIFF
--- a/caluma/caluma_workflow/domain_logic.py
+++ b/caluma/caluma_workflow/domain_logic.py
@@ -122,15 +122,17 @@ class CompleteWorkItemLogic:
         return validated_data
 
     @staticmethod
-    def post_complete(work_item, user, context=None):
+    def post_complete(
+        work_item,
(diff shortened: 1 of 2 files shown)


czosel:
@anehx ready for review

anehx (approved):

caluma/caluma_workflow/domain_logic.py
```diff
@@ -122,15 +122,17 @@ def pre_complete(work_item, validated_data, user, context=None):
         return validated_data
 
     @staticmethod
-    def post_complete(work_item, user, context=None):
+    def post_complete(
+        work_item,
+        user,
+        context=None,
+        event=events.completed_work_item,
```
sbor23, line 129: Is there a specific reason to make the event an argument? IMO this should be always the same event

sbor23: Nevermind

PR #1207, opened 2020-09-17
feat(logging): create caluma_logging app for access logging

sbor23:


caluma/settings/django.py
```diff
@@ -156,3 +156,11 @@ def parse_admins(admins):  # pragma: no cover
 # GraphQL
 if DEBUG:
     GRAPHENE["MIDDLEWARE"].append("graphene_django.debug.DjangoDebugMiddleware")
+
+
+# caluma access log
+ENABLE_ACCESS_LOG = env.bool("ENABLE_ACCESS_LOG", default=False)
+
+if ENABLE_ACCESS_LOG:
```
winged, line 164: I'd trigger only the logging, not enabling / disabling the whole app. Otherwise you might run into problems if after the config change `migrate` is not being run

caluma/caluma_logging/middleware.py
```diff
@@ -0,0 +1,22 @@
...
+class AccessLogMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        # user is available only after the request was processed
+        body = json.loads(request.body.decode("utf-8"))
+
+        AccessLog.objects.create(
```
winged, line 15: Maybe we should try to parse the request just a bit further: Figure out if it's a mutation or not, so we can easily search for the type of request. Might even go as far and store the query / mutation name as well if it's not too much work

winged (changes requested):
Nice work so far, but apart from the noted issues, I see a lack of tests, and some missing documentation

caluma/caluma_logging/middleware.py
```diff
@@ -0,0 +1,22 @@
...
+            username=request.user.username,
+            query=body.get("query"),
+            variables=body.get("variables"),
+            operation_name=body.get("operationName"),
```
open-dynaMIX, line 19: It'd make sense to also write the status code of the response into the DB, so we can distinguish between successful and unsuccessful requests.

open-dynaMIX (changes requested):
Nice work!

caluma/caluma_logging/middleware.py
sbor23: I thought we always return 200, even if there are errors...?!

caluma/caluma_logging/middleware.py
sbor23: I agree this could be usefull information but then I would rather move this somewhere deeper after the graphql parsing is done, not doing it ourselfs again...
Maybe the graphene MIDDLEWARE setting?

caluma/settings/django.py
sbor23: That's true, having the app in INSTALLED_APPS shouldn't hurt anyways.

caluma/caluma_logging/middleware.py
open-dynaMIX: We also return 401. And yes, for the 200s we still don't know if they produced some error (could be tracked too). But at least we could filter the 401s.

caluma/caluma_logging/tests/test_logging_middleware.py
```diff
@@ -0,0 +1,115 @@
...
+
+from .. import models
+
+
+@pytest.mark.parametrize(
+    "query,variables,operation,operation_name,selection,status_code",
+    [
+        (
+            """
+            query foo {
```
winged, line 12: I'd probably factor out the query literals here so the parametrisation list becomes a bit more readable. Put it in a couple module-level variables for example.

caluma/caluma_logging/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+        response = self.get_response(request)
+        # user is available only after the request was processed
+        try:
+            body = json.loads(request.body.decode("utf-8"))
+        except json.decoder.JSONDecodeError:
+            return response
+
+        vis = AccessLogVisitor()
+        try:
+            doc = parser.parse(body["query"])
```
winged, line 23: Aaah I like this one! Except for the fact that now the whole request parsing will happen twice. But we can leave that optimisation for another time, I don't think it's worth it to dig too deep just now

winged (changes requested):
Nice work! I like that we're parsing the request "properly" now, I feel this could be much more useful.

I'd still add some documentation so people know that it's happening.

Also: Are there plans to expose the log somehow? As a readonly endpoint for example?

caluma/caluma_logging/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+            variables=body.get("variables"),
+            status_code=response.status_code,
+            **vis.values,
+        )
+
+        return response
+
+
+class AccessLogVisitor(visitor.Visitor):
+    values = {}
```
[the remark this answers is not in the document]
sbor23: Ooof that hurts a little and was the source of failing tests. Thanks!

caluma/caluma_logging/middleware.py
sbor23: As seen above, we just do the parsing again. Not optimal from a performance standpoint but shouldn't be too bad either.

caluma/caluma_logging/tests/test_logging_middleware.py
sbor23: I agree it's not pretty but supporting different levels of nesting etc with parametrization would be too complicated to be useful.

caluma/caluma_logging/middleware.py
sbor23: Included that status code

caluma/caluma_logging/tests/test_logging_middleware.py
winged: No i mean just put the content of the query string into a variable, so the parametrization list looks a bit cleaner. structure-wise it'll still be the same

winged (changes requested):
Still needs a few words of documentation IMHO. Otherwise looks good now

winged (approved):

open-dynaMIX (approved):

PR #1215, opened 2020-09-25
fix(migrations): do not hang in migrations on broken data

winged:
When a case's parent work item points to itself, the migration will
loop forever. Catch this situation and abort in that case.

### DIFF
--- a/caluma/caluma_workflow/migrations/0020_case_families.py
+++ b/caluma/caluma_workflow/migrations/0020_case_families.py
@@ -16,8 +16,13 @@ def set_family(apps, schema_editor):
 
     for case in orphan_cases:
         family = case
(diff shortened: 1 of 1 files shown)


winged:
There was another bug in the migration as well in that it only ever worked on flat hierarchies. Deeper nesting wouldn't work either, so fixed that as well.

czosel (approved):
👍

open-dynaMIX (approved):
Awesome!

CI hangs again...

winged:
> Awesome!
> 
> CI hangs again...

Let's try to trigger it again via whitespace change :-)

open-dynaMIX:
> > Awesome!
> > CI hangs again...
> 
> Let's try to trigger it again via whitespace change :-)

PR #1220, opened 2020-09-29
fix: migrate db sequences with prefix "caluma_"

### LINKED ISSUE #1108: Wrong sequence names
We once migrated the `caluma_` prefix into all table names (928146577f3cb2f76a4e715ba29ae4e1eb547907) however we didn't consider the sequence names. Maybe we should migrate the sequence names as well to be consistent?

sbor23:
Resolves #1108

/cc @anehx

### DIFF
--- a/caluma/caluma_core/management/commands/migrate_to_prefixed_apps.py
+++ b/caluma/caluma_core/management/commands/migrate_to_prefixed_apps.py
@@ -49,6 +49,12 @@ class Command(BaseCommand):
             cursor.execute(query)
             return [row[0] for row in cursor.fetchall()]
 
+    def _get_sequence_list(self):
+        query = f"""SELECT sequence_name FROM information_schema.sequences WHERE sequence_name LIKE '{self._act_prefix}form_%' OR sequence_name LIKE '{self._act_prefix}workflow_%';"""
+        with connection.cursor() as cursor:
+            cursor.execute(query)
+            return [row[0] for row in cursor.fetchall()]
+
     def _collect_content_type_queries(self):
         queries = [
             f"""UPDATE django_content_type SET app_label = CONCAT('{self.prefix}', app_label) WHERE app_label = 'form' OR app_label = 'workflow';"""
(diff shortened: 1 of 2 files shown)


winged (changes requested):
Looks good! Will this require us to re-run that migration command to benefit from the fix? If so, this may be something for the docs as well, or at least for the release notes.

Also, generally I'd like to see formatted SQL for readability.  For example:

```sql
SELECT sequence_name
FROM information_schema.sequences
WHERE
    sequence_name LIKE '{self._act_prefix}form_%' OR
    sequence_name LIKE '{self._act_prefix}workflow_%';
```

sbor23:
> Looks good! Will this require us to re-run that migration command to benefit from the fix? If so, this may be something for the docs as well, or at least for the release notes.

AFAIK this is integrated with django's `migrate` command and will be run always.
@open-dynaMIX correct me please.

> Also, generally I'd like to see formatted SQL for readability. For example:
> 
> ```sql
> SELECT sequence_name
> FROM information_schema.sequences
> WHERE
>     sequence_name LIKE '{self._act_prefix}form_%' OR
>     sequence_name LIKE '{self._act_prefix}workflow_%';
> ```

Fair enough but I won't change it here because the other snippets are the same. Also newlines in multi-line strings are kinda ugly when getting the indentation correct.

winged:
> Fair enough but I won't change it here because the other snippets are the same

OK for me

> Also newlines in multi-line strings are kinda ugly when getting the indentation correct.

Not really. In those cases, just indent so it aligns with the surrounding python code. SQL doesn't care about indentation or lots of whitespace. The only time this gets ugly is if you have an inline string where whitespace matters (for example user-visible output)

winged (approved):

sbor23:
No idea why travis fails there, I can't reproduce the failing test locally whatever I'm trying...

open-dynaMIX:
> AFAIK this is integrated with django's `migrate` command and will be run always.
> @open-dynaMIX correct me please.

Correct.

winged:
I think the wrapper from graphene has a bug. Your code calls `execute(sql)` without parameters, which in Django defaults to `None`, as seen in this fragment (found via call stack of the error):
```python
    # from ...site-packages/django/db/backends/utils.py
    def execute(self, sql, params=None):
        return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
```

However in the wrapper (as seen in the call stack) defaults `params=()`, which is `not None` when actually executed:
```python
    # from ...site-packages/graphene_django/debug_sql/tracking.py:
    def execute(self, sql, params=()):
        return self._record(self.cursor.execute, sql, params)
```

To workaround this, try explicitly passing `params=None` in the `execute()` call

PR #1229, opened 2020-10-06
fix: do not crash during migrations

winged:
Some data was even worse so the fix from the previous PR[1] didn't
catch 100% of the cases. This fix addresses the issue and is able
to process the broken data as expected.

[1]:https://github.com/projectcaluma/caluma/pull/1215

### DIFF
--- a/caluma/caluma_workflow/migrations/0020_case_families.py
+++ b/caluma/caluma_workflow/migrations/0020_case_families.py
@@ -16,13 +16,23 @@ def set_family(apps, schema_editor):
 
     for case in orphan_cases:
         family = case
+
         # Navigate up the hierarchy, but avoid hanging
         # if we have a loop in the data structure
-        while (
-            hasattr(family, "parent_work_item")
-            and family.parent_work_item.case != family
-        ):
-            family = family.parent_work_item.case
+        try:
+
+            while (
+                hasattr(family, "parent_work_item")
+                and family.parent_work_item.case != family
+            ):
+                family = family.parent_work_item.case
+        except models.DoesNotExist:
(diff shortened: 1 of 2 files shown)


caluma/caluma_workflow/migrations/0020_case_families.py
```diff
@@ -16,13 +16,22 @@ def set_family(apps, schema_editor):
...
-            and family.parent_work_item.case != family
-        ):
-            family = family.parent_work_item.case
+        try:
+            while (
+                hasattr(family, "parent_work_item")
+                and family.parent_work_item.case != family
+            ):
+                family = family.parent_work_item.case
+        except Exception:  # pragma: no cover
```
open-dynaMIX, line 28: Is the Exception raised inconsistent? If not, I'd prefer catchinf the specific exception.

I'd like to have more information about that. `hasattr()` succeeds, but accessing the attribute still fails? Could it be that it fails when accessing `case` and not `parent_work_item`?

winged: I knew this would come up! The exception is an implicitly generated one. Due to non-repeatability, I didn't catch it's MRO to see which one makes most sense to catch.

Yes could be that it's due to the parent work items' case being inaccessible, but Don't exactly remember.

Either way, I think it'd be better to generate a test case for this anyway, with as much broken data as possible :)

winged: Was able to construct a test case, so we now have the correct exception to catch, and a `pragma no cover` less

open-dynaMIX (approved):

PR #1247, opened 2020-10-28
Initial python api for caluma_form

sbor23:
* [x] `saveDocument___Answer` -> `save_answer` handles all question types
* [x] `saveDocument`

Future considerations:
* [ ] `removeAnswer`
* [ ] `copyDocument`
* [ ] `save___Question`
* [ ] `saveForm`
* [ ] `copyForm`
* [ ] `addFormQuestion`
* [ ] `removeFormQuestion`
* [ ] `reorderFormQuestions`

### DIFF
--- a/caluma/caluma_form/api.py
+++ b/caluma/caluma_form/api.py
@@ -1,0 +1,40 @@
+from typing import Any, Optional
+
+from caluma.caluma_form import domain_logic, models
+from caluma.caluma_user.models import BaseUser
+
+
+def save_answer(
+    question: models.Question,
+    document: models.Document,
+    user: Optional[BaseUser] = None,
+    value: Optional[Any] = None,
+    context: Optional[dict] = None,
+    **kwargs
+) -> models.Answer:
+    """
+    Save an answer for given question, document.
+
+    Similar to saveDocumentStringAnswer and the likes, it performes upsert.
+    :param value: Must match the question type
+    """
+
+    data = {"question": question, "document": document, "value": value}
+    data.update(kwargs)
+
+    answer = models.Answer.objects.filter(question=question, document=document).first()
+
+    validated_data = domain_logic.SaveAnswerLogic.pre_save(
+        domain_logic.SaveAnswerLogic.validate_for_save(
+            data, user, answer=answer, origin=True
+        )
+    )
+
+    if answer is None:
+        answer = domain_logic.SaveAnswerLogic.create(validated_data, user)
+    else:
+        answer = domain_logic.SaveAnswerLogic.update(validated_data, answer)
+
+    domain_logic.SaveAnswerLogic.post_save(answer)
+
+    return answer
--- a/caluma/caluma_form/api.py
+++ b/caluma/caluma_form/api.py
@@ -38,3 +38,21 @@ def save_answer(
     domain_logic.SaveAnswerLogic.post_save(answer)
 
     return answer
+
+
+def save_document(
+    form: models.Form,
+    meta: Optional[dict] = None,
+    document: Optional[models.Document] = None,
+    user: Optional[BaseUser] = None,
+) -> models.Document:
+    """Save a document for a given form."""
+
+    if meta is None:
+        meta = {}
+
+    if not document:
+        return domain_logic.SaveDocumentLogic.create(form=form, meta=meta, user=user)
+
+    domain_logic.SaveDocumentLogic.update(document, form=form, meta=meta)
+    return document
--- a/caluma/caluma_form/domain_logic.py
+++ b/caluma/caluma_form/domain_logic.py
@@ -1,8 +1,22 @@
+from typing import Optional
+
 from django.db import transaction
 
+from caluma.caluma_core.models import BaseModel
 from caluma.caluma_form import models, validators
 from caluma.caluma_user.models import BaseUser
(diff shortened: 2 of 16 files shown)


caluma/caluma_form/api.py
```diff
@@ -0,0 +1,48 @@
+from typing import Any, Optional
+
+from caluma.caluma_form import domain_logic, models
+from caluma.caluma_user.models import BaseUser
+
+
+def save_answer(
+    question: models.Question,
+    document: models.Document,
+    user: BaseUser,
```
winged, line 10: I suggest typing `user: Optional[User]` instead, which would make usage from scripts a bit easier

winged (changes requested):
Nice work! Only some minor nitpick.

However due to the change's impact, I suggest @open-dynaMIX  have a look at it as well.

caluma/caluma_form/api.py
sbor23: What would be the default here, if no `user` is provided? `AnonymousUser`?
The goal would be that `created_by_user` and `created_by_group` are populated too, so we need a default.

winged: The default would be `None` for both fields, which in case of script-generated data would be the right thing anyway (and it's also why the `created_by_` fields in the model are explicitly nullable

sbor23: Ah I see, valid point :)

winged (approved):

open-dynaMIX (approved):
Awesome work!

caluma/caluma_form/tests/test_document.py
```diff
@@ -740,11 +743,22 @@ def test_save_document_answer(
...
+        if success:
+            snapshot.assert_match(result.data)
+    else:
+        try:
+            answer = api.save_answer(
+                question, answer.document, user=admin_user, value=inp["input"]["value"]
+            )
+            snapshot.assert_match(answer)
+        except Exception:
+            if success:
```
anehx, line 760: I'd prefer using `pytest.raises(TheExpectedException)`

caluma/caluma_workflow/tests/test_case.py
```diff
@@ -674,7 +674,7 @@ def test_cancel_suspend_resume_case(
     expected_work_item_status,
     error,
 ):
-    if use_api:
+    if not use_api:
```
anehx, line 677: `use_api` means use the graphql api, maybe we should reconsider the naming of that param..

caluma/caluma_form/tests/test_document.py
```diff
@@ -385,6 +385,7 @@ def test_save_document(db, document, schema_executor, update):
     assert same_id == update
 
 
+@pytest.mark.parametrize("use_api", [True, False])  # noqa:C901
```
anehx, line 388: The naming is not clear enough (also stated in an existing test). We should name it `use_graphql` or something..

caluma/caluma_form/serializers.py
```diff
@@ -625,6 +585,7 @@ class Meta:
 class RemoveDocumentSerializer(serializers.ModelSerializer):
     document = serializers.GlobalIDField(source="id")
 
+    @transaction.atomic
     def update(self, instance, validated_data):
         if hasattr(instance, "case"):
             raise Exception("You cannot remove a Document, if it's attached to a case.")
```
anehx, line 591: Can you remove a document if it's family is attached to a case?

caluma/caluma_form/domain_logic.py
```diff
@@ -0,0 +1,78 @@
...
+            validated_data["file"] = validated_data.pop("value")
+            validated_data = __class__._set_file(validated_data)
+
+        update_model(answer, validated_data)
+
+        if answer.question.type == models.Question.TYPE_TABLE:
+            answer.create_answer_documents(row_docs)
+
+    @staticmethod
+    def _set_file(validated_data):
```
anehx, line 74: This is a "private" method but is used in various other files, doesn't really make sense

anehx (changes requested):
Awesome work, but docs are missing :wink:

caluma/caluma_workflow/tests/test_case.py
sbor23: True, will do!

caluma/caluma_form/tests/test_document.py
sbor23: Hmm it's not really that we want to check *which* exception occurs, but rather *if* it occurs when we expect success. But I agree it's not so readable.

caluma/caluma_form/tests/test_document.py
sbor23: Thinking about it, we should use `use_python_api` or something. The vast majority of tests use graphql so we want to denote the exception

caluma/caluma_form/serializers.py
sbor23: I don't understand your question. I've only added the atomic transaction for consistency, didn't touch the mutation otherwise.

anehx: The question wasn't related to the PR it just popped up when I read it. NVM

anehx (approved):

PR #1249, opened 2020-10-28
feat(jexl): debug transforms

### LINKED ISSUE #1248: JEXL: Debug transform
During form configuration, it may sometimes be difficult to decide if an expression is correct or not, or if a value contains the expected data.

For this reason, I propose introducing a `debug` transform that will log the transformed value (`console.log()` in JS, `log.info()` in python), then return the value without modification.

This would allow you to write expressions like `'foo'|answer|debug == 'foo-selection'` and see if the expression left of the transform actually does what you think. It will also allow some exploratory experimenting for people who are not familiar with available variables.

winged:
Introduce debug transforms, which allow users to take a look into
values being processed within JEXL. This is intended for people who are
configuring forms, workflows etc and need to debug their expressions.

This is my proposal to fix #1248. Will need a corresponding implementation in [ember-caluma](https://github.com/projectcaluma/ember-caluma) as well.

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -46,6 +46,11 @@ Here are the available transforms:
   things currently in the fridge, and you want help the user to decide what to cook:
   `'ravioli' in 'fridge-contents'|answer|mapby('food-name')` will tell you whether
   there are ravioli.
+* `debug`: Does not modify the value, but writes the value to the log. This is
+  especially useful when you are exploring the data while building forms or
+  workflows. The log message may appear in different places depending on
+  where it's being run (Browser: console log in the debug tools, Server:
+  System or container logs, might depend on your logging configuration)
 
 ### Operators
 
--- a/caluma/caluma_core/jexl.py
+++ b/caluma/caluma_core/jexl.py
@@ -1,9 +1,12 @@
 from itertools import count
+from logging import getLogger
 
 import pyjexl
 from pyjexl.analysis import ValidatingAnalyzer
 from pyjexl.exceptions import ParseError
 from rest_framework import exceptions
+
+log = getLogger(__name__)
 
 
 class Cache:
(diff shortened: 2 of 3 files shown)


caluma/caluma_core/serializers.py
```diff
@@ -71,7 +71,6 @@ def create(self, validated_data):
         user = self.context["request"].user
         validated_data["created_by_user"] = user.username
         validated_data["created_by_group"] = user.group
-
```
open-dynaMIX, line 74: Was this line removed by black?

open-dynaMIX (approved):
Awesome! Just a strange empty line change.

caluma/caluma_core/serializers.py
winged: Hm, not sure, but no need to touch that file. fixed :)

docs/validation.md
```diff
@@ -46,6 +46,11 @@ Here are the available transforms:
   things currently in the fridge, and you want help the user to decide what to cook:
   `'ravioli' in 'fridge-contents'|answer|mapby('food-name')` will tell you whether
   there are ravioli.
+* `debug`: Does not modify the value, but writes the value to the log. This is
+  especially useful when you are exploring the data while building forms or
+  workflows. Depending on the environment, the log message may appear in
+  different places (Browser: console log in the debug tools, Server: System or
+  container logs, might depend on further configuration)
```
czosel, line 53: Can we be more precise here, or at least link to someplace explaining "further configuration"?

czosel (approved):
Awesome! Are we tracking the frontend implementation of this somewhere?

docs/validation.md
winged: Clarified a bit, but I don't want to go into a full explanation as to where the log messages may go and what might influence it (depending on django config, whether you run in a container or not, within kubernetes or not, where is the log routed, is it collected or sent to a syslogd, etc)

czosel: :+1:

winged:
Note: Converted to draft, as there might be an issue with the dependency graph story if used incorrectly:

`'foo'|answer|debug` would certainly work, but `foo|debug|answer` would cause issues, as the transform extraction visitor would not recognize the lefthand side as a valid input, and may crash (The frontend does have this problem, so I'm planning on fixing this in the backend too).

winged:
Ready for re-review

PR #1260, opened 2020-11-06
feat(forms): implement default_answers

open-dynaMIX:
Introduce default_answers

This feature is available for following question types:

 - Text
 - TextArea
 - Integer
 - Float
 - Choice
 - MultipleChoice
 - Table


caluma/caluma_form/domain_logic.py
```diff
@@ -98,12 +98,29 @@ def set_file(validated_data):
...
-        **kwargs,
-    ):
-        # return BaseLogic.create(models.Document, form=form, meta=meta, user=user)
-        return BaseLogic.create(models.Document, **kwargs)
+    def create(**kwargs):
+        document = BaseLogic.create(models.Document, **kwargs)
+
+        for question in document.form.questions.filter(
+            default_answer__isnull=False
+        ).iterator():
```
winged, line 106: Hm, What is the `.iterator()` here for?

caluma/caluma_form/models.py
```diff
@@ -273,24 +277,9 @@ def copy(self, family=None):
...
-
-            # TableAnswer: copy AnswerDocument too
-            for answer_doc in AnswerDocument.objects.filter(answer=source_answer):
-                new_doc = answer_doc.document.copy(family=family)
-
-                AnswerDocument.objects.create(
-                    answer=new_answer, document=new_doc, sort=answer_doc.sort
-                )
+            new_answer = source_answer.copy(document_family=family)
+            new_answer.document = new_document
```
winged, line 281: How about adding a `to_document` parameter to the `Answer.copy()` method? This way you'd avoid double-saving the answer

winged (approved):
Awesome work! :tada: Just a minor question / suggestion

caluma/caluma_form/models.py
open-dynaMIX: Great idea! Done.

caluma/caluma_form/domain_logic.py
open-dynaMIX: Please correct my if I'm wrong:

We only need to evaluate those queries once. When using `iterator()` in combination with a postgresql DB, we're disabling caching of the result on QuerySet level and use a server-side cursor to fetch the records one-by-one form the DB. From the [docs](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#iterator):

> For a QuerySet which returns a large number of objects that you only need to access once, this can result in better performance and a significant reduction in memory.

winged: Awesome! Today I learned!

PR #1268, opened 2020-11-16
chore: fix flaky test

open-dynaMIX:


winged (approved):
Looks good!

Long term, we should probably think about implementing some sort of default ordering for all connection fields, so we won't have to normalize data in the tests though.

PR #1275, opened 2020-11-18
chore: separate version pinning for install_requires

### LINKED ISSUE #1271: Separate dependency pinning for requirements.txt and setup.py
As caluma can be used as an app, there should be more loose version requirements in setup.py.
Requiremens.txt should be used for development purposes and for using caluma as a standalone Django application.
Setup.py should make sure that the necessary dependencies are available, but not specify exact versions to avoid conflicts downstream.

https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py#answer-33685899

sbor23:
Explicitly declare install_requires instead of copying requirements.txt.
The setup.py needs looser requirements to ease package installation.
This has been an issue before, when caluma installed as a package, but
with the upcoming pip resolver downstream projects will break.

Add a separate travis build for the package.

Resolves #1271

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,6 @@
 # When adding new dependencies here, also add them to the renovate.json to
 # enable auto-merge
 
--r requirements.txt
 black==19.3b0
 factory-boy==3.0.1
 faker==4.16.0
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -21,6 +21,7 @@ Files:
     manage.py
     MANIFEST.in
     renovate.json
+    requirements-all.txt
     requirements-dev.txt
     requirements.txt
     setup.cfg
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,6 @@ env:
     - DOCKER_COMPOSE_VERSION=1.24.0
 
 before_install:
-  # install newer compose version
-  - sudo rm /usr/local/bin/docker-compose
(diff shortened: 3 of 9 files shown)


sbor23:
Maybe we should add a test step to the package publishing github action too...

requirements-dev.txt
```diff
@@ -1,31 +1,2 @@
...
-pytest-env==0.6.2
-pytest-factoryboy==2.0.3
-pytest-mock==3.3.1
-pytest-randomly==3.5.0
-pytest-xdist==2.1.0
-python-semantic-release==7.3.0
-requests-mock==1.8.0
-reuse==0.11.1
-syrupy==0.8.4
+-r requirements-devel.txt
```
winged, line 2: Why this indirection? I find it a bit confusing to have -dev and -devel here

sbor23: The indirection is needed, so that the package build step doesn't install the exact requirements from "requirements.txt" too, that would make the build step useless.
The reasoning behind this is, that the package should install and work with the loosened "install_requires" after all!

I agree the naming is not optimal, actually I would prefer something like "requirements-all.txt" what used to be "requirements-dev.txt" until now (since it loads all the requirements) and "requirements-dev.txt" holding just the dev dependencies.
However that would break our workflow / muscle memory...

winged: Hmm, I do like the `-all` suffix, even if it does break muscle memory. Makes it more obvious what's happening, than trying to remember whether to use `-dev` or `-devel` every time.

sbor23: fine with me!

winged (approved):
Looks good!

PR #1277, opened 2020-11-19
fix(package): move deps to setup.py

sbor23:
Splitting the config between setup.py and setup.cfg didn't work out as planned, the resulting package on pypi had no caluma source.
So move to a complete setup.py config instead


ganwell (approved):
:metal:

PR #1294, opened 2020-11-27
feat: pass user to copy method on answer and doc

open-dynaMIX:


winged (approved):
Looks good!

I wonder whether we'll need similar "acting as group" logic as we do in Alexandria (See projectcaluma/alexandria/pull/115) in the future, to avoid ambiguities

open-dynaMIX:
> I wonder whether we'll need similar "acting as group" logic as we do in Alexandria (See projectcaluma/alexandria/pull/115) in the future, to avoid ambiguities

The concept of "acting as group" in Caluma is to override the `group` property of the user in the validation layer. This can be done with the help of a header for example. But this is not so obvious, so maybe it would make sense the have something like you propose.

PR #1295, opened 2020-11-27
fix(form): fix empty answers

anehx:
This fixes a regression introduced in 6b71559eb7a1dca280de625c89d7b4e65c02decc


anehx:
I wonder why the tests weren't failing before..

anehx:
> I wonder why the tests weren't failing before..

The test tested emptying an answer that already had the value `None` :joy: Fixed the test so it was failing before the change in this PR.

open-dynaMIX (approved):
Great catch! :tada:

Just needs a rebase.

PR #1297, opened 2020-11-27
chore: disable gitlint on master

open-dynaMIX:


winged (approved):
Looks good! [Ref regarding the fix](https://github.com/actions/checkout) says:
> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer here to learn which commit $GITHUB_SHA points to for different events.

PR #1303, opened 2020-12-09
fix(validation): fix custom validation for save answer mutations

anehx:
Fixes a regression introduced in 6b71559eb7a1dca280de625c89d7b4e65c02decc

### DIFF
--- a/caluma/caluma_form/serializers.py
+++ b/caluma/caluma_form/serializers.py
@@ -475,9 +475,10 @@ class DocumentSerializer(serializers.ModelSerializer):
 
 class SaveAnswerSerializer(serializers.ModelSerializer):
     def validate(self, data):
-        return domain_logic.SaveAnswerLogic.validate_for_save(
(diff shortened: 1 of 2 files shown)


sbor23:
Which exact case does this fix? Can you please create a test case for it?

We're a bit in a weird spot, because we try to streamline the the graphql and python APIs as much as possible. Now some validation still seems to happen in rest_framework only, which isn't covered in the domain_logic (and thus in the python API).
I would prefer if that case is handled in domain logic instead of calling into rest_framework, so that both APIs correct.

anehx:
This fixes the custom validation classes which can't be used with the python API and yes I can write a test for this. However, I'd like to mention that this is the second fix I'm doing for this exact commit. The quality control has got to improve...

anehx:
Added the test @sbor23

sbor23 (approved):
LGTM, thanks!

PR #1322, opened 2020-12-29
feat(form): calculated question

sbor23:
Support calculated question of type float.
This commit adds all necessary bits for calculated questions with the
implementation of the calculated_float type. Answers for a calculated
question will be calculated from its `calcExpression` jexl.

Example:
slug: "total-price"
calcExpression:
'"quantity"|answer * ("product-price"|answer * (1 + "vat"|answer))'

Caluma will try to generate the answers as soon as possible, with a
default value of `None`. The value will be updated whenever the
expression is valid (eg. all referenced answers are filled).
Recalculation occurs whenver a referenced answer is updated or if the
calcExpression is changed.

Limitations:
Referencing another calculated question is not permitted and will yield
a validation error.


caluma/tests/__snapshots__/test_schema.ambr
```diff
@@ -563,7 +581,7 @@
     slug: String!
     label: String!
     document: Document!
-    question: StaticQuestion!
+    question: CalculatedFloatQuestion!
```
sbor23, line 584: It's a mystery to me what's going on with the question here. Even the previous value `StaticQuestion` is non-sense IMO.
It seems like this is tied to the order of declaration in schema.py, where `CalculatedFloatQuestion` is next after `StaticQuestion`.

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+```
+'"quantity"|answer * ("product-price"|answer * (1 + "vat"|answer))'
+```
+
+### Implementation
+
+This feature makes heavy use of Django signals. There are two "main loops" for this.
+
+
+#### `calcDependents`
```
winged, line 20: I would move this fragment to internal documentation. We're not exposing it, and the user won't need to know about it

winged (changes requested):
Awesome work! For such a complex change, I'd like to have at least one more dev look into it however.

Since @open-dynaMIX is on vacation, maybe @czosel or @anehx could try to refresh their python skills? Maybe also have the person doing the frontend implementation have a good look as well.

caluma/caluma_form/jexl.py
```diff
@@ -103,93 +94,96 @@ def use_question_context(self, question_slug):
...
         # field's parent is the fieldset - which is a valid structure object
         old_structure = self._structure
-        self._structure = self._structure.get_field(question_slug).parent()
+        self._structure = field.parent() or self._structure
         yield
         self._structure = old_structure
 
-    def _all_containers_hidden(self, question):
-        """Check whether all containers of the given question are hidden.
+    def _get_referenced_fields(self, field: Field, dependencies: List[str]):
```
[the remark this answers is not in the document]
sbor23: Thought about that too, makes sense!

caluma/caluma_core/jexl.py
```diff
@@ -114,4 +114,18 @@ def visit_Transform(self, transform):
...
+            # "mapby" is the only transform that accepts args, those can hold
+            # references. We need a "lookahead" to peek if the subject is an
+            # answer transform in this case we assume the arguments are
+            # references (as opposed to simple literals)
+            if (
+                transform.name == "mapby"
+                and transform.subject
+                and transform.subject.name == "answer"
+            ):
+                for arg in transform.args:
```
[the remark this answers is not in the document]
sbor23: Yeah since this is explicitly for `mapby`, we don't have to deal with more than one arg. No need for a general solution.

caluma/caluma_form/structure.py
```diff
@@ -140,35 +140,40 @@ def fields(self):
...
+                for field in self.children()
+                for child in field.children()
+                if field.question.type == Question.TYPE_TABLE
+            ]
         return self._sub_forms
 
-    def get_field(
-        self, question_slug: str, check_parent: bool = True
-    ) -> Optional[Field]:
+    def get_fields(self, question_slug: str, check_parent: bool = True) -> List[Field]:
```
czosel, line 157: Can this ever return more than one field?

caluma/caluma_form/jexl.py
```diff
@@ -103,93 +94,96 @@ def use_question_context(self, question_slug):
...
         )
         if all_deps_hidden:
             self._cache["hidden"][cache_key] = True
             return True
 
         # Also check if the question is hidden indirectly,
         # for example via parent formquestion.
-        if self._all_containers_hidden(question):
+        parent = field.parent()
+        if parent and parent.question and self.is_hidden(parent):
```
czosel, line 148: Why is `all_containers_hidden` not needed anymore? Do we assume that there is always only one path to a given question?

caluma/caluma_form/signals.py
```diff
@@ -19,3 +31,154 @@ def set_historical_answer_type(sender, **kwargs):
...
+    for field in struc.get_fields(question.slug):
+        jexl = QuestionJexl(
+            {
+                "form": field.form,
+                "document": field.document,
+                "structure": field.parent(),
+            }
+        )
+
+        value = jexl.evaluate(field.question.calc_expression, raise_on_error=False)
```
czosel, line 93: Why is it safe to ignore errors here? Might deserve a comment.

caluma/caluma_core/jexl.py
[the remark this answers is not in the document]
sbor23: Yes, I included a test for calculated questions. There we need to distinguish between literal args and "question" args so we can track updates thereof.

caluma/caluma_form/jexl.py
sbor23: Because we handle this in `is_hidden()` via recursion now. This is possible because we always look from fields, not from questions anymore. `is_required()` did this partially already.

caluma/caluma_form/signals.py
sbor23: Fair point, will do that.

caluma/caluma_form/structure.py
sbor23: Yes, in case the question is in a table, there will be more than one answer for the question and thus multiple fields.

czosel: Can you point me to the place in the code where this happens? I've re-read `get_fields` a few times and don't see where this would return more than one field. Sorry if i'm missing something trivial.

caluma/caluma_form/jexl.py
```diff
@@ -58,40 +59,30 @@ def __init__(self, validation_context=None, **kwargs):
...
+        fields = self._structure.get_fields(question_slug)
 
-        if self.is_hidden(question):
-            if question.type in [
-                Question.TYPE_MULTIPLE_CHOICE,
-                Question.TYPE_DYNAMIC_MULTIPLE_CHOICE,
-                Question.TYPE_TABLE,
-            ]:
-                return []
+        if len(fields) > 1:
```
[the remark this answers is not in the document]
sbor23: Done

caluma/caluma_core/jexl.py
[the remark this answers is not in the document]
sbor23: On a side note, this is still problematic. And think for the general case (any transforms) too. Theoretically the slug could be composed in some arbitrary expression, in which case the analyzer will fail.

Example: "('question' + '-' + 'slug')|answer"

But we can ignore this for now I think...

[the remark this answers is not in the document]
winged: Yes, this came up in a recent, unrelated discussion as well. The conclusion was that we don't support this. Answer transform generally only works on string inputs.

caluma/caluma_form/structure.py
sbor23: Huh, actually you are right, this is not finished, although calculated fields on table questions should work.
We shall address this follow-up, I think I have a matching test case already.

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+
+Every question that is referenced in a `calcExpression` will memoize the referencing calculated question.
+This list of calculation dependents is mutated whenever a calculated question is created, updated or deleted.
+
+#### Calculated Answers
+
+The results of calculated questions are of course stored in the corresponding answers.
+As it is highly dependent on JEXL expression, what kind dependencies to other questions are created and where in the form (sub-forms, tables etc.) they are located.
```
czosel, line 28: I don't get this sentence ^^

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+In our generalistic approach, this complexity is mitigated by evaluating greedly at every possible change.
+
+This leads to behavior that might be unexpected compared to other question types.
+
+First of all, of course a calculated answer is read-only for the GraphQL (and python) API.
+The answer is fetchable, but there is no `CreateDocumentCalculated...Answer`.
+
+A couple of mutation lead to the creation or update of a calculated answer.
```
czosel, line 36: `mutation` -> `mutations`

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+The answer is modified when:
+* the calculated question is modified
+* the calculated question is added to / removed from a form.
```
czosel, line 39: drop the period at the end

czosel (commented):
Just some documentation nitpicks :upside_down_face:

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+* a new document for a form containing calculated questions is modified
+* a new row document of a row form containing calculated questions is created / deleted
+
+
+### Limitations
+
+The current approach has some limitations that stem from compromises in the calculation of answers.
+
+* 1. calculated questions cannot reference other calculated questions
+* 2. calculated questions can only reference questions that are on the same form-level or lower
```
czosel, line 50: Isn't this a problem? Does this mean when I have the following structure
```
Main Form
 - Subform A
   - Question X
 - Subform B
   - Question Y
```
That I can't have `Y` be a calculated field depending on `X`?

caluma/tests/__snapshots__/test_schema.ambr
sbor23: Anyway this is not in the scope of this pr

docs/calculated-questions.md
sbor23: Which internal documentation do you mean? ;)

docs/calculated-questions.md
sbor23: Seems to miss a piece :D

docs/calculated-questions.md
sbor23: Your intuition is right, it should work the same way as `hidden` / `required`. This is what I mean by `on the same level`.

czosel: Ah, good! So that means we currently don't support JEXL expressions in table questions that have references to the parent form?

docs/calculated-questions.md
czosel: Maybe just a docstring in the corresponding class :wink:

sbor23: It doesn't really fit anywhere because there are just an number of signal together in a file... But yeah it doesn't fit in the user docs either.

docs/calculated-questions.md
sbor23: Not entirely sure about all variants, I think this should work too though.
We wrote down the possible cases and desired outcomes [here](https://github.com/projectcaluma/caluma/pull/1322/files#diff-4200093d628340ea25db028f99d1543cb5c63af275a4a670a9c08adee6444effR158), some of which are missing.
This should go in a separate issue, and the docs should be unified on this topic.

czosel: Thanks for the pointer! I think I meant `7. question in upper structure (from table row)`, which seems to be covered already. I totally agree when should consider this a separate issue. Shall we remove the limitation (`* 2. calculated questions can only reference questions that are on the same form-level or lower`) though, since it's not really specific to calculated fields and might cause more confusion than good?

caluma/caluma_form/jexl.py
czosel: Alright, thanks!

docs/calculated-questions.md
sbor23: Yep makes sense.

docs/calculated-questions.md
```diff
@@ -1,50 +1,50 @@
...
 * a new row document of a row form containing calculated questions is created / deleted
 
 
 ### Limitations
 
-The current approach has some limitations that stem from compromises in the calculation of answers.
-
-* 1. calculated questions cannot reference other calculated questions
-* 2. calculated questions can only reference questions that are on the same form-level or lower
+Currently the calculated questions cannot reference other calculated questions. This was an intentional limitation, as it would required additional validation to ensure no circular dependencies are involved.
```
czosel, line 50: This is an intentional limitation to avoid implementational complexity (dependency calculation "propagation", circular dependency checks).

caluma/caluma_form/signals.py
czosel: @sbor23 this is still missing, right?

docs/calculated-questions.md
sbor23: Done

caluma/caluma_form/signals.py
sbor23: Ah true this went missing during rebase

caluma/caluma_form/structure.py
```diff
@@ -140,35 +140,60 @@ def fields(self):
...
+        """
 
         field = self.fields.get(question_slug)
 
-        if not field and check_parent:
-            field = self.parent().get_field(question_slug) if self.parent() else None
         if field:
-            return field
+            return [field]
+        if not field and check_parent:
```
winged, line 183: The `not field and` part is redundant here. If `field` exists, we would have returned already in the line above

caluma/caluma_form/tests/test_answer.py
```diff
@@ -312,3 +313,38 @@ def validate(self, mutation, data, info):
...
+        "input": {
+            "document": to_global_id("CalculatedFloatAnswer", document.pk),
+            "question": to_global_id("CalculatedFloatAnswer", question.pk),
+            "value": 1.0,
+        }
+    }
+
+    if use_graphql:
+        result = schema_executor(query, variable_values=variables)
+        assert result.errors
```
winged, line 347: I don't think you're testing what you think you're testing ;-)

The type `CalculatedFloatAnswer` doesn't exist, so you're probably testing schema validation here instead of checking your actual code.

I suggest asserting the exact content of the errors here, or,  seeing that the schema doesn't even have the required types save a calculated answer, you may also just ditch the graphql part in this test

docs/calculated-questions.md
```diff
@@ -0,0 +1,50 @@
...
+* the calculated question is modified
+* the calculated question is added to / removed from a form
+* any referenced answer is modified
+* a new document for a form containing calculated questions is modified
+* a new row document of a row form containing calculated questions is created / deleted
+
+
+### Limitations
+
+Currently the calculated questions cannot reference other calculated questions. This is an intentional limitation to avoid implementation complexity (dependency calculation "propagation", circular dependency checks).
```
winged, line 50: I would also add that currently, no arithmetic functionality besides bracketing and basic operators are available.

winged (changes requested):
Some minor things remain, but good work otherwise!

docs/calculated-questions.md
sbor23: I would consider this a limitation of pyjexl. As discussed above this should be handled elsewhere in the docs, along with the allowed scopes / lookups in jexl fields.

caluma/caluma_form/tests/test_answer.py
sbor23: Done

caluma/caluma_form/structure.py
sbor23: done

winged (approved):

anehx:
I thought we agreed to wait with features until frontend and backend can be merged..

PR #1334, opened 2021-01-11
fix(api): __str__() should return string

winged:
The `__str__()` method, according to python, MUST return a string.
The UUIDModel's `id` field is (not surprisingly) an UUID, and thus
must be converted. Otherwise, this will break string formatting, for example.

### DIFF
--- a/caluma/caluma_core/models.py
+++ b/caluma/caluma_core/models.py
@@ -41,6 +41,9 @@ class BaseModel(models.Model):
         max_length=150, blank=True, null=True, db_index=True
     )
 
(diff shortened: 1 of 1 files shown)


open-dynaMIX (commented):
Why not just `str(self.pk)` in the `BaseModel`?

winged:
> Why not just `str(self.pk)` in the `BaseModel`?

You're right - the different `__str__()` implementations can really be boiled down to exactly that. Done!

open-dynaMIX (approved):
:tada:

PR #1335, opened 2021-01-11
fix(form): fix validation of table questions

anehx:
This is another regression of 445741333fd77494bd323078c70d3717ad54135a. Before that commit, the value was saved on table answers. This caused problems when validating table answers that were created before that change since they had a string value which caluma tried to validate as documents.


winged (approved):
Looks good!

PR #1347, opened 2021-01-22
feat(jexl): allow usage of form meta in question jexl

anehx:


caluma/caluma_form/structure.py
```diff
@@ -38,6 +38,10 @@ def root(self):
         return self
 
     def get(self, name, default=None):
+        for original, aliases in self.aliases.items():
```
sbor23, line 41: `self.aliases` will only be defined for `Fieldset`. Not sure if the jexl will strictly only work on `FieldSet` but I would prefer either error handling here or having `aliases = {}` in `Element`.

sbor23 (changes requested):

caluma/caluma_form/structure.py
anehx: Good point, fixed

sbor23 (approved):
Good job!

caluma/caluma_form/structure.py
```diff
@@ -22,6 +22,8 @@ def new_method(self, *args, **kwargs):
 
 
 class Element:
+    aliases = {}
```
winged, line 25: Is this a class property? If not, you should move it to `__init__()` instead

caluma/caluma_form/structure.py
```diff
@@ -124,11 +130,25 @@ def children(self):
         ]
 
 
+class MetaObject(Element):
```
winged, line 133: You could probably inherit from `Element, dict` so you don't have to proxy through to the "inner" dict.

Also, seeing as it doesn't actually do much - is the `Element` inheritance really required? Could you do with using a simple dict instead?

caluma/caluma_form/structure.py
```diff
@@ -38,6 +40,10 @@ def root(self):
         return self
 
     def get(self, name, default=None):
+        for original, aliases in self.aliases.items():
```
winged, line 43: I think if you flip the data structure here, you might be able to get rid of the loop and improve lookup speeds. If it's of the structure `name` -> `actual_name`, you could reduce the slow lookup loop into this:

```python
name = self.aliases.get(name, name)
```

winged (changes requested):
Nice work, but needs some cleanups & optimisations. Not sure the meta thing needs to be an `Element`

winged (approved):

PR #1351, opened 2021-01-26
fix(answer): prevent recalculation of calculated answers for default answer changes

anehx:


caluma/caluma_form/tests/test_answer.py
```diff
@@ -256,6 +256,35 @@ def test_save_default_answer_python_api(
...
+def test_save_calculated_dependency_default_answer(
+    db,
+    snapshot,
+    question,
+    answer,
+    question_option,
+    document_factory,
+    answer_factory,
+    answer_document_factory,
+    question_factory,
```
sbor23, line 269: There are a couple of unused fixture / factories used here.

sbor23 (commented):
Is this a fix or an optimization? Did the current behaviour trigger any unexpected outcome?
If so, a test case which would trigger the bug would be nice.

anehx:
> Is this a fix or an optimization? Did the current behaviour trigger any unexpected outcome?
> If so, a test case which would trigger the bug would be nice.

It's a fix and there is a test.....

caluma/caluma_form/tests/test_answer.py
anehx: removed

sbor23 (approved):
LGTM

PR #1354, opened 2021-01-27
feat: health endpoint

### LINKED ISSUE #1137: Provide a health endpoint for loadbalancer health check (e.g. Kubernetes readiness probes)
For monitoring and ReadinessChecks in Kubernetes it would be beneficial if a dedicated API endpoint (e.g. `/healthz`) would be available to check the health status of the backend. Possible django plugin that seem to be well maintained would be [django-health-check](https://github.com/KristianOellegaard/django-health-check) or [django-watchman](https://github.com/mwarkentin/django-watchman). What we're going to consider healthy in the context of caluma (DB connection possible, ...) needs to be defined.

luytena:
Introduce /healthz endpoint to allow for readiness checks in Kubernetes.
Use Django-Watchman to execute checks (already includes checks for caches and databases)
and is extended with additional checks for minio, db models and db migrations.

Resolves #1137

### DIFF
--- a/caluma/caluma_core/health_checks.py
+++ b/caluma/caluma_core/health_checks.py
@@ -1,0 +1,133 @@
+import uuid
+from io import StringIO
+
+import requests
+from django.conf import settings
+from django.core import management
+from watchman.decorators import check
+
+from caluma.caluma_form import storage_clients
+from caluma.caluma_logging.models import AccessLog
+
+
+@check
+def _check_pending_migrations(db_name):
+    """Check database for pending migrations.
+
+    Returns JSON mapping if no pending migrations, otherwise raises Exception.
+    @check django-watchman decorator catches and handles exceptions.
+    """
+    # check for unapplied migrations
+    out = StringIO()
+    management.call_command("showmigrations", "--plan", stdout=out)
+    plan = out.getvalue()
+    if "[ ]" in plan:
+        raise Exception("Database has unapplied migrations (migrate).")
+
+    return {db_name: {"ok": True}}
+
+
+@check
+def _check_media_storage_service():
+    """Check media storage service connection and operations."""
+    # generate object name
+    storage_client = storage_clients.client
+    prefix = "healthz_tmp-object_"
+    object_name = prefix + str(uuid.uuid4())
+
+    try:
+        # request upload url and upload
+        upload_url = storage_client.upload_url(object_name)
+
+        filename = "healthz_tmp-file.txt"
+        content = "Test\nTest\n"
+        files = {"file": (filename, content)}
+        upload_request = requests.put(upload_url, files=files)
+        assert upload_request.status_code == 200
+
+        # stat object
+        stat = storage_client.stat_object(object_name)
+        assert stat.object_name == object_name
+
+        # request download url and download file
+        download_url = storage_client.download_url(object_name)
+
+        download_request = requests.get(download_url)
+        assert download_request.status_code == 200
+        assert filename in download_request.text and content in download_request.text
+
+        # remove object
+        storage_client.remove_object(object_name)
+        assert not storage_client.stat_object(object_name)
+
+        return {"ok": True}
+
+    finally:
+        storage_client.remove_object(object_name)
+
+
+@check
+def _check_models(database):
+    """Check model instantion and operations on database."""
+    log = None
+    try:
+        # Instantiate model
+        log = AccessLog.objects.create(status_code=22)
+        assert (
+            log in AccessLog.objects.all()
+        ), "Failed to save model instance to database."
+
+        # Retrieve object
+        retrieved_log = AccessLog.objects.get(pk=log.pk)
+        assert log == retrieved_log, "Unexpected object retrieved from database."
+        assert (
+            retrieved_log.status_code == log.status_code
+        ), "Unexpected object data retrieved from database."
+
+        # Update object
+        log.status_code = 42
+        log.save()
+        retrieved_log = AccessLog.objects.get(pk=log.pk)
+        assert log == retrieved_log, "Unexpected object retrieved from database."
+        assert (
+            retrieved_log.status_code == log.status_code
+        ), "Unexpected object data retrieved from database."
+
+        # Remove object
+        log.delete()
+        log = None
+        assert (
+            log not in AccessLog.objects.all()
+        ), "Failed to delete model instance from database."
+
+        return {database: {"ok": True}}
+
+    finally:
+        # Cleanup
+        if log:  # pragma: no cover
+            log.delete()
+
+
+def check_media_storage_service():
+    """Check media storage service connection and operations."""
+    # Check if media storage service is available
+    if not storage_clients.client:
+        return {"media storage service (not configured)": {"ok": True}}
+
+    return {"media storage service": _check_media_storage_service()}
+
+
+def check_migrations():
+    """Check available databases for unapplied migrations."""
+    databases = sorted(settings.DATABASES)
+    checked_databases = [_check_pending_migrations(db_name) for db_name in databases]
+
+    return {"database migrations": checked_databases}
+
+
+def check_models():
+    """Check model instantiation."""
+    databases = sorted(settings.DATABASES)
+    checked_databases = [_check_models(db_name) for db_name in databases]
+
+    return {"database models": checked_databases}
--- a/caluma/caluma_core/tests/test_health_checks.py
+++ b/caluma/caluma_core/tests/test_health_checks.py
@@ -1,0 +1,334 @@
+from django.core import management
+from django.urls import reverse
+from watchman import settings as watchman_settings
+
+
+def test_db_connection_working(
+    disable_logs, track_errors, capsys, client, snapshot, minio_mock_working, db
+):
+    """Test /healthz/ endpoint response with working database connection.
+
+    Provide database access through the 'db' fixture with working connection.
+    Passes health checks for databases (connection), database models and
+    database migrations.
+    """
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert database checks pass
+    json_response = response.json()
+    assert json_response["databases"] == [{"default": {"ok": True}}]
+    assert json_response["database models"] == [{"default": {"ok": True}}]
+    assert json_response["database migrations"] == [{"default": {"ok": True}}]
+    assert response.status_code == 200
+
+    # assert exception type
+    *_, err = capsys.readouterr()
+    assert not err
+
+
+def test_db_connection_broken(
+    disable_logs,
+    track_errors,
+    capsys,
+    client,
+    snapshot,
+    db_broken_connection,
+    minio_mock_working,
+):
+    """Test /healthz/ endpoint response with broken database connection.
+
+    Uses DB fixture with broken connection.
+    Affects the health checks for databases (connection), database models and
+    database migrations, which should trigger an connection error.
+    """
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert database checks fail
+    json_response = response.json()
+    assert json_response["databases"] == [{"default": {"ok": False}}]
+    assert json_response["database models"] == [{"default": {"ok": False}}]
+    assert json_response["database migrations"] == [{"default": {"ok": False}}]
+    assert response.status_code == watchman_settings.WATCHMAN_ERROR_CODE
+
+    # assert exception type
+    *_, err = capsys.readouterr()
+    assert (
+        'django.db.utils.OperationalError: could not translate host name "not-db" to address'
+        in err
+    )
+
+
+def test_db_model_accessible(
+    disable_logs,
+    track_errors,
+    capsys,
+    client,
+    snapshot,
+    minio_mock_working,
+    db_privileges_working,
+):
+    """Test /healthz/ endpoint response when necessary DB tables are accessible.
+
+    Provide a database connected through a new test user with the necessary
+    privileges to access the tables needed in the health checks.
+    Passes database models health check and should be able to access the
+    necessary model table.
+    """
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert database models check passes
+    assert response.json()["database models"] == [{"default": {"ok": True}}]
+    assert response.status_code == 200
+
+    # assert exception type
+    *_, err = capsys.readouterr()
+    assert not err
+
+
+def test_db_model_read_error(
+    disable_logs,
+    track_errors,
+    capsys,
+    client,
+    snapshot,
+    minio_mock_working,
+    db_read_error,
+):
+    """Test /healthz/ endpoint response when necessary DB tables are inaccessible.
+
+    Provide a database connected through a new test user without the necessary
+    privileges to access the tables needed in the health checks.
+    Affects the database models health check when trying to access the
+    specified model table and triggers a read error due to lack of privileges.
+    """
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert database models check fails
+    assert response.json()["database models"] == [{"default": {"ok": False}}]
+    assert response.status_code == watchman_settings.WATCHMAN_ERROR_CODE
+
+    # assert exception type
+    *_, err = capsys.readouterr()
+    assert (
+        "django.db.utils.ProgrammingError: permission denied for relation caluma_logging_accesslog"
+        in err
+    )
+
+
+def test_db_model_write_error(
+    disable_logs,
+    track_errors,
+    capsys,
+    client,
+    snapshot,
+    minio_mock_working,
+    db_write_error,
+):
+    """Test /healthz/ endpoint response when necessary DB tables are read-only.
+
+    Provide a database connected through a new test user without the necessary
+    privileges to write to the tables needed in the health checks.
+    Affects the database models health check when trying to create, update or
+    delete an object for the specified model table and triggers write error
+    due to lack of privileges.
+    """
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert database models check fails
+    assert response.json()["database models"] == [{"default": {"ok": False}}]
+    assert response.status_code == watchman_settings.WATCHMAN_ERROR_CODE
+
+    # assert exception type
+    *_, err = capsys.readouterr()
+    assert (
+        "django.db.utils.ProgrammingError: permission denied for relation caluma_logging_accesslog"
+        in err
+    )
+
+
+def test_minio_connection_working(
+    disable_logs, track_errors, capsys, client, snapshot, db, minio_mock_working
+):
+    """Test /healthz/ endpoint response when minio connection is working.
+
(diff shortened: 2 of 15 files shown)


caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Check database for pending migrations.
+    Returns JSON mapping if no pending migrations, otherwise raises Exception.
+    @check decorator catches and handles exception.
+    """
+
+    # check for changes i.e. migrations that haven't been made
+    # makemigrations --check exists with exit status 1 if changes exist
+    try:
+        management.call_command("makemigrations", "--check", "--dry-run")
```
winged, line 33: I think this bit could be left out, as it's enforced in CI anyway. If you run into this situation on a prod container, something's gone reeeeally wrong

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+    except SystemExit:
+        raise Exception("Migrations aren't up-to-date (makemigrations).")
+
+    # check for unapplied migrations
+    try:
+        out = StringIO()
+        management.call_command("migrate", "--plan", stdout=out)
+        plan = out.getvalue()
+        print(plan)
```
winged, line 42: Don't `print()` - this will end up in the logs, which we don't want.

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+        if not "No planned migration operations." in plan:
```
winged, line 43: I'd try to find another pattern or another way to check. This output is very likely translated and thus the check will stop working if someone runs it on a non-english machine (or possibly even when using a non-english browser during testing)

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+            raise Exception("Database has unapplied migrations (migrate).")
+
+        return {db: {"ok": True}}
+
+    finally:
+        # Cleanup
+        out.close()
```
winged, line 50: Good thinking, but not really needed - The garbage collector will do the work for you anyway.

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+    Check minio connection and operations.
+    """
+
+    minio_client = storage_clients.client
+    response = None
+
+    # Check minio operations
+    try:
+        object_name = "my-object"
+        new_object_name = "my-object-new"
```
winged, line 65: Try to generate request-specific (randomized) object names. If two checks happen at the same time, you may get errors despite the system working correctly. I suggest using the `uuid` module for that. Then, also prefix the object's name so people will know what it is for, if it's not deleted for some reason

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+        # Remove object
+        minio_client.remove_object(new_object_name)
+        assert not minio_client.stat_object(new_object_name)
+
+        return {"ok": True}
+
+    finally:
+        # Cleanup
+        if response:
+            response.close()
```
winged, line 99: Again, not really needed, the garbage collector will take care of it

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def check_minio():
+    """
+    Check minio connection and operations.
+    """
+
+    # Check if minio is available
+    if not storage_clients.client:
+        return {"minio (not implemented)": {"ok": True}}
```
winged, line 157: Maybe change the string to "minio (not configured)" so people don't assume the minio check isn't implemented

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,213 @@
...
+        if key in dictionary:
+            del dictionary[key]
+    for value in dictionary.values():  # retrieve nested values
+        if isinstance(value, dict):
+            remove_keys(value, keys)  # apply to dict
+        if isinstance(value, list):
+            [remove_keys(v, keys) for v in value]  # apply to list
+
+
+def status(request):
```
winged, line 199: Since this is a view, I'd move it to the `views.py` module instead

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+from caluma.caluma_form import storage_clients
+from django.db.migrations.executor import MigrationExecutor
+
+from minio import Minio
+from minio.definitions import Object as MinioStatObject
+from minio import ResponseError
+
+from django.core import management
+
+import sys
```
winged, line 28: Hmm, don't we have `isort` checking anymore? Anyway, please run your code through `isort` to clean up your imports

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+    """
+    Test /healthz/ endpoint response when database connection is
+    available or unavailable.
+    Conditionally provides database access through the 'db' fixture.
+    This affects the watchman check for databases and the db models
+    check and prevents access to the database.
+    """
+
+    # use database fixture if test in working state
+    if working_db:
```
winged, line 45: This test is basically split in two, with only two lines of common code. I think it would be better to remove the parametrization and turn it into two separate tests. This may also solve your problem with the fixtures and the migrations

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+
+def switch_user_caluma():
+    """
+    Switch database user to 'tester' with password 'test'
+    """
+
+    switch_user("caluma", "caluma")
+
+
+def create_test_user():
```
winged, line 105: You're using this together with `remove_test_user()` in a fixture only. You could simplify this code by turning it into a fixture as well

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+def remove_test_user():
+    """
+    Remove user 'tester' and revoke privileges for health checks
+    """
+
+    # get db connection
+    cursor = connections["default"].cursor()
+
+    # revoke privileges and remove user
+    cursor.execute("REVOKE ALL ON caluma_logging_accesslog FROM tester;")
```
winged, line 131: No need to `REVOKE` - this happens implicitly when dropping the user

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+    cursor.execute("DROP USER tester;")
+
+
+def modify_privileges(sql):
+    """
+    Execute SQL statements to modify privileges
+    """
+
+    # retrieve db connection and execute sql statements
+    cursor = connections["default"].cursor()
```
winged, line 145: You can also access `django.db.connection.cursor()` directly (singular connection!). This is an alias for the default DB, and may be simple enough to avoid having this helper function (`modify_privileges()`) at all.

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,542 @@
...
+    mocker.patch.object(Minio, "put_object")
+
+    # replace return values
+    Minio.stat_object.side_effect = side_effect
+    Minio.get_object.side_effect = side_effect
+    Minio.put_object.return_value = None
+
+
+@pytest.mark.parametrize("working_minio", [True, False])
+def test_minio_read(request, client, snapshot, db, working_minio):
```
winged, line 448: This one would probably also benefit from splitting it into two tests instead

winged (changes requested):
Awesome work so far! Most of the requested changes are small cleanups. Overall I think this looks really good already! :+1:

caluma/caluma_core/health_checks.py
luytena: Searching for pattern `[ ]` in output of `showmigrations --plan`.
For later versions: In Django 3.1 `migrate --plan` exits with a non-zero status when unapplied migrations are detected ([link](https://docs.djangoproject.com/en/3.1/ref/django-admin/#migrate)).

caluma/caluma_core/health_checks.py
luytena: Added `uuid` to health check. The associated Pytest ran into issues because it couldn't predict the name of the mocked object it was supposed to return. At the moment the health check ignores a wrong object name if it's running in a Pytest. How does one usually solve this kind of problem?

caluma/caluma_core/health_checks.py
luytena: Does one never close files etc. in Python because the garbage collector eventually takes care of it? Or it's just not necessary to double check in a `finally` block?

caluma/caluma_core/tests/test_health_checks.py
luytena: Split all tests, because this was the case for all of them. Removed duplicate tests, usually the ones that checked that everything was working with a `200` response. 
It _did_ solve the problem with the fixtures and the migrations.

caluma/caluma_core/tests/test_health_checks.py
luytena: When the `REVOKE` statements are removed the tests trigger a `role "tester" cannot be dropped because some objects depend on it` error with reference to the privileges that weren't revoked explicitly. Is there a command that handles the dropping of those privileges automatically?

caluma/caluma_core/tests/test_health_checks.py
luytena: The db connection error test now uses a db fixture that cannot connect to the database (by modifying the database host setting). Instead of not requesting the `db` fixture.

caluma/caluma_core/health_checks.py
winged: The "proper" way would be to use a `with` block, but in this context it's not really usable. I'd do an explicit `close()` if we need to assume that other parts of the code may have a reference to the file, but we do want it to be closed. But that's not the case here, so we can trust the GC to do it for us

caluma/caluma_core/health_checks.py
winged: Aah, didn't think of that :D I'll send you a link to an internal project that contains code to mock UUID for exactly that purpose.

winged (approved):

winged:
Great work! :tada:

winged:
As per "in-person" review, following things were discovered and should be addressed:

* the UUID mocker/fixture is static, and could be reduced to a single `mocker.patch()` line
* the URLs in the test could be resolved using the `reverse()` helper
* Exception in `switch_user()` is not used, thus can be removed, as it's purely test code

winged (approved):
Looks good to me!

Lets await the review form @open-dynaMIX, but from my side, it looks ready to merge

caluma/caluma_core/tests/test_health_checks.py
```diff
@@ -0,0 +1,445 @@
...
+    # get /healthz/ response and compare to previous snapshot
+    response = client.get(reverse("healthz"))
+    snapshot.assert_match(response.json())
+
+    # assert db migrations check fails
+    assert response.json()["db migrations"] == [{"default": {"ok": False}}]
+    assert response.status_code == watchman_settings.WATCHMAN_ERROR_CODE
+
+    # re-apply migrations for app contenttypes
+    management.call_command("migrate", "contenttypes")
```
open-dynaMIX, line 445: Is this really needed? If yes, I'd put it in a fixture, to make sure it's also run, in case of an exception:

```python
@pytest.fixture
def roll_back_migrations():  # maybe you come up with a better name ;)
    # undo applied migrations for app contenttypes
    management.call_command("migrate", "contenttypes", "zero")

    yield

    # re-apply migrations for app contenttypes
    management.call_command("migrate", "contenttypes")
```

caluma/caluma_core/urls.py
```diff
@@ -1,8 +1,12 @@
...
 
+from caluma.caluma_core import views
 from caluma.caluma_user.views import AuthenticationGraphQLView
 
 urlpatterns = [
-    url(r"", AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG), name="graphql")
+    url(r"healthz", views.health_check_status, name="healthz"),
+    url(
+        r"", AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG), name="graphql"
+    ),
```
open-dynaMIX, line 11: I don't think this works as intended:

```
$ docker-compose run --rm caluma python manage.py show_urls
Creating caluma_caluma_run ... done
/graphql	caluma.caluma_user.views.AuthenticationGraphQLView	graphql
/graphqlhealthz	caluma.caluma_core.views.health_check_status	healthz
```

caluma/caluma_core/health_checks.py
```diff
@@ -0,0 +1,140 @@
...
+from django.conf import settings
+from django.core import management
+from watchman.decorators import check
+
+from caluma.caluma_form import storage_clients
+from caluma.caluma_logging.models import AccessLog
+
+
+@check
+def _check_pending_migrations(db):
```
open-dynaMIX, line 13: I'd prefer having a different name for this argument (e.g. `db_name`) in order to not confuse it for the `db` fixture.

requirements-dev.txt
```diff
@@ -28,3 +28,4 @@ python-semantic-release==7.3.0
 requests-mock==1.8.0
 reuse==0.12.1
 syrupy==0.8.4
+django-watchman==1.2.0
```
open-dynaMIX, line 31: This should go into `requirements.txt`. And we try to keep our requirements files in alphabetical order.

open-dynaMIX (changes requested):
**Awesome work!** :tada:

Some general thoughts:

## Fixtures

There quite some fixtures in the tests file. Personally I prefer fixtures to be grouped together at the top of the file (there are exceptions to that) or even in their own sub `conftest.py`. The reason for that is, that it makes it easier for the next dev to get an overview of the available fixtures.

## Potential for ddos

The healthz endpoint accepts unauthenticated requests that lead to write operations in caluma and minio. IMHO this should be mitigated. Here some possible strategies:

* Implement some form of access control. This doesn't necessarily need to be solved via OIDC, but could also be IP based (assuming only a small number of hosts need to access the healthz endpoint)
* Implement some form of rate-limiting

What are your thoughts on that @winged ?

open-dynaMIX:
@luytena Additionally, I suggest using the `settings` fixture in the tests that need access to the settings. The `settings` fixture automatically resets the settings after every test.

This means you never need things like this:

```python
@pytest.fixture
def my_settings_fixture():
    old = settings.MY_VAL
    settings.MY_VAL = "foo"

    yield

    settings.MY_VAL = old
```

winged:
> @luytena Additionally, I suggest using the `settings` fixture in the tests that need access to the settings. The `settings` fixture automatically resets the settings after every test.
> 

As far as I know, this won't work if you only modify a dict (like `DATABASES`) within the settings. It only works if you explicitly assign a new setting like in your example

open-dynaMIX:
> As far as I know, this won't work if you only modify a dict (like `DATABASES`) within the settings. It only works if you explicitly assign a new setting like in your example

TIL. Not so cool... [Here's the corresponding `pytest-django` issue](https://github.com/pytest-dev/pytest-django/issues/601).

caluma/caluma_core/urls.py
```diff
@@ -1,8 +1,14 @@
...
+from django.urls import path
 
+from caluma.caluma_core import views
 from caluma.caluma_user.views import AuthenticationGraphQLView
 
 urlpatterns = [
-    url(r"", AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG), name="graphql")
+    path("healthz/", views.health_check_status, name="healthz"),
+    path(
+        "graphql/",
```
open-dynaMIX, line 10: Trailing slash should be optional:

```suggestion
    re_path("healthz/?", views.health_check_status, name="healthz"),
    re_path(
        "graphql/?",
```

Don't forget to import `re_path` instead of `path`.

setup.py
```diff
@@ -83,5 +83,6 @@ def run(self):
         "requests<3",
         "urllib3<2",
         "uwsgi<2.1",
+        "django-watchman~=1.2.0",
```
open-dynaMIX: Nitpick: Please order it alphabetical.

open-dynaMIX (changes requested):
Awesome! Thanks for addressing the mentioned points. :rocket: 

There are some more things I'd like to have addresses, but we're getting closer :)

Additionally, please...

* ... add a switch to enable the `healthz` endpoint (default: disabled)
* ... add some documentation about this new feature

docs/configuration.md
```diff
@@ -142,3 +142,11 @@ We are using the sane uWSGI-defaults researched by [bloomberg](https://www.techa
 - `UWSGI_PROCNAME_PREFIX`=`"caluma "`
 
 See https://git.io/JemA2 for more information.
+
+
+## Healthz Endpoint
+The `healthz` endpoint performs a number of health checks using the django-watchman package extended by custom checks.
+This allows systems, such as Kubernetes, to perform monitoring and readiness checks on caluma.
+
```
winged, line 150:
```suggestion
Please note that when you enable it, make sure it is not reachable from the outside (proxy configuration!).
Due to it's nature, it takes some resources to check everything, which makes it prone to abuse (denial
of service, DoS).

```

caluma/caluma_core/urls.py
```diff
@@ -1,8 +1,17 @@
...
+    re_path(
+        "graphql/?",
+        AuthenticationGraphQLView.as_view(graphiql=settings.DEBUG),
+        name="graphql",
+    ),
 ]
+
+# if enabled, include healthz endpoint path
+if settings.ENABLE_HEALTHZ_ENDPOINT:
+    urlpatterns.append(re_path("healthz/?", views.health_check_status, name="healthz"))
```
open-dynaMIX, line 17: Would be nice to have a small test that asserts a 404 if `ENABLE_HEALTHZ_ENDPOINT = False`.

open-dynaMIX (changes requested):

docs/configuration.md
luytena: Thanks, added it to both the `configuration.md` and the `healthz.md`.

caluma/caluma_core/urls.py
luytena: Added a test, it has to reload the URL confs though. Is there a way to provide different values for environment variables to different tests, so that the URL confs are loaded with the correct value for that test?

winged: Just had the idea that you could make the decision in the view instead of trying to reload/rebuild the URL conf

Return a HTTP 404 response if `settings.ENABLE_HEALTHZ_ENDPOINT` is `False`, and otherwise continue executing the checks

anehx:
Resolves #1137

sbor23:
@luytena could use a rebase on top of master as some requirements are conflicting.

open-dynaMIX (approved):

luytena:
@winged The health check concerning the media storage service (minio) has been reworked. It now makes use of the available storage client and tests the media storage service through the upload / download urls. The corresponding tests have been updated and in addition, they now also check the received exception and in turn output less information to the console. Lastly, the `healthz` view now sets an `AnonymousUser` for the request.

winged:
> @winged The health check concerning the media storage service (minio) has been reworked. It now makes use of the available storage client and tests the media storage service through the upload / download urls. The corresponding tests have been updated and in addition, they now also check the received exception and in turn output less information to the console. Lastly, the `healthz` view now sets an `AnonymousUser` for the request.

Awesome work! Let's get it in!

open-dynaMIX:
# :tada:

sbor23:
Great job @luytena ! :tada:

luytena:
:tada::tada: Thanks for all the reviews everybody!

Issue #1359, opened 2021-01-29
[RFC] Form Builder: Config Import / Export


czosel:
A typical setup of a caluma installation involves multiple environments (e.g. dev, test, prod). To make form configuration more portable between different environments, it might make sense to offer some import/export functionality in the form builder. Some concrete ideas:

* export a question (including options etc.)
* export a form (including questions)
* export all forms (basically the entire `caluma_form` config)

During import, we might have to check if any of the slugs mentioned in the dump already exist in the database. If they do, I can see two options:

* remove everything associated with those slugs before the import ("hard" import)
* only create and update, no deletions (basically how `loaddata` works, "soft" import)

Example: I have question "foo" with options "a" and "b" in the database, and in the dump there is a question "foo" with options "b" and "c". The hard import would yield options "b" and "c", while the soft one would yield "a", "b" and "c".

cc @sbor23 @anehx @open-dynaMIX

PR #1367, opened 2021-02-03
feat: dump_caluma command for entire forms

sbor23:
Dump an entire form as django fixtures using `./manage.py dump_caluma --model caluma_form.Form my-form-slug`.
This command lives in core so it can extended to other design-time aspects such as workflows, questions etc.

`load_caluma` is not necessary as the built-in `loaddata` does this already.

Partially addresses #1359

### DIFF
--- a/caluma/caluma_core/management/commands/dump_caluma_config.py
+++ b/caluma/caluma_core/management/commands/dump_caluma_config.py
@@ -1,0 +1,112 @@
+import itertools
+
+from django.apps import apps
+from django.core import serializers
+from django.core.management.base import BaseCommand, CommandError
+from django.db import DEFAULT_DB_ALIAS
+from django.db.models import Q
+
+
+def get_form_filters(form_slugs):
+    return {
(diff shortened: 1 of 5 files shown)


winged:
(Not a review yet): Please add some documentation as to why it is neccessary and what it is for. Also not quite happy with the name... How about "export_config"?

sbor23:
> (Not a review yet): Please add some documentation as to why it is neccessary and what it is for. 

Sure!

> Also not quite happy with the name... How about "export_config"?

IMO it should have "caluma" in it's name for usage as package / django apps in other projects.
Also `dump` aligns with what django uses for `dumpdata`, also implies that it produces django fixtures which are readable by `loaddata`.
Maybe `dump_caluma_config`?

winged:
> IMO it should have "caluma" in it's name for usage as package / django apps in other projects.

Fair point, I didn't consider that perspective

> Maybe `dump_caluma_config`?

:+1:

winged (approved):
Nice work! :tada:

PR #1368, opened 2021-02-04
feat(events): add argument to on decorator to raise exceptions in the receiver

anehx:
This allows us to control (on a per receiver basis) whether an exception in the receiver will cause an exception in the current transaction and therefore rollback said transaction.

### DIFF
--- a/caluma/caluma_core/events.py
+++ b/caluma/caluma_core/events.py
@@ -1,35 +1,65 @@
 import logging
 from warnings import warn
 
-from django.dispatch import receiver
-
 logger = logging.getLogger(__name__)
 
 
-def on(event, *args, **kwargs):
+class SignalHandlingError(RuntimeError):
+    def __init__(self, exceptions, *args, **kwargs):
+        self.caught_exceptions = exceptions
+
+        super().__init__(*args, **kwargs)
+
+    def __str__(self):
+        return ", ".join([str(e) for e in self.caught_exceptions])
+
+
+def on(event, raise_exception=False, *args, **kwargs):
     deprecation_reason = getattr(event, "_deprecation_reason", None)
 
     if deprecation_reason:  # pragma: no cover
         warn(deprecation_reason, DeprecationWarning)
 
-    return receiver(event, *args, **kwargs)
+    def _decorator(func):
+        func._raise_exception = raise_exception
+
+        if isinstance(event, (list, tuple)):
+            for e in event:
+                e.connect(func, **kwargs)
+        else:
+            event.connect(func, **kwargs)
+        return func
+
+    return _decorator
 
 
 def send_event(signal, **kwargs):
     """
     Send events and handle exceptions in receiver functions.
 
-    This wrapper handles the sending of signals as well as handling of any Exceptions that
-    occur. Exceptions are logged with `ERROR` level, but not raised, in order to prevent
-    rollback of the transaction.
+    This wrapper handles the sending of signals as well as handling of any
+    Exceptions that occur. Exceptions are logged with `ERROR` level per
+    default, but not raised, in order to prevent rollback of the transaction.
+    However, if the `raise_exception` flag on the receiver is set to `True`
+    it will collect those errors and raise a `SignalHandlingError` containing
+    all caught exceptions.
     """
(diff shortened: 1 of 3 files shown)


caluma/caluma_core/events.py
```diff
@@ -1,18 +1,26 @@
 import logging
 from warnings import warn
 
-from django.dispatch import receiver
-
 logger = logging.getLogger(__name__)
 
 
-def on(event, *args, **kwargs):
+def on(event, raise_exceptions=False, *args, **kwargs):
```
open-dynaMIX, line 7: It will just raise the first exception that occurs, so maybe make this arg singular.

docs/events.md
```diff
@@ -97,6 +98,8 @@ There are plans to provide non-blocking event receivers, so stay tuned.
 
 Exceptions in event receivers are logged, but will not affect the current db transaction.
 That way, you can be sure the transaction will not be rolled-back because of an Exception in an event receiver.
+However, if the `raise_exceptions` argument of the receiver is `True` it will raise any
+exceptions that ocurred in the receiver function and the transaction will be rolled back.
```
open-dynaMIX, line 102: Same here, max one exception will be raised per receiver.

open-dynaMIX (changes requested):
Awesome! Should also be reviewed by someone else, because it was part my idea :)

sbor23 (approved):
Neat, LGTM!

caluma/caluma_core/events.py
```diff
@@ -1,18 +1,26 @@
...
     deprecation_reason = getattr(event, "_deprecation_reason", None)
 
     if deprecation_reason:  # pragma: no cover
         warn(deprecation_reason, DeprecationWarning)
 
-    return receiver(event, *args, **kwargs)
+    def _decorator(func):
+        func._raise_exceptions = raise_exceptions
+
+        if isinstance(event, (list, tuple)):
```
winged, line 16: Not sure this is a good idea. I'd prefer chaining decorators over this

caluma/caluma_core/events.py
anehx: True! Fixed it

docs/events.md
anehx: Fixed

caluma/caluma_core/events.py
```diff
@@ -26,6 +34,9 @@ def send_event(signal, **kwargs):
     result = signal.send_robust(**kwargs)
     for func, exc in result:
         if exc is not None:
+            if func._raise_exception:
```
winged, line 37: This will raise the first exception, which could mask out if there were others in other event handlers. I'm ok with only raising one, but the others should at least be logged.

Maybe we could instead raise our own exception here and add the caught ones to it's properties:
```python
class SignalHandlingError(RuntimeError):
    def __init__(self, msg, exceptions):
        self.caught_exceptions = exceptions
        # super init etc omitted
```

winged (changes requested):
Cool idea! I do have some suggestions though

caluma/caluma_core/events.py
anehx: This is copied from the original implementation of djangos receiver. Removing it would be a breaking change.

winged: Fair point, in this case we *must* keep it :)

caluma/caluma_core/events.py
anehx: Nice! Should the argument then be named `raise_exceptions` (plural) again?

open-dynaMIX: IMO no, because it's max one exception per receiver.

anehx: True, I'll leave it like this!

anehx: Done

open-dynaMIX (approved):

open-dynaMIX:
:tada:

anehx:
I'll go ahead and merge this since it's kind of urgent. @winged I implemented your feedback and reviewed it with @open-dynaMIX. If you still have feedback just go ahead, I'll process it later.

PR #1375, opened 2021-02-11
Drop calc answer validation, add useful jexl transforms

sbor23:
cc @aziiee

### DIFF
--- a/caluma/caluma_core/jexl.py
+++ b/caluma/caluma_core/jexl.py
@@ -1,3 +1,6 @@
+import math
+import numbers
+from functools import partial
 from itertools import count
 from logging import getLogger
 
@@ -67,6 +70,55 @@ class JEXL(pyjexl.JEXL):
         super().__init__(*args, **kwargs)
         self.add_transform("debug", self._debug_transform)
         self._expr_stack = []
+
+        self.add_transform(
+            "mapby", lambda arr, key: [obj.get(key, None) for obj in arr]
+        )
+        self.add_binary_operator(
+            "intersects", 20, lambda left, right: any(x in right for x in left)
+        )
+
+        def _handle_error(func, subject, *args):
+            try:
+                if not args:
+                    return func(subject)
+                return func(subject, *args)
+            except (TypeError, ValueError, ZeroDivisionError):
+                return None
+
+        def _handle_error_with_filter(func, subject, *args):
+            return _handle_error(
+                func,
+                [
+                    s
+                    for s in subject
+                    if isinstance(s, numbers.Number) and not math.isnan(s)
+                ],
+                *args,
+            )
+
+        # list operations
+        self.add_transform("min", partial(_handle_error_with_filter, min))
+        self.add_transform("max", partial(_handle_error_with_filter, max))
+        self.add_transform("sum", partial(_handle_error_with_filter, sum))
+        self.add_transform(
+            "avg",
+            partial(_handle_error_with_filter, lambda arr: sum(arr) / len(arr)),
+        )
+        # float operations
+        self.add_transform("ceil", partial(_handle_error, math.ceil))
+        self.add_transform("floor", partial(_handle_error, math.floor))
+        self.add_transform(
+            "round",
+            partial(
+                _handle_error,
+                self._round_compat,
+            ),
+        )
+
+    def _round_compat(self, num, ndigits=0):
+        power = 10 ** ndigits
+        return float(math.floor((num * power) + 0.5) / power)
 
     def parse(self, expression):
         parsed_expression = self.expr_cache.get_or_set(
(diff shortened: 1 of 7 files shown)


anehx (changes requested):
Same thing as in https://github.com/projectcaluma/ember-caluma/pull/1230#issuecomment-777249657 , we should at least add a parameter for round

caluma/caluma_core/jexl.py
```diff
@@ -68,6 +68,17 @@ def __init__(self, *args, **kwargs):
...
+        self.add_transform(
+            "mapby", lambda arr, key: [obj.get(key, None) for obj in arr]
+        )
+        self.add_binary_operator(
+            "intersects", 20, lambda left, right: any(x in right for x in left)
+        )
+        self.add_transform("min", lambda arr: min(arr))
+        self.add_transform("max", lambda arr: max(arr))
+        self.add_transform("sum", lambda arr: sum(arr))
+        self.add_transform("round", lambda num, ndigits=None: round(num, ndigits))
```
anehx, line 80: Error handling is missing. It's absolutely possible that `None` is coming in as `num` or in an `arr`. Did you guys test this, except from the unit tests?

anehx (changes requested):

caluma/caluma_core/jexl.py
sbor23: Shouldn't we just crash in such a case? Or should we really do a catch-all and return None?

anehx: No we definitely shouldn't. I don't know what the requirements are but it needs to be handled without crashing.

caluma/caluma_core/jexl.py
```diff
@@ -68,6 +70,29 @@ def __init__(self, *args, **kwargs):
...
+
+        def _handle_type_error(func, *args):
+            try:
+                return func(*args)
+            except TypeError:
+                return None
+
+        self.add_transform("min", partial(_handle_type_error, min))
+        self.add_transform("max", partial(_handle_type_error, max))
+        self.add_transform("sum", partial(_handle_type_error, sum))
```
anehx, line 88: Those should not return `None` but filter out the faulty values (non numbers)

anehx: Also, `sum` is not implemented in JS yet - did you discuss this?

caluma/caluma_core/tests/test_jexl.py
```diff
@@ -103,3 +103,81 @@ def test_jexl_cache():
...
+@pytest.mark.parametrize(
+    "expression,result",
+    [
+        ("[1, 2, 3]|min", 1),
+        ("['a', 'b']|min", "a"),
+        ("[1, 2, '3']|min", None),
+        ("['a', None]|min", None),
+        ("[1, 2, 3]|max", 3),
+        ("['a', 'b']|max", "b"),
+        ("[1, 2, '3']|max", None),
```
anehx, line 146: This for example should return `2` not `None`

anehx (changes requested):

caluma/caluma_core/jexl.py
aziiee: `sum` is now implemented in JS. @sbor23 can you add `avg` as well?

anehx (changes requested):
??? nothing changed since my last review...

caluma/caluma_core/tests/test_jexl.py
winged: TBH I'm not entirely sure where to go with this regarding user friendlyness. Not saying following the JS implementation is wrong, but:

* Should we accept non-numbers at all?
* Should we try to convert non-number arguments to numbers first?
* What if we can't convert - fail or just ignore said argument?

I think we'd be in a better place if we maybe even used separate operators for string versus number variants of the functions: We could use `|asort` to sort alphabetically, followed by `|first` or `|last` to do the same for textual `min`/`max`. Then we could do some more sane checking/enforcing/converting for the numerical variants

sbor23:
> ??? nothing changed since my last review...

My bad, it's there now.

caluma/caluma_core/tests/test_jexl.py
anehx: We already discussed this in detail...

1. Yes, we should. Imagine this: a calculation over a table field that is optional. Some values are bound to be `None` - this implementation would fail.
2. No, we should not. If someone tries to calculate with a text field and it suddenly works because we convert it it would confuse. Also, we'd need to do that for everything in those calculated fields to be "fair"

sbor23: @winged I had the same thoughts. But the application here is very domain specific so it makes sense to narrow it down.
@anehx python `sort` also sorts string alphabetically, that's why. But I don't really see the application in our jexls either.

winged: OK, I wasn't part of that discussion, so didn't know there's an explicit decision already. I only mentioned the string variants because we have test parametrisations for that just a few lines further up. Feel free to disregard my feedback

anehx (approved):

PR #1376, opened 2021-02-11
feat(form): add python API and command for copying forms

anehx:


caluma/caluma_form/domain_logic.py
```diff
@@ -200,3 +200,24 @@ def create(
...
+
+class CopyFormLogic:
+    @staticmethod
+    def copy(validated_data: dict, user: Optional[BaseUser] = None):
+        source = validated_data["source"]
+        validated_data["meta"] = dict(source.meta)
+        form = BaseLogic.create(model=models.Form, user=user, **validated_data)
+
+        for sort, form_question in enumerate(
+            reversed(models.FormQuestion.objects.filter(form=source)), start=1
```
sbor23, line 213: Use `.order_by("sort")` instead of `reversed` to make it more clear what this is about.

sbor23 (commented):
Just a nitpick, looking good otherwise :+1:

caluma/caluma_form/domain_logic.py
anehx: Actually, this whole sorting is useless :rofl: I now set `sort` to `form_question.sort` which is what we want here.

caluma/caluma_form/api.py
```diff
@@ -68,3 +68,25 @@ def save_document(
...
     domain_logic.SaveDocumentLogic.update(document, form=form, meta=meta)
     return document
+
+
+def copy_form(
+    source: models.Form,
+    slug: str,
+    name: Optional[str] = "",
+    description: Optional[str] = "",
+    is_published: Optional[bool] = False,
```
winged, line 78: If you default to empty string or `False`, it's not really an `Optional` type. Just `str`, or `bool`, respectively.

`Optional[YourType]` is short for `Union[NoneType, YourType]`, and does not denote whether the parameter is optional or not

caluma/caluma_form/management/commands/copy_form.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def handle(self, *args, **options):
+        source_slug = options["source"]
+        new_slug = options["slug"]
+
+        try:
+            source = Form.objects.get(pk=source_slug)
+        except Form.DoesNotExist:
+            self.stdout.write(f"Form '{source_slug}' not found - can't copy")
+            return
```
winged, line 59: Should return / exit with error here

caluma/caluma_form/tests/test_form.py
```diff
@@ -106,28 +107,36 @@ def test_save_form_created_as_admin_user(db, form, admin_schema_executor, admin_
     assert result.data["saveForm"]["form"]["createdByUser"] == admin_user.username
 
 
+@pytest.mark.parametrize("use_graphql", [True, False])
 @pytest.mark.parametrize("form__meta", [{"meta": "set"}])
-def test_copy_form(db, form, form_question_factory, schema_executor):
+def test_copy_form(db, form, form_question_factory, schema_executor, use_graphql):
```
winged, line 112: The `use_grapql` parameter splits the test in two parts, with almost no common code. Not worth the parametrisation. Just write two tests, it will be much more readable.

winged (changes requested):
Looks good! Some minor issues to be addressed however

caluma/caluma_form/tests/test_form.py
anehx: Actually, this is exactly how we did it for the most of the other extracted API functions. This test shares like 5 assertions and _most importantly_ the same input to make sure the both APIs do the same.

caluma/caluma_form/domain_logic.py
sbor23: That's also correct, but you might have gaps in the `sort`. With the previous method there would be nice refresh sequence 1..n.

caluma/caluma_form/tests/test_form.py
winged: The places where we did it are getting uglier each time we need to extend them. It was a bad idea from the start, just not obvious back then. And we can still factor out the asserting code. Still cleaner IMHO

caluma/caluma_form/domain_logic.py
anehx: Not really, those are being "refreshed" every time one of them change. Also, it's supposed to be a 1:1 copy.

caluma/caluma_form/tests/test_form.py
anehx: Ok, we now have a duplicated test.

sbor23 (approved):

winged (approved):

PR #1380, opened 2021-02-12
fix: use homebrew jexl analyzer for question extraction

sbor23:
Change analyzer baseclass to include array literals when walking AST.

Previously the ValidatingAnalyzer would not walk Array or ObjectLiterals, thus not reaching nested AST nodes.


czosel (approved):
:+1:

PR #1392, opened 2021-02-25
fix(filters): fix search filter when used in collection

anehx:
Using the search filter in the filter collection resulted in an error where the filter did not have a model. This was caused because the
class method `filterset.get_filters()` was used instead of the properly initialized `filterset.filters` where the model was properly set.


winged (approved):
:tada:

PR #1395, opened 2021-03-01
chore: upgrade deps

sbor23:


winged (approved):
Awesome work! :tada:

caluma/caluma_core/management/commands/cleanup_history.py
```diff
@@ -3,9 +3,11 @@
...
 # Instances which are created at run-time, referencing config-time models
 # eg. Answers are created at run-time by users and referencing Question
+# NOTE: This checks indirectly if not-nullable referenced objects exist.
+# Theoretically it doesn't matter which field to check for, but localized fields seem flaky.
 RELATED_HISTORICAL_MODELS = {
     "Answer": {"question__type__isnull": True},
-    "Document": {"form__name__isnull": True},
+    "Document": {"form__is_published__isnull": True},
     "DynamicOption": {"question__type__isnull": True},
 }
```
anehx, line 12: Wouldn't checking for the related `pk` make more sense? Or am I missing something?

caluma/caluma_form/factories.py
```diff
@@ -1,3 +1,4 @@
+import faker
 from factory import Faker, LazyAttribute, Maybe, SubFactory, lazy_attribute
```
anehx, line 2: Do we still need `Faker` here? It seems like we're importing it kind of twice.

caluma/caluma_form/tests/test_option.py
```diff
@@ -50,26 +50,3 @@ def test_copy_option(db, option, schema_executor):
...
-        }
-    """
-
-    dynamic_option = dynamic_option_factory()
-    inp = {"filter": [{"document": dynamic_option.id}]}
-    result = schema_executor(query, variable_values=inp)
-
-    assert not result.errors
-    assert dynamic_option.slug == "service-bank-arm"
-    assert dynamic_option.label["en"] == "Calvin Graham"
```
anehx, line 75: Why was this test deleted?

caluma/caluma_core/management/commands/cleanup_history.py
sbor23: The point of this command is clean up broken data, so required ForeignKeys that don't exist anymore. This wouldn't be possible in normal usage and is not allowed by Django, but can only happen by mishandling fixtures or manipulating the db directly.

Trying to access the pks directly will give an error, hence the indirection via a field of the related table.

caluma/caluma_form/factories.py
sbor23: `Faker` uses `faker.Faker` internally, but works only for field declarations in factories since the recent version.
So `Faker` is still used for declared fields, while we need `faker.Faker` for generating fake values manually at runtime.

caluma/caluma_form/tests/test_option.py
sbor23: The test would really only test if the query allUsedDynamicOptions works, which is verified in other places already, the involved filters are trivial.

Now the current  filter also doesn't make sense and the test doesn't do anything with the returned values, hence it's useless.

anehx: Alright :+1:

caluma/caluma_form/factories.py
anehx: :+1:

caluma/caluma_core/management/commands/cleanup_history.py
anehx: :+1:

anehx (approved):

PR #1408, opened 2021-03-04
feat(minio): allow disabling ssl verification

winged:
When running Caluma in a local dev environment that uses HTTPS (for
example, using the Caddy web server), working with Minio wouldn't work
anymore, as Python (and in turn, the Minio client) would check the
certificate's validity.

Introduce a new setting MINIO_DISABLE_CERT_CHECKS to bypass this in
development environments.  Explicitly verify that DEBUG is set to True
as well, as we don't want this to be used in production.


caluma/caluma_form/tests/test_minio.py
```diff
@@ -0,0 +1,30 @@
...
+        assert client.client._http.connection_pool_kw["cert_reqs"] == "CERT_NONE"
+
+        # This should succeed, as we're not verifying the server certificate
+        client.client._http.urlopen("get", "https://self-signed.badssl.com/")
+    else:
+        assert client.client._http.connection_pool_kw["cert_reqs"] == "CERT_REQUIRED"
+
+        with pytest.raises(urllib3.exceptions.RequestError):
+            # This should fail, as we are in fact verifying certificates
+            client.client._http.urlopen("get", "https://self-signed.badssl.com/")
```
sbor23, line 30: I understand that faking all this would be very cumbersome but I generally dislike calling real URLs during tests... Seems like a open gate for flaky CI behaviour and the likes. But since I don't have a better idea either we have to live with it I guess.

open-dynaMIX (approved):
Awesome! :tada:

sbor23 (approved):

caluma/caluma_form/tests/test_minio.py
winged: I do agree, but I wanted to test not only whether the setting is there, but also whether it's actually doing (or not doing, in our case) it's thing. I think it'd be obvious enough if it ever begins flaking, and we can remove the line

Issue #1394, opened 2021-03-01
modified_content properties on document


anehx:
3 new properties should be added to the document:

- `modified_content_at`
- `modified_content_by_user`
- `modified_content_by_group`

Those should return the `datetime`, `user` and `group` of the latest `HistoricalAnswer` record

sbor23:
There is one main issue with using `HistoricalAnswer`: There is `history_date` and `history_user` (which we set correctly) but no `history_group`. Now we could add a new field, hook up a signal to set the group from the request etc. See [docs](https://django-simple-history.readthedocs.io/en/latest/historical_model.html#adding-additional-fields-to-historical-models)

Instead I would propose to use the `modified_at` and add `modified_by_user` and `modified_by_group` to our base class and set it on every model directly, like we do with `created_by...`.
That's actually useful and would make the feature even simpler. This way we could even avoid the history entries completely.

anehx:
I like the idea. @winged @open-dynaMIX @czosel @kaldras what do you guys think about this?

czosel:
Agreed! :+1:

winged:
I concur with @sbor23 - don't rely on `HistoricalAnswer` for this - the feature might be disabled, or historical records might get cleaned up after a while, depending on the actual use case of the installation.

I don't think updating the `Document` at every save of an answer would be too problematic - we'd just have to update the document's family as well (to propagate across tables). Unfortunately this would create historical document entries for every answer save - which could lead to quite a bit more data there that doesn't actually reflect a change in the document model.

Thus, we might also consider read-only fields on the document that basically does `Answer.objects.filter(document__family=self).order_by('-modified_at').first()` and pulls the `modified_*` fields from there. Less redundant data this way, both in storage and in the historical records, and I think it's acceptable performance-wise.

sbor23:
Resolved in #1410

PR #1410, opened 2021-03-05
feat: modified_content properties on document

sbor23:
Implements #1394


anehx (approved):
LGTM, awesome work! :100:

caluma/caluma_form/models.py
```diff
@@ -292,6 +295,24 @@ def copy(self, family=None, user=None):
...
+    def modified_content_at(self):
+        return self.last_modified_answer.modified_at
+
+    @property
+    def modified_content_by_user(self):
+        return self.last_modified_answer.modified_by_user
+
+    @property
+    def modified_content_by_group(self):
+        return self.last_modified_answer.modified_by_group
```
winged, line 314: `last_modified_answer` can be empty - for example in a freshly created document. These properties would cause an exception in those cases. Better to guard against it and return `None` if last_modified_answer is not set

caluma/caluma_form/domain_logic.py
```diff
@@ -95,7 +109,7 @@ def create(validated_data: dict, user: Optional[BaseUser] = None) -> models.Answ
 
     @staticmethod
     @transaction.atomic
-    def update(validated_data, answer):
+    def update(answer, validated_data, user: Optional[BaseUser] = None):
```
winged, line 112: Any reason for the change in parameter ordering? If someone uses this instead of the public API, this will break their code

caluma/caluma_form/migrations/0040_auto_20210304_1411.py
```diff
@@ -0,0 +1,315 @@
+# Generated by Django 2.2.18 on 2021-03-04 14:11
```
winged, line 1: Please rename the migration to something meaningful

caluma/caluma_workflow/migrations/0027_auto_20210304_1411.py
```diff
@@ -0,0 +1,199 @@
+# Generated by Django 2.2.18 on 2021-03-04 14:11
```
winged, line 1: Please rename migration to something meaningful.

winged (changes requested):
Nice work! There's some details to be taken care of, and some questions about implementation choices. Otherwise, I like it!

caluma/caluma_form/domain_logic.py
sbor23: You said it yourself, this is considered a private api, thus it can be subject to change in a non-breaking manner.
The ordering makes things more consistent, which I feel is necessary.

winged (approved):

PR #1411, opened 2021-03-05
fix(dependencies): update minio client to adjust for changed exceptions

winged:
Minio in the newest version changed some API sturucture. Mainly, the
exception NoSuchBucket does not exist anymore, and is now part of the
S3Error exception.

While we're at it, also remove some duplicate code and move it into
a decorator (Any other place would imply more boilerplate).

### DIFF
--- a/caluma/caluma_form/storage_clients.py
+++ b/caluma/caluma_form/storage_clients.py
@@ -1,8 +1,43 @@
 from datetime import timedelta
+from functools import wraps
+from logging import getLogger
 
 import minio
 import urllib3
 from django.conf import settings
+from minio.error import S3Error
+
+log = getLogger(__name__)
+
+
+class Retry(BaseException):
+    pass
+
+
+def _retry_on_missing_bucket(fn):
+    """Create missing bucket if needed (decorator).
+
+    If enabled in the settings, try to create the bucket if it
+    doesn't exist yet, then retry.
+    """
+
+    @wraps(fn)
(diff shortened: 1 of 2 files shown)


open-dynaMIX (approved):
Beautiful! :heart_eyes:

sbor23:
Uhm, how could I miss this in the deps upgrades? I changed some stuff in the tests to reflect this, at least I thought so. But I didn't have a proper use case for integration testing.

So thanks!

sbor23 (approved):

winged:
Do not merge yet! I haven't caught this before, but in another fix I just noticed that `minio.error.NoSuchKey` is gone too!

winged:
Ready for re-review

caluma/caluma_form/storage_clients.py
```diff
@@ -1,8 +1,43 @@
 from datetime import timedelta
+from functools import wraps
+from logging import getLogger
 
 import minio
 import urllib3
 from django.conf import settings
+from minio.error import S3Error
+
+log = getLogger()
```
sbor23, line 10:
```suggestion
log = getLogger(__name__)
```

sbor23 (commented):
Excellent work :100:

caluma/caluma_form/storage_clients.py
winged: Good catch, fixed

PR #1412, opened 2021-03-05
fix(minio metadata): new minio has different metadata structure

winged:
The metadata format from Minio has changed in the recent update. While
we're at it, add some resiliency to errors - don't crash when requesting
metadata from not-yet-uploaded files, and ensure a fixed format of the
metadata at GraphQL API level.


caluma/caluma_form/tests/test_answer.py
```diff
@@ -333,3 +334,70 @@ def validate(self, mutation, data, info):
...
+            "ETag": '"5d41402abc4b2a76b9719d911017c592"',
+            "Last-Modified": "Fri, 05 Mar 2021 15:24:33 GMT",
+            "Server": "MinIO",
+            "X-Amz-Request-Id": "16697BAAD69D2214",
+            "X-Xss-Protection": "1; mode=block",
+            "Vary": "Origin",
+            "Content-Type": "binary/octet-stream",
+        },
+        "size": 8200,
+    }
```
sbor23, line 403: I would prefer a snapshot here, when we are not interested in specific properties

sbor23 (approved):
Awesome, thanks!

caluma/caluma_form/tests/test_answer.py
winged: Thought about this, but it would be too easy to lose sight of the important bits, such as the types etc. So changed it to ensure the actually important parts, as you're right, we're not interested in the specific values

sbor23: Looks good :+1:

PR #1420, opened 2021-03-10
fix(form): remove unneeded default_answer field on serializer

sbor23:


open-dynaMIX (changes requested):
Hmmm.... Shouldn't this be also removed from `fields` in the `Meta` class? And then trigger a [schema change](https://github.com/projectcaluma/caluma/blob/master/caluma/tests/__snapshots__/test_schema.ambr#L1376)? AFAIS the change as it is right now makes it writable instead of removing it.

sbor23:
> Hmmm.... Shouldn't this be also removed from `fields` in the `Meta` class? And then trigger a [schema change](https://github.com/projectcaluma/caluma/blob/master/caluma/tests/__snapshots__/test_schema.ambr#L1376)? AFAIS the change as it is right now makes it writable instead of removing it.

Of course, what was i thinking...
Just realized this would strictly speaking be breaking change.

Do you consider the fix to be urgent? I first thought it would be good to have it merged asap, but the "bug", `defaultAnswer` being part of the schema on question inputs, shouldn't be dangerous anyways (saving the answer ID on a question input should just lead to an error if that was attempted I suppose).

We can also drop this PR and include the change in https://github.com/projectcaluma/caluma/pull/1140

open-dynaMIX (approved):

PR #1425, opened 2021-03-15
fix(minio): minio client requires different call signature for copy

winged:
Use CopySource now instead of the old "bucket/object_name" string syntax
when copying / moving files in Minio.

### DIFF
--- a/caluma/caluma_form/storage_clients.py
+++ b/caluma/caluma_form/storage_clients.py
@@ -5,6 +5,7 @@ from logging import getLogger
 import minio
 import urllib3
 from django.conf import settings
+from minio.commonconfig import CopySource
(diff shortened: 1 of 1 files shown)


sbor23 (approved):
Excellent, thanks for the fixes!

On a side note: I was thinking if actual, recorded integration tests would make sense for minio.
Similar to what we did in e-learning and nextcloud.

winged:
> On a side note: I was thinking if actual, recorded integration tests would make sense for minio.
> Similar to what we did in e-learning and nextcloud.

Hmm, I guess in a way it would be useful generally. However, all the current issues were with changes of the client library, not with the "wire protocol" itself, which didn't change at all. So in this case, that wouldn't have helped.

PR #1426, opened 2021-03-16
feat(jexl): improve mapby and add stringify transform

anehx:
This extends the mapby transform to accept multiple keys as argument
which then returns an array instead of the single value. This will
result in nested arrays which is why we added a stringify transform that
converts such a collection into a string to be used in comparisons.

### DIFF
--- a/caluma/caluma_core/jexl.py
+++ b/caluma/caluma_core/jexl.py
@@ -1,3 +1,4 @@
+import json
 import math
 import numbers
 from functools import partial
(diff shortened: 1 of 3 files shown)


caluma/caluma_core/jexl.py
```diff
@@ -13,6 +14,16 @@
...
+def mapby(arr, *keys):
+    if not isinstance(arr, list):
+        return None
+
+    return [
+        [obj.get(key) for key in keys] if len(keys) > 1 else obj.get(keys[0])
+        for obj in arr
+    ]
+
+
```
sbor23, line 26: Let's keep all the transforms as methods of `JEXL` and use `self.add_transform("mapby", self.mapby)`.

sbor23 (changes requested):
Neat!

caluma/caluma_core/jexl.py
anehx: Done

sbor23 (approved):
LGTM :100:

PR #1433, opened 2021-03-22
chore: fix gitlint config for dependabot

sbor23:

PR #1434, opened 2021-03-22
chore: fix ignore line for gitlint

sbor23:
Follow-up https://github.com/projectcaluma/caluma/pull/1433

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -3,4 +3,4 @@ ignore=body-is-missing
 
 [ignore-by-title]
 regex=^chore\(deps\)(.*)
-ignore=all
+ignore=title-max-length, body-is-missing, body-max-line-length, body-min-length


.gitlint
```diff
@@ -3,4 +3,4 @@ ignore=body-is-missing
 
 [ignore-by-title]
 regex=^chore\(deps\)(.*)
-ignore=all
+ignore=body-is-missing, body-max-line-length, body-min-length
```
open-dynaMIX, line 6: `title-max-length` is missing, right?

open-dynaMIX (changes requested):

.gitlint
sbor23: looks like

open-dynaMIX (approved):

PR #1440, opened 2021-03-25
fix(file): ignore errors due to missing file

open-dynaMIX:
If no file is uploaded to the storage service, we want to ignore the resulting
exceptions in order to still be able to modify/delete the File.

### DIFF
--- a/caluma/caluma_form/tests/test_model.py
+++ b/caluma/caluma_form/tests/test_model.py
@@ -1,5 +1,5 @@
 import pytest
-from minio import Minio
+from minio import Minio, S3Error
 
 from ...caluma_form.models import File, Question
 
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/tests/test_model.py
```diff
@@ -34,3 +34,29 @@ def test_update_file(db, file_factory, mocker):
     file.save()
     Minio.copy_object.assert_called()
     Minio.remove_object.assert_called()
+
+
+@pytest.mark.parametrize("should_raise", [False, True])
+def test_missing_file(db, file, mocker, should_raise):
+    exc_code = "NoSuchKey"
+    if should_raise:
+        exc_code = "SomeOtherError"
```
sbor23, line 43: You don't like python's ternary? :)

sbor23 (commented):
Just a curious comment.

What's the reasoning behind keeping the file and renaming the pk in `_move_blob`?
We want to keep all historical versions of a file?

caluma/caluma_form/tests/test_model.py
open-dynaMIX: Changed, thx

open-dynaMIX:
> We want to keep all historical versions of a file?

Exactly

sbor23 (approved):

PR #1441, opened 2021-03-25
fix: modified_content on row documents

sbor23:


caluma/caluma_form/models.py
```diff
@@ -295,11 +295,13 @@ def copy(self, family=None, user=None):
...
 
+    def is_root(self):
+        return self == self.family
+
     @cached_property
     def last_modified_answer(self):
-        return (
-            Answer.objects.filter(document__family=self).order_by("modified_at").last()
-        )
+        _filter = "document__family" if self.is_root() else "document"
```
winged, line 303: It took me quite a while to understand why this is required. I think a comment and an extended commit message would be very useful.

Somethings along the lines of "for row documents, we want the most recently changed answer within this document, not traversing to the parent/root document. Only for root documents should we look at all answers, including rows"

winged (changes requested):
Nice catch! But please explain in code & commit message

caluma/caluma_form/models.py
```diff
@@ -295,11 +295,13 @@ def copy(self, family=None, user=None):
...
+        return Answer.objects.filter(**{_filter: self}).order_by("-modified_at").first()
```
anehx, line 304: Is it really necessary to differentiate? Can't we just do `Answer.objects.filter(Q(document=self) | Q(document__family=self)).order_by("modified_at").last()`?

caluma/caluma_form/models.py
sbor23: You are right, added comment and commit message.b

caluma/caluma_form/models.py
sbor23: That should also work and is much nicer to read! I think we could save a join by differentiating but it's not worth it.

anehx (approved):

winged (approved):

anehx:
@sbor23 can we merge and release this?

PR #1448, opened 2021-03-30
fix(workflow): correctly set workitem name from task

winged:
When a work item is created from a workflow event (triggered via task),
we used bulk_create_with_history(), which doesn't send signals. Thus,
the `set_name_and_description` signal was not triggered, leaving us with
anonymous work items.

### DIFF
--- a/caluma/caluma_workflow/domain_logic.py
+++ b/caluma/caluma_workflow/domain_logic.py
@@ -73,7 +73,7 @@ class StartCaseLogic:
         workflow = case.workflow
         tasks = workflow.start_tasks.all()
 
-        work_items = utils.bulk_create_work_items(tasks, case, user, None, context)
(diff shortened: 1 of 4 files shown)


czosel (approved):
:+1:

winged:
See also [the Django docs](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#bulk-create) regarding bulk-create.

sbor23 (approved):
LGTM, shouldn't have too much a performance impact as long as the number of tasks / groups is not too big.

open-dynaMIX (approved):
Nice catch!

winged:
> LGTM, shouldn't have too much a performance impact as long as the number of tasks / groups is not too big.

It does affect performance, but since we officially use signals for extensions, we shouldn't write code that (sometimes) doesn't send them

PR #599, opened 2019-07-22
Extract queryset from manager when no filters are set

sliverc:
Fixes #598

Issue #598, opened 2019-07-22
[regression] 'Manager' object has no attribute 'query'


open-dynaMIX:
#569 introduced a regression that is triggered in one of our projects.

Here is one example:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/promise/promise.py", line 487, in _resolve_from_executor
    executor(resolve, reject)
  File "/usr/local/lib/python3.6/site-packages/promise/promise.py", line 754, in executor
    return resolve(f(*args, **kwargs))
  File "/usr/local/lib/python3.6/site-packages/graphql/execution/middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "/app/caluma/core/filters.py", line 501, in connection_resolver
    **args,
  File "/usr/local/lib/python3.6/site-packages/graphene_django/fields.py", line 158, in connection_resolver
    queryset = cls.resolve_queryset(connection, default_manager, info, args)
  File "/usr/local/lib/python3.6/site-packages/graphene_django/fields.py", line 88, in resolve_queryset
    return connection._meta.node.get_queryset(queryset, info)
  File "/app/caluma/form/schema.py", line 152, in get_queryset
    queryset = super().get_queryset(queryset, info)
  File "/app/caluma/core/types.py", line 31, in get_queryset
    queryset = visibility_class().filter_queryset(cls, queryset, info)
  File "/app/caluma/core/visibilities.py", line 85, in filter_queryset
    result_queryset = result_queryset.union(class_result)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 946, in union
    return self._combinator_query('union', *other_qs, all=all)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 936, in _combinator_query
    clone.query.combined_queries = (self.query,) + tuple(qs.query for qs in other_qs)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 936, in <genexpr>
    clone.query.combined_queries = (self.query,) + tuple(qs.query for qs in other_qs)
graphql.error.located_error.GraphQLLocatedError: 'Manager' object has no attribute 'query'

[22/Jul/2019 12:15:02] "POST /graphql/ HTTP/1.0" 200 369
```

That's the query:

```
"query GetOptionsForFilter($slug: String!) {
  allQuestions(slug: $slug) {
    edges {
      node {
        id
        slug
        ... on ChoiceQuestion {
          options {
[...]

PR #1474, opened 2021-04-23
fix(caluma): fix regression for union visibilities

open-dynaMIX:
This commit fixes a regression that was first introduced with
cf1f91dd88480d3ae206fc790db49c649f91ccc3 and then fixed with
2c195534db89a5622b62ae93a56ad5fe9ffec2c4, but then re-introduced with
b85ae9676fb4245a6d2817f376c23c495db0c146.

This time I was able to reproduce it and this commit contains a test.

Related issues and PRs:

* #569
* #598
* #599
* #841

### DIFF
--- a/caluma/caluma_core/tests/test_visibility.py
+++ b/caluma/caluma_core/tests/test_visibility.py
@@ -1,6 +1,6 @@
 import pytest
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import CharField
+from django.db.models import CharField, QuerySet
 
 from .. import models
 from ..types import DjangoObjectType, Node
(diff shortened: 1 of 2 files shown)


caluma/caluma_core/tests/test_visibility.py
```diff
@@ -167,6 +167,42 @@ class ConfiguredUnion(Union):
     assert queryset.count() == 0
 
 
+def test_union_visibility_return_managers(db, history_mock):
+    FakeModel = get_fake_model(
+        dict(name=CharField(max_length=255)), model_base=models.UUIDModel
```
winged, line 172: Any particular reason for `dict()` vs `{}`? Not really important, but I'd prefer the literal syntax

caluma/caluma_core/visibilities.py
```diff
@@ -79,6 +80,8 @@ def filter_queryset(self, node, queryset, info):
         result_queryset = None
         for visibility_class in self.visibility_classes:
             class_result = visibility_class().filter_queryset(node, queryset, info)
+            if isinstance(class_result, Manager):
```
winged, line 83: Instead of checking the type, I'd suggest just calling `.all()` on the result. This method is supported by both managers and querysets.

caluma/caluma_core/tests/test_visibility.py
```diff
@@ -167,6 +167,42 @@ class ConfiguredUnion(Union):
...
+    class CustomNode(DjangoObjectType):
+        class Meta:
+            model = FakeModel
+
+    class Name1Visibility(BaseVisibility):
+        @filter_queryset_for(Node)
+        def filter_queryset_for_node(self, node, queryset, info):
+            return queryset
+
+    class Name2Visibility(BaseVisibility):
```
winged, line 188: Not sure having two visibility classes has any impact on the aspect being tested

caluma/caluma_core/tests/test_visibility.py
```diff
@@ -167,6 +167,42 @@ class ConfiguredUnion(Union):
...
+        visibility_classes = [Name1Visibility, Name2Visibility]
+
+    queryset = FakeModel.objects
+    result = Name1Visibility().filter_queryset(CustomNode, queryset, None)
+    assert result.count() == 4
+    result = Name2Visibility().filter_queryset(CustomNode, queryset, None)
+    assert result.count() == 4
+    queryset = ConfiguredUnion().filter_queryset(CustomNode, queryset, None)
+    assert queryset.count() == 4
+    assert queryset.get(name="Name2")
```
winged, line 203: We also should test for the result being of type `QuerySet` I think - because that's what we want, no?

winged (changes requested):
Good catch! Some simplifications should be possible though

caluma/caluma_core/tests/test_visibility.py
open-dynaMIX: It's a copy/paste from the test above.

caluma/caluma_core/tests/test_visibility.py
open-dynaMIX: It has: `.union()` is not called on the first, hence the bug is not triggered.

caluma/caluma_core/visibilities.py
open-dynaMIX: Fixed.

caluma/caluma_core/tests/test_visibility.py
open-dynaMIX: Fixed.

caluma/caluma_core/tests/test_visibility.py
open-dynaMIX: Sorry, it's the opposite: `.union()` is called on the first with the second one as argument. And that triggers the bug.

winged (approved):

PR #1505, opened 2021-05-17
Dependabot native

open-dynaMIX:


.reuse/dep5
```diff
@@ -8,7 +8,7 @@ Files:
     .dependabot/config.yml
```
winged, line 8: This line's not required anymore. Otherwise LGTM

winged (changes requested):

.reuse/dep5
open-dynaMIX: Right! Done.

winged (approved):

sbor23:
Are we considering a solution for automerging? There are some actions floating around in the marketplace, that we could easily copy one-to-one, without having a dependency on potentially untrusted sources.

PR #1509, opened 2021-05-20
feat: make selected option(s) accessible from answers

### LINKED ISSUE #1301: Field for selected option on answer
It would be very convenient to have a field `selected_option` on `StringAnswer` and a field `selected_options` on `ListAnswer` if the answer is linked to a (dynamic) (multiple) choice question.

This would help the user of the API to get the display value in an easy way. For "normal" choices it's possible to get the options relatively easy via the question. However, for dynamic choices the user may need to fetch the `allUsedDynamicOptions` task first and then aggregate those with the current dynamic options which is very cumbersome.

Those new fields would make the usage of the API when using dynamic questions alot less painful.

open-dynaMIX:
This commit adds a new field `selected_option` to `StringAnswer`,
holding the selected option for (dynamic) choice-questions and a new
connection `selected_options` to `ListAnswer` for (dynamic)
multiple-choice-questions.

Closes #1301

### DIFF
--- a/caluma/caluma_form/schema.py
+++ b/caluma/caluma_form/schema.py
@@ -739,8 +739,23 @@ class DateAnswer(AnswerQuerysetMixin, FormDjangoObjectType):
         interfaces = (Answer, graphene.Node)
 
 
+class SelectedOption(ObjectType):
+    label = graphene.String(required=True)
+    slug = graphene.String(required=True)
+
+
+class SelectedOptionConnection(CountableConnectionBase):
+    class Meta:
+        node = SelectedOption
+
+
 class StringAnswer(AnswerQuerysetMixin, FormDjangoObjectType):
     value = graphene.String()
+    selected_option = graphene.Field(SelectedOption)
+
+    def resolve_selected_option(self, info, **args):
+        selected_options = self.selected_options
+        return selected_options.first() if selected_options else None
 
     class Meta:
         model = models.Answer
@@ -751,6 +766,7 @@ class StringAnswer(AnswerQuerysetMixin, FormDjangoObjectType):
 
 class ListAnswer(AnswerQuerysetMixin, FormDjangoObjectType):
     value = graphene.List(graphene.String)
+    selected_options = ConnectionField(SelectedOptionConnection)
 
     class Meta:
         model = models.Answer
--- a/caluma/caluma_form/tests/test_document.py
+++ b/caluma/caluma_form/tests/test_document.py
@@ -1444,3 +1444,87 @@ def test_document_modified_content_properties(
     assert node["modifiedContentAt"] == sub_a.modified_at.isoformat()
     assert node["modifiedContentByUser"] == sub_a.modified_by_user
     assert node["modifiedContentByGroup"] == sub_a.modified_by_group
+
+
+@pytest.mark.parametrize(
+    "question__type,answer__value",
+    [
+        (Question.TYPE_CHOICE, "somevalue"),
+        (Question.TYPE_MULTIPLE_CHOICE, ["somevalue", "anothervalue"]),
+        (Question.TYPE_DYNAMIC_CHOICE, "somevalue"),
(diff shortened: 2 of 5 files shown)


caluma/caluma_form/schema.py
```diff
@@ -751,6 +766,10 @@ class Meta:
 
 class ListAnswer(AnswerQuerysetMixin, FormDjangoObjectType):
     value = graphene.List(graphene.String)
+    selected_options = ConnectionField(SelectedOptionConnection)
+
+    def resolve_selected_options(self, info, **args):
```
winged, line 771: Not entirely sure, it's been a while since I messed with the schema, but:  Is the resolve method really needed here? It only returns `self.selected_options` anyway.

caluma/caluma_form/tests/test_document.py
```diff
@@ -1444,3 +1444,77 @@ def test_document_modified_content_properties(
...
+    document,
+    answer,
+    question_option_factory,
+    dynamic_option_factory,
+    schema_executor,
+    question,
+):
+    query = """
+        query AllDocumentsQuery($search: String) {
+          allDocuments(search: $search) {
```
winged, line 1471: the `search` filter is deprecated (should be `filter: [ {"search": $search} ]`, but I think for this test it's not needed at all. You could just query `node($globalid)` instead and go for the answer directly. Would make the test a little bit shorter

caluma/caluma_form/tests/test_document.py
```diff
@@ -1444,3 +1444,77 @@ def test_document_modified_content_properties(
...
+        question_option_factory(option__slug=answer.value, question=question)
+    elif question.type == Question.TYPE_DYNAMIC_CHOICE:
+        dynamic_option_factory(slug=answer.value, question=question, document=document)
+
+    elif question.type == Question.TYPE_MULTIPLE_CHOICE:
+        for slug in answer.value:
+            question_option_factory(option__slug=slug, question=question)
+    elif question.type == Question.TYPE_DYNAMIC_MULTIPLE_CHOICE:
+        for slug in answer.value:
+            dynamic_option_factory(slug=slug, question=question, document=document)
```
winged, line 1516: To truly test the correct behaviour, I think we should also have at least one option (dynamic and regular) that should not be returned here

caluma/caluma_form/tests/test_document.py
```diff
@@ -1444,3 +1444,77 @@ def test_document_modified_content_properties(
...
+
+    result = schema_executor(query, variable_values={"search": search})
+    assert not result.errors
+    snapshot.assert_match(result.data)
```
winged, line 1520: I'd like to see some explicit checks here as well. Sometimes we don't look at the snapshots too closely and may miss something

winged (changes requested):
Awesome work! Needs some minor cleanups, but otherwise looks good!

caluma/caluma_form/schema.py
open-dynaMIX: Right, done.

caluma/caluma_form/tests/test_document.py
open-dynaMIX: Done.

caluma/caluma_form/tests/test_document.py
open-dynaMIX: Done.

caluma/caluma_form/tests/test_document.py
open-dynaMIX: Done.

winged (approved):
Awesome!

czosel:
We just stumbled over another use case for this - would be awesome if someone could make a release including this feature :slightly_smiling_face:

winged:
> We just stumbled over another use case for this - would be awesome if someone could make a release including this feature slightly_smiling_face

#5313

PR #1523, opened 2021-06-07
fix(validation): exclude hidden cells in answer transforms for tables

anehx:


caluma/caluma_form/jexl.py
```diff
@@ -61,6 +62,22 @@ def answer_transform(self, question_slug):
         if self.is_hidden(field):
             return field.question.empty_value()
 
+        # This overrides the logic in field.value() to consider visibility for
+        # table cells
+        elif (
+            field.question.type == Question.TYPE_TABLE
+            and field.answer is not None
+            and field.answer.value is None
```
winged, line 70: Not sure we should even do that last check (`answer.value is None`) If some code didn't keep proper data hygiene, this may return wrong data

caluma/caluma_form/tests/test_jexl.py
```diff
@@ -411,6 +411,50 @@ def test_answer_transform_in_tables(
...
+    ).question
+
+    # create question that depends on the value of the previously created hidden
+    # column
+    form_question_factory(
+        form=form,
+        question__slug="check",
+        question__is_hidden=f"('yes' in '{table_question.slug}'|answer|mapby('{col2_question.slug}'))",
+        question__is_required="true",
+    ).question
```
winged, line 443: The `.question` is not required here, we're not using the return value anyways

winged (changes requested):
Nice work! Some minor things to clean up, otherwise looks good.

caluma/caluma_form/jexl.py
anehx: True, I did this check because it's what would happen in `field.value()` so I wanted to be consistent. I'll remove it

caluma/caluma_form/tests/test_jexl.py
anehx: done

caluma/caluma_form/jexl.py
anehx: done

winged (approved):

winged:
I sort-of dislike that this introduces a coverage hole that wasn't there before, but on the other hand it's obvious why it's there. Also we can't really remove this code, as it *is* useful in certain cases (debugging, ...).

We should probably build a unified model of a form/document's structure that takes visibility jexl, calculated fields etc into account. Not today though.

PR #1526, opened 2021-06-08
fix: extend uwsgi buffer

winged:
UWSGI has a buffer of 4096 bytes by default, in which all request headers
must fit. If a user has particularly many roles/groups in their JWT token,
this limit is exceeded. UWSGI will then abort the request, and the reverse
proxy in front will respond with an error such as HTTP/502.


winged:
Related issue in Alexandria: https://github.com/projectcaluma/alexandria/pull/193

czosel (approved):

PR #1529, opened 2021-06-10
chore: update pyjexl

czosel:


open-dynaMIX (changes requested):
Change looks good.

It'd be senseful to have a commit message explaining why we're doing this. Otherwise LGTM.

czosel:
True - done!

open-dynaMIX (approved):

czosel:
cc @winged

winged:
Hmm.. A comment in the requirements file would have been the "standard procedure" for this, so we know from looking at the code when to remove the explicit reference (and are reminded more often to go check for a release). No need to reopen however

PR #1536, opened 2021-06-16
feat(jexl): allow optional answer transforms with default value

anehx:
This allows the user to write reusable JEXLs using answer transforms that don't necessarily resolve to a question in a certain form context.

https://github.com/projectcaluma/ember-caluma/pull/1435

### DIFF
--- a/caluma/caluma_core/jexl.py
+++ b/caluma/caluma_core/jexl.py
@@ -166,6 +166,10 @@ class CalumaAnalyzer(JEXLAnalyzer):
     TODO: Upstream this.
     """
 
+    def __init__(self, config, transforms=None):
+        self.transforms = transforms if transforms else []
+        super().__init__(config)
+
     def generic_visit(self, expression):
         for child in expression.children:
             assert child is not None
@@ -192,10 +196,6 @@ class ExtractTransformSubjectAnalyzer(CalumaAnalyzer):
     If no transforms are given all references of all transforms will be extracted.
     """
 
-    def __init__(self, config, transforms=None):
-        self.transforms = transforms if transforms else []
-        super().__init__(config)
-
     def visit_Transform(self, transform):
         if not self.transforms or transform.name in self.transforms:
             # can only extract subject's value if subject is a Literal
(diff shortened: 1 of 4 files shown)


caluma/caluma_form/jexl.py
```diff
@@ -109,18 +120,19 @@ def use_field_context(self, field: Field):
...
-        referenced_fields = [self._structure.get_field(slug) for slug in deps]
+        deps = list(self.extract_referenced_questions_with_arguments(expr))
+        referenced_fields = [self._structure.get_field(slug) for slug, _ in deps]
 
         referenced_slugs = [ref.question.slug for ref in referenced_fields if ref]
 
-        for slug in deps:
-            if slug not in referenced_slugs:
+        for slug, args in deps:
+            required = not len(args) or (args[0] and args[0].value)
```
winged, line 129: I think an explanatory comment would be helpful: Why is having args impacting the requiredness? In the context of the PR it's clear, but future code readers might get confused

caluma/caluma_core/jexl.py
```diff
@@ -233,3 +229,21 @@ def visit_Transform(self, transform):
...
+
+    If no transforms are given all references of all transforms will be extracted.
+    """
+
+    def visit_Transform(self, transform):
+        if not self.transforms or transform.name in self.transforms:
+            if not isinstance(transform.subject, type(transform)):
+                yield (
+                    transform.subject.value,
+                    [arg for arg in transform.args if isinstance(arg, Literal)],
```
winged, line 246: What's the reason behind only returning literal arguments? Do we even have cases where that wouldn't be true?

winged (changes requested):
Nice work, looks good! Just hoping for some minor explanations

caluma/caluma_core/jexl.py
anehx: No idea, this was taken from `ExtractTransformArgumentAnalyzer`. I can remove it if you want.

caluma/caluma_form/jexl.py
anehx: > Why is having args impacting the requiredness?

That statement is not entirely correct. It's not having any args that impacting the requiredness, it's the first argument specifically that needs to be `False` to impact the requiredness. This behaviour is clearly documented in the user docs. I really think this line right here is self-explanatory and doesn't need a comment.

winged (approved):
Nice work! Like this much better than the earlier version! Let's rebase and see if #1537 fixed it

:100:

PR #1551, opened 2021-07-30
feat(config): introduce configurable user factory

winged:
Allow users to provide a custom user factory to build the OIDCUser
object. This enables us to extend it and make it compatible with
other OIDC RP apps (such as Emeis, Alexandria etc).

### DIFF
--- a/caluma/caluma_user/views.py
+++ b/caluma/caluma_user/views.py
@@ -8,6 +8,7 @@ from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
 from django.http.response import HttpResponse
 from django.utils.encoding import force_bytes, smart_text
+from django.utils.module_loading import import_string
 from graphene_django.views import GraphQLView, HttpError
 from rest_framework.authentication import get_authorization_header
 
@@ -69,6 +70,10 @@ class AuthenticationGraphQLView(GraphQLView):
         response.raise_for_status()
         return response.json()
 
+    def _oidc_user(self, *args, **kwargs):
+        factory = import_string(settings.CALUMA_OIDC_USER_FACTORY)
+        return factory(*args, **kwargs)
+
     def get_user(self, request):
         token = self.get_bearer_token(request)
 
@@ -90,7 +95,7 @@ class AuthenticationGraphQLView(GraphQLView):
                 userinfo_method,
                 timeout=settings.OIDC_BEARER_TOKEN_REVALIDATION_TIME,
             )
-            return models.OIDCUser(token=token, userinfo=userinfo)
+            return self._oidc_user(token=token, userinfo=userinfo)
         except requests.HTTPError as e:
             try:
                 if (
@@ -108,7 +113,7 @@ class AuthenticationGraphQLView(GraphQLView):
                     if "client_id" not in introspection:
                         response = HttpResponse(status=401)
                         raise HttpError(response)
-                    return models.OIDCUser(token=token, introspection=introspection)
+                    return self._oidc_user(token=token, introspection=introspection)
                 else:
                     raise e
             except requests.HTTPError as internal_exception:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -32,6 +32,7 @@ Caluma expects a bearer token to be passed on as [Authorization Request Header F
 * `OIDC_GROUPS_CLAIM`: Name of claim to be used to represent groups (default: caluma_groups)
 * `OIDC_USERNAME_CLAIM`: Name of claim to be used to represent the username (default: sub)
 * `OIDC_BEARER_TOKEN_REVALIDATION_TIME`: Time in seconds before bearer token validity is verified again. For best security token is validated on each request per default. It might be helpful though in case of slow Open ID Connect provider to cache it. It uses [cache](#cache) mechanism for memorizing userinfo result. Number has to be lower than access token expiration time. (default: 0)
+* `CALUMA_OIDC_USER_FACTORY`: User object factory (default: `caluma.caluma_user.models.OIDCUser`). Use it to provide a custom OIDC user object. The factory is expected to accept a mandatory `token` parameter and two optional parameters `userinfo` and `introspection`. Only one of them will be filled, depending on which OIDC endpoint the user information comes from.
 
 ## Cache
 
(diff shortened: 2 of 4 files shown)


docs/configuration.md
```diff
@@ -32,6 +32,7 @@ Caluma expects a bearer token to be passed on as [Authorization Request Header F
 * `OIDC_GROUPS_CLAIM`: Name of claim to be used to represent groups (default: caluma_groups)
 * `OIDC_USERNAME_CLAIM`: Name of claim to be used to represent the username (default: sub)
 * `OIDC_BEARER_TOKEN_REVALIDATION_TIME`: Time in seconds before bearer token validity is verified again. For best security token is validated on each request per default. It might be helpful though in case of slow Open ID Connect provider to cache it. It uses [cache](#cache) mechanism for memorizing userinfo result. Number has to be lower than access token expiration time. (default: 0)
+* `CALUMA_OIDC_USER_FACTORY`: User object factory. Use it to provide your own OIDC user object
```
czosel, line 35:
```suggestion
* `CALUMA_OIDC_USER_FACTORY`: User object factory. Use it to provide your own OIDC user object (default: "caluma.caluma_user.models.OIDCUser")
```
That way it's the expected format is a little more obvious.

czosel: Also, should we call it `CALUMA_OIDC_USER_FACTORY_CLASS` to be consistent with the other extension points?

caluma/caluma_user/views.py
```diff
@@ -108,7 +113,7 @@ def get_user(self, request):
                     if "client_id" not in introspection:
                         response = HttpResponse(status=401)
                         raise HttpError(response)
-                    return models.OIDCUser(token=token, introspection=introspection)
+                    return self._oidc_user(token=token, introspection=introspection)
```
czosel, line 116: Should be add some documentation about the two different cases? (userinfo, introspection)

docs/configuration.md
winged: > Also, should we call it `CALUMA_OIDC_USER_FACTORY_CLASS` to be consistent with the other extension points?

No, as the factory is just that - a factory. Could just as well be a standalone function that returns some object. The other ones are always classes of a specific type

czosel: :+1: makes sense, thanks!

czosel: I'll leave this "unresolved" because it was actually 2 comments in one :see_no_evil:

docs/configuration.md
```diff
@@ -32,6 +32,7 @@ Caluma expects a bearer token to be passed on as [Authorization Request Header F
 * `OIDC_GROUPS_CLAIM`: Name of claim to be used to represent groups (default: caluma_groups)
 * `OIDC_USERNAME_CLAIM`: Name of claim to be used to represent the username (default: sub)
 * `OIDC_BEARER_TOKEN_REVALIDATION_TIME`: Time in seconds before bearer token validity is verified again. For best security token is validated on each request per default. It might be helpful though in case of slow Open ID Connect provider to cache it. It uses [cache](#cache) mechanism for memorizing userinfo result. Number has to be lower than access token expiration time. (default: 0)
+* `CALUMA_OIDC_USER_FACTORY`: User object factory (default: `caluma.caluma_user.models.OIDCUser`). Use it to provide a custom OIDC user object. The factory is expected to accept a mandatory  `token`  parameter as two optional parameters `userinfo` and `introspection`. Only one of them will be filled, depending on which OIDC endpoint the user information comes from
```
czosel, line 35: sorry, super-nitpick
```suggestion
* `CALUMA_OIDC_USER_FACTORY`: User object factory (default: `caluma.caluma_user.models.OIDCUser`). Use it to provide a custom OIDC user object. The factory is expected to accept a mandatory `token` parameter as two optional parameters `userinfo` and `introspection`. Only one of them will be filled, depending on which OIDC endpoint the user information comes from
```

docs/extending.md
```diff
@@ -361,3 +361,33 @@ will raise an exception.
...
+```
+
+The resulting object is expected to provide the following properties:
+* `username` - returns the username.
+* `group` - The primary group of the user. Used to fill in `created_by_group` etc
+* `is_authenticated` - Boolean, should always be `True`
+
+You may extend the object as you please, for example by providing
+shortcuts to data you need often  (reference to a Django user model if
+you use one, etc).
```
czosel, line 393: Excellent! :100:

docs/configuration.md
czosel: On second thought, might make more sense to just link to the docs just like in the emeis PR (to avoid duplication)?

docs/extending.md
```diff
@@ -361,3 +361,33 @@ will raise an exception.
...
+
+Especially when using Caluma embedded in another Django project, you need to
+ensure compatibility with other components. Caluma does not represent users
+as a database object, but instead uses "shim" object in it's role as an OIDC
+relying party (RP) that represents the requesting user.
+Oher projects may have differing requirements however.
+
+Therefore, you can define a custom `CALUMA_OIDC_USER_FACTORY`. The setting is a string
+that points to a callable, which is expected to return an OIDC user object.
+default it points to `caluma.caluma_user.models.OIDCUser`.
```
open-dynaMIX, line 375:
```suggestion
By default it points to `caluma.caluma_user.models.OIDCUser`.
```

docs/extending.md
```diff
@@ -361,3 +361,33 @@ will raise an exception.
 ## Caluma events
 
 For reacting upon Caluma Events, you can setup [event receivers](events.md)
+
+## OIDC User factory
+
+Especially when using Caluma embedded in another Django project, you need to
+ensure compatibility with other components. Caluma does not represent users
+as a database object, but instead uses "shim" object in it's role as an OIDC
```
open-dynaMIX, line 369:
```suggestion
as a database object, but instead uses a "shim" object in it's role as an OIDC
```

docs/configuration.md
```diff
@@ -160,4 +161,4 @@ outside (proxy configuration!). Due to it's nature, it takes some resources to
 check everything, which makes it prone to abuse (denial of service, DoS).
 
 It's enabled by setting the environment variable:
-`ENABLE_HEALTHZ_ENDPOINT`: Defaults to `false`.
\ No newline at end of file
+`ENABLE_HEALTHZ_ENDPOINT`: Defaults to `false`.
```
open-dynaMIX, line 164: Trailing new line change?

open-dynaMIX (approved):
Awesome work :100:

Just some nitpickts. Feel free to ignore...

PR #1552, opened 2021-08-05
fix(documentation): Updated readme to reflect that GraphiQL is not available anymore

StephanH90:
Also added set the DEV flag to true in the example `docker-compose.yml` file, which is something that I struggled with at the start.

However, that change is perhaps controversial.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,8 @@ services:
     #   - ./validations.py:/app/caluma/extensions/validations.py
     #   - ./data_sources.py:/app/caluma/extensions/data_sources.py
     #   - ./format_validators.py:/app/caluma/extensions/format_validators.py
-    environment:
+    environment
+      - ENV=dev
       - DATABASE_HOST=db
       # following options are a must to configure on production system:
       # https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SECRET_KEY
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -24,6 +24,7 @@ services:
     #   - ./data_sources.py:/app/caluma/extensions/data_sources.py
     #   - ./format_validators.py:/app/caluma/extensions/format_validators.py
     environment:
+      - ENV=dev
```
czosel, line 27: Should we add a comment explaining that this should be removed in production?

open-dynaMIX, line 27: This is already set [in the docker-compose.override.yml#L26](https://github.com/projectcaluma/caluma/blob/master/docker-compose.override.yml#L26). Value can be anything except `production`.

open-dynaMIX (changes requested):
Otherwise LGTM :100:

docker-compose.yml
czosel: Ah, right! @StephanH90 has the override file not been picked up in your case?

StephanH90: In that case I'm not sure why that didn't work. I remember that when I made a fresh install of docker and a clone of the repo I got a bunch of warning along the lines of: "secret key needs to be set", etc.

However, I suspect that this was a problem with my local setup.

open-dynaMIX (approved):

PR #1554, opened 2021-08-13
chore(models): add related name for dynamic option to document relation

anehx:


anehx:
This is not necessary since we already have `document.dynamic_option_set` as reverse relation. We'll close this since we don't want to cause any breaking changes in implementations that possibly use the mentioned reverse relation.

PR #1562, opened 2021-09-07
feat(signals): introduce event filter

winged:
The event filter decorator allows you to check against some predicate
functions before executing an event handler. This simplifies the handlers
by removing boilerplate code and allowing them to concentrate on the
work at hand.

### DIFF
--- a/caluma/caluma_core/events.py
+++ b/caluma/caluma_core/events.py
@@ -1,4 +1,6 @@
+import inspect
 import logging
+from functools import wraps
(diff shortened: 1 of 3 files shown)


anehx (approved):
Damn that was fast! Awesome! :100:

czosel (approved):
(only read the docs) nice!

open-dynaMIX (approved):
Awesome :tada:

PR #1566, opened 2021-09-15
Rebase onto master and cleanup xfaillist

fugal-dy:


fugal-dy:
we expect over time multiple people to work on the target branch so rebasing would be messing with their repos. to risky: closing and doing the work after merging master in PR#1567

PR #1568, opened 2021-09-27
chore: jwt token in tests should be bytes not str

open-dynaMIX:


winged (approved):
Looks good, but I'm not sure why the change is needed

open-dynaMIX:
> Looks good, but I'm not sure why the change is needed

I'm working in a project where I have to use the token in another request. In the tests the token is a string, while in reality it's bytes, so right now I have to handle this discrepancy in the tests.

Issue #1569, opened 2021-09-28
Analytics module


winged:
# Requirements

We need an analytics module that allows us to run queries, get summaries, and find insights into the data stored in Caluma.

Requirements are as follows:

* Analytics over multiple entity types possible
* Possibility to run aggregate functions over any part of a case - work item attributes, document answers, and so on
* Respect permission and visibility rules

# Implementation - Specification

## Architecture overview

The analytics module's functionality will be guided by the "Pivot Table" feature found in Microsoft Excel and LibreOffice Calc. There are two steps for a full analysis view:

1) Generate a table of data to be analyzed
2) Apply the pivot table mechanism on selected fields

The queries shall be stored as objects in the database, and be queried via GraphQL. This way, the module splits into a design time part and a runtime part, the same as the rest of Caluma.

### Table building

The first step in building an analysis is selecting the relevant data. A starting point is required to identify each row. From there, the data needed for the analysis can be selected.

For example, let's start with `Cases`. From there, all answers in the case's document are available for selection. To also be able to select answers from workitem (and sub-workitem!) documents, those should be available as well.

```python
class CaseStartingPoint(BaseStartingPoint):
    def get_fields(self):
        """Return a list of (slug, label) tuples that can be selected as fields.
        """
        fields = set([
            ("document.form", _("Case doc: form slug")),
            ("worflow", _("Workflow slug")),
        ])
        for case in Case.objects.all():
            # naive impl!
            for metakey in case.meta:
                fields.add((f'meta.{metakey}', _("Case: Meta key {}").format(metakey)))
            for question in case.document.form.questions.all():
                fields.add(
                    (f'document.{question.slug}', _("Case Doc: Question {}").format(question.label))
                )
                # Descend into subforms etc as needed
            # descend into work items, repeat
        return sorted(fields)
    
    def join_field(self, field):
        """Generate SQL to JOIN the given field"""
        # ...
```

For every such field, there must be a method to tells the system on how to `JOIN` the value, such that an efficient SQL query can be built.

The above example shows the basic structure on how this could look. 

Users should then be able to filter the resulting table. As every field contains a representable number of values, a simple list of values to show for each column should work fine: For example, one could request that only cases with a given workflow slug should be shown.

Of the selected fields, users should also be able to rename them (make aliases), such that they make more sense in the context they're being used. Some fields also may be hidden when the query is being run, as they may be used for filtering only.

Thus, a "table" specification model could look roughly like the following in the database:

[...]

PR #1570, opened 2021-09-29
feat(analysis): add path attribute / mixin to structural models

winged:
Part of a series of PRs that will lead to the analysis module: #1569

Add a denormalisation step to the structural data models (case, workitem,
and document) that will allow faster queries across case structures down
the road.


caluma/caluma_core/models.py
```diff
@@ -89,6 +91,62 @@ class Meta:
...
+        models.CharField(max_length=150),
+        default=list,
+        help_text="Stores a path to the given object",
+    )
+
+    def calculate_path(self, _seen_keys=None):
+        if not _seen_keys:
+            _seen_keys = set()
+        self_pk_list = [str(self.pk)]
+        if _seen_keys.intersection(set(self_pk_list)):
```
open-dynaMIX, line 118:
```suggestion
        if str(self.pk) in _seen_keys:
```

caluma/caluma_core/tests/test_set_path.py
```diff
@@ -0,0 +1,21 @@
+def test_set_paths(db, case, work_item_factory):
```
open-dynaMIX, line 1: Please put this test into the `caluma_workflow` app, because it depends on it as well as on the `caluma_form` app.

open-dynaMIX (changes requested):
Awesome work :tada:

Just minor nitpicks.

caluma/caluma_core/tests/test_set_path.py
winged: Done, ready for re-review

open-dynaMIX (approved):

PR #1577, opened 2021-10-28
fix(form): set allow_blank for static_content

### LINKED ISSUE #1576: static_content in staticQuestion is optional
The backend doesn't accept an empty string as `Static content`. On removal of this content, it needs to be set to `null`.

Steps to reproduce:

1. Save a question with `Static content`
2. Remove the `Static content`
3. Saving will now fail because the frontend sets `Static content` to `""`

Yelinz:
fixes #1576

### DIFF
--- a/caluma/caluma_form/serializers.py
+++ b/caluma/caluma_form/serializers.py
@@ -416,7 +416,7 @@ class SaveFileQuestionSerializer(SaveQuestionSerializer):
 
 
 class SaveStaticQuestionSerializer(SaveQuestionSerializer):
-    static_content = CharField(required=False)
+    static_content = CharField(required=False, allow_blank=True)
 
     def validate(self, data):
         data["type"] = models.Question.TYPE_STATIC


Yelinz:
If you want to use staticQuestion for an override component which doesnt display the text, for example a custom submit button in the form would be a staticQuestion without static_content

czosel:
Ah I see, so it's a bit of an edge case. Fine by me :+1:

Yelinz:
After some discussion it seems to be a problem of the `staticQuestion` schema, as `static_content` optional in the schema and therefore the frontend also marks it as an optional field. 
image
`static_content` should be required, otherwise the question would just be blank when its rendered which wouldn't make sense.

For the case of using a `staticQuestion` for an override component, a new question type of `widgetQuestion` should be created for that, instead of abusing a `staticQuestion`.

czosel:
Much cleaner! 👍

PR #1606, opened 2021-11-02
fix(schema): define interface for dynamic questions

anehx:


caluma/schema.py
```diff
@@ -47,6 +47,7 @@ class Mutation(form_schema.Mutation, workflow_schema.Mutation, graphene.ObjectTy
     form_schema.DateAnswer,
     form_schema.TableAnswer,
     form_schema.FileAnswer,
+    form_schema.CalculatedFloatQuestion,
```
winged, line 50: Nice catch this one!

winged (approved):
:tada:

PR #1607, opened 2021-11-02
feat(form): add new action button question type

anehx:
Add a new question type for action buttons that render a button for completing or skipping the work item related to the current document.


caluma/caluma_form/schema.py
```diff
@@ -82,6 +87,25 @@ class QuestionJexl(graphene.String):
 )
 
 
+WorkItemButtonActionType = graphene.Enum(
+    "WorkItemButtonActionType",
+    [(key.upper(), key) for key, desc in models.Question.ACTION_TYPE_CHOICES_TUPLE],
+)
+WorkItemButtonColor = graphene.Enum(
```
winged, line 94: Can we rename this to just `ButtonColor`? Even if it's currently only used on workitem buttons, it may be useful in other situations as well, so no need to have a over-specific name

caluma/caluma_form/models.py
```diff
@@ -167,6 +194,46 @@ def min_value(self):
...
+    @property
+    def validate_on_enter(self):
+        return self.configuration.get("validate_on_enter")
+
+    @validate_on_enter.setter
+    def validate_on_enter(self, value):
+        self.configuration["validate_on_enter"] = value
+
+    @property
+    def on_success_action_name(self):
```
winged, line 222: What is this used for? Is this not rather frontend-specific?

winged (approved):
Nice work! I have some questions/suggestions regarding naming, but otherwise looks good

winged (approved):
Nice! Thanks for the cleanup :tada:

PR #1613, opened 2021-11-08
fix(migrations): correct behaviour for path model migration

winged:


winged:
Close this - we keep that feature out of `main` for now (#1622)

PR #1615, opened 2021-11-11
feat(config): validate permission decorators

### LINKED ISSUE #1614: Permissions: mutation inheritance hierarchy does not match documentation
The permissions system assumes a hierarchy, where one can define permissions (or object permissions) for any given point in the hierarchy, and the most-specific one for any given actual mutation will be used.

For example, there is a `SaveDocumentAnswer` mutation, that will be used for all specific mutations, such as `SaveDocumentStringAnswer`.

However, the most base mutation, as per documentation, should be `caluma.schema.Mutation`. If you define a permission method for this one, it should be used if no other, more-specific permission method is found. This is not quite true: This mutation class is not the base of all other mutations, but a subtype of the "root mutations" of each module. It is used to consolidate all mutations from all django apps into a single interface.

The fix probably is relatively simple:

1. Rename the `class Mutation` in `caluma.schema`, and use that to consolidate the mutations entry points
2. Create a new `class Mutation` that is then subsequently used as a base class for all other mutations
3. Put that mutation class somewhere to resolve the ensuing circular dependencies :-)

winged:
When usig an invalid mutation class in the permission classes, things
might work unexpectedly, even though not actually being correct.

This validation checks that your permission classes actually
use a mutation that exists in the system.

fixes #1614

### DIFF
--- a/caluma/caluma_core/permissions.py
+++ b/caluma/caluma_core/permissions.py
@@ -5,8 +5,33 @@ from django.core.exceptions import ImproperlyConfigured
 from .collections import list_duplicates
 
 
+def _full_class_name(cls):
+    return f"{cls.__module__}.{cls.__name__}"
+
+
+def _collect_subclasses(cls):
+    yield cls
+    for sub in cls.__subclasses__():
+        yield from _collect_subclasses(sub)
+
+
+def _validate_mutation_type(mutation):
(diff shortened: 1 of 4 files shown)


open-dynaMIX (approved):
Awesome! :tada:

PR #1622, opened 2021-11-19
Revert "feat(analysis): add path attribute / mixin to structural models"

czosel:
This reverts commit 9f762f6. We might
need it in the future, but for now it is premature optimization.


winged (approved):
Yeah I think it's better to keep this unreleased and out of `main` for now. I'll add it once it's clear we actually need it

PR #1624, opened 2021-11-25
chore: push images to ghcr

open-dynaMIX:


winged (approved):
Nice, let's see if this works :)

PR #1628, opened 2021-11-29
Release 7.14

winged:
Contains some additional stuff (as discussed) next to the version bump:

* Builds on the `main` branch will push a tag named `dev`
* Builds on releases (tags) will push a tag named `latest`
  as well as a tag with the proper version number.

Thus, `latest` should always be the last-properly-released version,
and if you need tha latest development version, you can use the `dev`
docker tag instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+# v7.14.0
+
+### Feature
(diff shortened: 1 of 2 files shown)


open-dynaMIX (approved):
Awesome! :tada:

Very minor nitpick: The 2 commits should be in reversed order, right?

winged:
> Awesome! tada
> 
> Very minor nitpick: The 2 commits should be in reversed order, right?

Sorta would make sense, yes! Even more better would be to have two separate PRs :-)

winged:
@open-dynaMIX see #1629

open-dynaMIX (approved):

PR #1632, opened 2021-12-01
fix(settings): use int for MINIO_PRESIGNED_TTL_MINUTES

Yelinz:
timedelta throws an error when giving it an string


open-dynaMIX (approved):
Nice catch!

PR #1651, opened 2021-12-28
chore: add a code of conduct

czosel:


winged:
I'd like to see some reasoning behind why exactly *this* CoC (what were the alternatives? what are the differences?).

Also, similar to defining the license of a project, some broader discussion should take place IMHO, with all currently involved devs agreeing (not just one reviewer as with code changes)

czosel:
Thanks for the feedback @winged! You're right, this decision needs some more thought.

I didn't dive too deep into available CoC's, and went for what GitHub suggests for "projects of all sizes":
image

A potential alternative might be the [Django CoC](https://www.djangoproject.com/conduct/) - but I think I'd prefer the contributor covenant because it's name is not linked to Django and its widely used (see https://www.contributor-covenant.org/adopters/ - e.g. Linux and k8s use it as well).

Also relevant for us: (see https://www.contributor-covenant.org/)

> Before You Adopt Contributor Covenant
It’s important to understand that simply adopting Contributor Covenant will not prevent conflict or toxicity in your community.

> As a leader you are responsible for the safe, fair, and transparent enforcement of your community’s code of conduct. A code of conduct without such enforcement sends a false signal that a community is welcoming and inclusive, which can have a disastrous impact on marginalized or otherwise vulnerable people.

> Before you decide to adopt Contributor Covenant, take the time to discuss enforcement with other trusted members of your community. Thoroughly read the Enforcement Guidelines section of the document. If the suggested actions and consequences don’t suit the context of your community, consider moving this section to another document (such as an FAQ or ‘How to Report’ page) and adapting it to your needs.

Before merging this we should probably have a call about it.

winged (approved):
We now had some discussions in the team, and the consensus is to accept it. I agree with the consensus.

PR #1655, opened 2022-01-12
feat(form): enable hint messages on questions

luytena:
Add a hint text field on all question types except for
form, static and action button questions. This functionality
requires a db migration.


czosel (approved):
LGTM :+1: @winged or @open-dynaMIX, could you take a quick look as well?

caluma/caluma_form/serializers.py
```diff
@@ -243,7 +245,7 @@ def validate(self, data):
         return super().validate(data)
 
     class Meta(SaveQuestionSerializer.Meta):
-        fields = SaveQuestionSerializer.Meta.fields
+        fields = SaveQuestionSerializer.Meta.fields + ["hint_text"]
```
winged, line 248: Should probably be moved to `SaveQuestionSerializer.Meta`, so you won't have to explicitly extend the fields list in every subtype

winged (changes requested):
Code generally looks good except a small cleanup. However I'm wondering what is the use of this? What's the use case of hint vs. info texts?

caluma/caluma_form/serializers.py
winged: Ah, nevermind, needs to be this way because form, static and action buttons don't have hints

winged (approved):

winged:
Ah, ok, makes sense!

PR #1656, opened 2022-01-13
feat(workflow): implement WorkItem redo pattern

### LINKED ISSUE #1510: Implementing the `redo` pattern
The `redo` pattern (see http://workflowpatterns.com/patterns/resource/detour/wrp34.php) would allow already finished `workItems` to be reopened and change the course of action from this point. This would be mostly beneficial for `workItems` which are associated with a `childCase` or a `taskForm` since you want to edit/modify this specific data on the `workItem` and not create a new `workItem` with an empty data set or copy the previous generated data to the new one.

open-dynaMIX:
This commit implements a possibility to redo an already finished
WorkItem and all the ones that were finished in between.

Closes #1510


caluma/caluma_workflow/models.py
```diff
@@ -253,6 +260,27 @@ class WorkItem(UUIDModel):
         null=True,
     )
 
+    def get_redoable(self):
+        jexl = None
+        if self.previous_work_item:
```
winged, line 265: Why are we looking at the `previous_work_item`'s redoable field instead of "our own" here?

Also: if you'd invert the `if` condition, you could simplify the code by using an early return

caluma/caluma_workflow/domain_logic.py
```diff
@@ -528,3 +528,66 @@ def post_resume(work_item, user, context=None):
         )
 
         return work_item
+
+
+class RedoWorkItemLogic:
+    @classmethod
+    def validate_for_redo(cls, work_item):
+        if work_item.status == models.WorkItem.STATUS_READY:
```
winged, line 536: I think this needs to be documented. I was struggling with verifying the correctness here. Something along the lines of

> Validate if we can redo (jump back to) the given work item.
> Every task has a potential JEXL list of other tasks that can be redone from that task's workitem (redoable).
> We allow redoing a work item if at least one work item that came after this one allows "our" task to be re-done.

caluma/caluma_workflow/domain_logic.py
```diff
@@ -528,3 +528,66 @@ def post_resume(work_item, user, context=None):
...
+
+    @classmethod
+    def set_succeeding_work_item_status_redo(cls, work_item):
+        work_items = cls._find_ready_in_work_item_tree(work_item)
+        for wi in work_items:
+            wi.status = models.WorkItem.STATUS_REDO
+            wi.save()
+
+    @classmethod
+    def _find_ready_in_work_item_tree(cls, work_item, allowed_states=None, ready=None):
```
winged, line 557: If you'd convert this method into a generator, you'll improve two thing: The code gets simpler as you can use `yield from` instead of building up a list and passing it along. And also, it could improve performance, as we won't have to walk the tree further than needed once a workitem is found that allows the redo

caluma/caluma_workflow/domain_logic.py
```diff
@@ -528,3 +528,66 @@ def post_resume(work_item, user, context=None):
...
+    def validate_for_redo(cls, work_item):
+        if work_item.status == models.WorkItem.STATUS_READY:
+            raise ValidationError("Ready work items can't be redone.")
+        ready_work_items = cls._find_ready_in_work_item_tree(
+            work_item, allowed_states=[models.WorkItem.STATUS_READY]
+        )
+        allowed = False
+        for wi in ready_work_items:
+            if work_item.task in wi.get_redoable():
+                allowed = True
```
winged, line 544: You don't need the `allowed` variable. Just return early in case we are allowed to execute the `redo`. Then the  `ValidationError` below can be thrown without `if`

winged (changes requested):
Code looks great! However I do have a couple of issues that need to be addressed.

Additionally to the things noted in the code, I feel like this feature needs some more documentation. In the Markdowns as well as in the code

caluma/caluma_workflow/models.py
open-dynaMIX: > Why are we looking at the previous_work_item's redoable field instead of "our own" here?

Because there we have the `redoable` field. If we would set it on our own task_flow, we wouldn't be able to redo a workitem if we reached one without a task_flow (the last one in a branch).

> Also: if you'd invert the if condition, you could simplify the code by using an early return

Done

caluma/caluma_workflow/domain_logic.py
open-dynaMIX: Done, but it's not exactly as u wrote.

```
        Validate if we can redo (jump back to) the given work item.

        Every taskflow has a potential JEXL list of other tasks that can be redone from
        that task's next workitem (redoable).
        We allow redoing a work item if at least one ready work item allows "our" task
        to be re-done.
```

caluma/caluma_workflow/domain_logic.py
open-dynaMIX: Nice, done.

caluma/caluma_workflow/domain_logic.py
open-dynaMIX: True, done.

caluma/caluma_workflow/models.py
winged: > > Why are we looking at the previous_work_item's redoable field instead of "our own" here?
> 
> Because there we have the `redoable` field. If we would set it on our own task_flow, we wouldn't be able to redo a workitem if we reached one without a task_flow (the last one in a branch).

Aah, that makes sense. However, I'm really not too happy about that. I think this really should be directly on the affected work item ('s task flow) instead. The mental indirection required sounds rather tricky IMO. Consider that a TaskFlow tells us "where to go from here" (`flow.next`), I would strongly expect the `redo` to tellus "where to go back from here", not "where to go back from the *next* item".

This unfortunately sounds like a bit more work, but I think we should store the "redo" data local to where it's relevant, not "on the previous task's task flow".

One solution I think would be acceptable for this is to make the `flow` foreign key nullable, so we can have a task flow that doesn't lead anywhere but may allow redoing. In the long term, as discussed, the `Flow` may even be dropped alltogether

open-dynaMIX: Done

winged (approved):
Awesome, I like it much better this way :tada: 

We're still lacking documentation, but as we discussed in another place, we should overhaul it anyway, and we can add that bit then.

PR #1659, opened 2022-01-17
feat(workflow): Conditional workItem creation: if not exists

### LINKED ISSUE #1512: Conditional `workItem` creation: if not exists
If multiple `flows` point to a `task` and this `task` is not a multiinstance `task` there should no `workItem` be created when already one of this `task` exists.

open-dynaMIX:
Only create a new workitem, if it doesn't exist already.

Closes #1512


winged (changes requested):
Nice work! Code and test look good! However I'd like to have some documentation / comments / commit message to explain the change or the new behaviour, respectively.

PR #1665, opened 2022-02-03
Fix validators

anehx:
Related to https://github.com/projectcaluma/ember-caluma/pull/1779

### DIFF
--- a/caluma/caluma_form/format_validators.py
+++ b/caluma/caluma_form/format_validators.py
@@ -39,23 +39,23 @@ class BaseFormatValidator:
 
 class EMailFormatValidator(BaseFormatValidator):
     slug = "email"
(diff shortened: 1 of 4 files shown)


czosel (approved):
Classic case of frontend / backend communication breakdown :see_no_evil:

winged (approved):
Nice catch! I'm wondering whether we really need the `reverse()` migration, but otherwise looks good!

anehx:
> Nice catch! I'm wondering whether we really need the `reverse()` migration, but otherwise looks good!

I added it since it might happen that someone upgrades the backend without upgrading the frontend which would cause the validation to be broken in the frontend. This way it can at least be reversed if this would happen.

PR #1675, opened 2022-02-15
fix(form): update or create answers when copying to a new document

anehx:
Currently `document.copy` fails when a calculated question is configured in the form because the signal already creates an answer for that question.


winged (approved):
Nice catch!

czosel (approved):

PR #1677, opened 2022-02-16
feat(filter): add lookup type "IN"

czosel:
Two things I'm still wondering:

- `test_query_all_questions` is running a truckload of tests, but the parametrization doesn't make sense to me
-  why do we allow the same lookup types for single choice than for multiple choice? How does e.g. INTERSECTS make sense for single choice questions?

### DIFF
--- a/caluma/caluma_form/tests/test_filter_by_answer.py
+++ b/caluma/caluma_form/tests/test_filter_by_answer.py
@@ -140,6 +140,9 @@ def test_query_all_questions(
         }
     """
 
+    if search_lookup == "IN":
+        search_value = [search_value]
+
     variables = {"hasAnswer": [{"question": search_slug, "value": search_value}]}
     if search_lookup is not None:
         variables["hasAnswer"][0]["lookup"] = search_lookup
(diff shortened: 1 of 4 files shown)


caluma/caluma_form/tests/test_filter_by_answer.py
```diff
@@ -157,6 +157,43 @@ def test_query_all_questions(
...
+def test_has_answer_in(
+    schema_executor, db, question, answer_factory, values, search, expect_count
+):
+
+    for value in values:
+        answer_factory(question=question, value=value)
+
+    query = """
+        query asdf ($hasAnswer: [HasAnswerFilterType]!) {
+          allDocuments(hasAnswer: $hasAnswer) {
```
winged, line 178: Please use new-style filter syntax - this one will be removed soon.
```suggestion
          allDocuments(filter: [{hasAnswer: $hasAnswer}]) {
```

winged (changes requested):
Nice work, just a minor nitpick for the tests

winged (approved):
Awesome work! :tada:

winged:
Looks like it needs a snapshot update, but otherwise all good!

czosel:
Fixed

PR #1680, opened 2022-02-17
fix: restrict valid filters for choice questions

czosel:
Based on #1677


czosel:
True! Let's try and remember to list this as breaking in the release notes.

PR #1705, opened 2022-03-07
feat(workflow): allow multiple instance work items to continue async

anehx:


caluma/caluma_workflow/migrations/0029_task_continue_async.py
```diff
@@ -0,0 +1,40 @@
...
+            model_name="task",
+            name="continue_async",
+            field=models.BooleanField(
+                blank=True,
+                default=None,
+                help_text="Whether to continue the flow if the multiple instance work item has ready siblings",
+                null=True,
+            ),
+        ),
+        migrations.RunPython(set_default_continue_async, migrations.RunPython.noop),
```
winged, line 39: This should probably have a follow-up step where the `null=True` is removed in favour of `default=False`, no? Otherwise, we wouldn't have the need to set it explicitly to `False` here anyway.

If I remember correctly, this might require an additional migration file to avoid overzealous optimisation by Django.

winged (changes requested):
Nice work! I think the model should make the new option non-nullable (See inline comment), but otherwise LGTM!

caluma/caluma_workflow/migrations/0029_task_continue_async.py
anehx: No, `null=True` is intended since a default to `False` would be missleading for tasks that are not multiple instance. To ensure the default behaviour it's explicitly set to `False` for multiple instance tasks.

winged: Not sure I'm buying it.. the behaviour would be exactly the same (The `not task.continue_async` check), it's just not having an effect in non-multiple-instance tasks.

I'm also not sure it would really be misleading. An explanatory comment would probably be a good idea in either case though

anehx: Sure, the behaviour there would be the same. However I think that having a value set to `False` when it is not relevant for that object is less clean than setting it to `None`. I don't really care enough about this to discuss this any further so I'll change it.

winged (approved):

PR #1706, opened 2022-03-07
feat(workflow): only create work item if they don't exist yet

### LINKED ISSUE #1512: Conditional `workItem` creation: if not exists

anehx:
Work items should only be created if they don't exist yet or if the task allows multiple instances. This makes sure that we don't have accidental duplicate of work items. This intentionally doesn't use `get_or_create` to avoid the creation of an obsolete documents since django doesn't support lazy defaults.

Closes #1512

*Disclaimer: This is basically copy paste of #1659 with a refactoring to use a preliminary query instead of `get_or_create` to avoid obsolete documents and deadline computations.*

Closes #1659


winged (approved):
Nice! :tada:

PR #1708, opened 2022-03-10
fix(graphene): avoid default connection limit

anehx:


caluma/caluma_core/types.py
```diff
@@ -66,44 +66,60 @@ class DjangoConnectionField(DjangoConnectionField):
...
-    TODO: properly implement max_limit, see
-    https://github.com/graphql-python/graphene-django/blob/b552dcac24364d3ef824f865ba419c74605942b2/graphene_django/fields.py#L133
     """
 
     @classmethod
     def resolve_connection(cls, connection, args, iterable, max_limit=None):
+        # Remove the offset parameter and convert it to an after cursor.
+        offset = args.pop("offset", None)
+        after = args.get("after")
+        if offset:  # pragma: no cover
```
winged, line 76: I'd really like to see a test for this - it's a bit too complex to just "no cover" it IMHO

caluma/caluma_core/types.py
```diff
@@ -66,44 +66,60 @@ class DjangoConnectionField(DjangoConnectionField):
...
         # would try to do a negative slicing which makes django throw an
         # AssertionError
-        after = min(get_offset_with_default(args.get("after"), -1) + 1, _len)
-        if max_limit is not None and "first" not in args:  # pragma: no cover
-            args["first"] = max_limit
+        after = min(get_offset_with_default(args.get("after"), -1) + 1, list_length)
+
+        if (
+            max_limit is not None and args.get("first", None) is None
+        ):  # pragma: no cover
```
winged, line 105: Same here - this is nontrivial, so I'd also like to see a test (or, if the code is never ever really used, could probably be dropped or an exception raised)

winged (changes requested):
Nice catch! Looks good, but some corner cases that are implemented should in my opinion be covered by tests explicitly, as they're non-trivial

anehx:
@winged why would we test code that's 100% copy pasted from a framework that already tests it?

caluma/caluma_core/types.py
anehx: Done

caluma/caluma_core/types.py
anehx: This is now explicitly commented out since we disable it in our settings and it was impossible to test.

czosel (approved):
:100:

PR #1713, opened 2022-03-14
Feat: reopen case

StephanH90:
# Description
Adds the ability to reopen a previously `completed` case.

During re-opening of a case the ID of the case as well as an array of `work-items` have to be supplied.

### Case reopen validation:
- Case status has to be `STATUS_COMPLETED` or `STATUS_CANCELED`. Suspended cases can not be reopend.
- A case can not be a child case
### Work item validation:
- Only work items at the end of a branch can be re-opened
- Only work items belonging to the case that is being re-opened can be re-opened

### DIFF
--- a/caluma/caluma_workflow/events.py
+++ b/caluma/caluma_workflow/events.py
@@ -8,7 +8,7 @@ module = sys.modules[__name__]
 
 MODEL_ACTIONS = {
     "work_item": ["create", "complete", "cancel", "skip", "suspend", "resume", "redo"],
-    "case": ["create", "complete", "cancel", "suspend", "resume"],
+    "case": ["create", "complete", "cancel", "suspend", "resume", "reopen"],
 }
 
 # create "pre|post"_"create|complete|..."_"work_item|case" events in all given combinations
--- a/caluma/caluma_workflow/tests/test_reopen_case.py
+++ b/caluma/caluma_workflow/tests/test_reopen_case.py
@@ -1,0 +1,202 @@
+import pytest
+from django.core.exceptions import ValidationError
+from graphene_django.utils.str_converters import to_const
+
+from .. import api, models
+
+
+@pytest.fixture
+def reopen_case_setup(case_factory, work_item_factory):
+    def case_work_item_setup(case_status, work_item_statuses):
+        case = case_factory(status=case_status)
+        work_items = []
+
+        for work_item_status in work_item_statuses:
+            work_items.append(work_item_factory(case=case, status=work_item_status))
+
+        return case, work_items
+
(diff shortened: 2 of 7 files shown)


caluma/caluma_workflow/events.py
```diff
@@ -48,6 +49,7 @@
     "post_create_case": ["created_case"],
     "post_suspend_case": ["suspended_case"],
     "post_resume_case": ["resumed_case"],
+    "post_reopen_case": ["reopened_case"],
```
winged, line 52: That's a new signal, no need to deprecate it

caluma/caluma_workflow/events.py
```diff
@@ -29,6 +29,7 @@
 created_case = Signal()
 suspended_case = Signal()
 resumed_case = Signal()
+reopened_case = Signal()
```
winged, line 32: Not needed, that's the "old" way of doing it. No need to introduce a deprecated signal :)

caluma/caluma_workflow/tests/test_reopen_case.py
```diff
@@ -0,0 +1,214 @@
+import pytest
+from django.core.exceptions import ValidationError
+from graphene_django.utils.str_converters import to_const
+
+from .. import api, models
+
+
+def reopen_case_setup(case_status, work_item_statuses, case_factory, work_item_factory):
```
winged, line 8: You can turn this into a fixture using the `@pytest.fixture` decorator, returning a function to do the actual work. This would save you some typing on the call sites.

caluma/caluma_workflow/tests/test_reopen_case.py
```diff
@@ -0,0 +1,214 @@
...
+@pytest.mark.parametrize("use_graphql", [False, True])
+def test_allow_completed_cases_to_be_reopened(
+    db, admin_user, case_factory, work_item_factory, schema_executor, use_graphql
+):
+    completed_case, completed_work_items = reopen_case_setup(
+        models.Case.STATUS_COMPLETED,
+        [models.WorkItem.STATUS_COMPLETED],
+        case_factory,
+        work_item_factory,
+    )
```
winged, line 27: This test has exactly one statement common between the parametrisations, everything else is in separate code paths. Therefore I suggest splitting this test into two (once with graphql, once with api)

winged (changes requested):
Awesome work! Some things need cleaning up / refactoring, but we're getting there..

winged:
Looking good now! I think after cleaning up the commits, we can merge this

winged (approved):

PR #1687, opened 2022-02-24
Feat: analytics pivot table

winged:
The pivot table implements aggregate queries for the analytics module. It
allows taking data from the SimpleTable, and aggregating and grouping
the result.

The implementation deviates a bit from the spec: We originally intended to
use different fields for the "grouping" and "aggregating" parts. However
it was realized that those differ in fact only in the way they're
displayed, not in the way they need to be defined in the GQL level. So
now, analytics tables only have one sort of fields instead of two.

Instead of having "row fields" and "data fields", the aggregate fields
now have a pseudo-function "GROUP" which marks a field as "GROUP BY"
instead of an aggregate function. Those will then need to be rendered
as rows in the (frontend) output.

PR #1719, opened 2022-03-18
Feat: Unified pivot analytics table

winged:
The pivot table implements aggregate queries for the analytics module. It
allows taking data from the SimpleTable, and aggregating and grouping
the result.

The implementation deviates a bit from the spec: We originally intended to
use different fields for the "grouping" and "aggregating" parts. However
it was realized that those differ in fact only in the way they're
displayed, not in the way they need to be defined in the GQL level. So
now, analytics tables only have one type of fields instead of two.

Instead of having "row fields" and "data fields", the aggregate fields
now have a pseudo-function "GROUP" which marks a field as "GROUP BY"
instead of an aggregate function. Those will then need to be rendered
as rows in the (frontend) output.

Also, instead of having distinct "extraction" and "pivot" tables that
depend on each other, we now only have one type of table. If you choose
to apply aggregate functions to a field, it will be a pivot table, otherwise
it will be a plain "extraction" table.

This is a "simpler" variant but functoinally equal to #1687


caluma/caluma_analytics/conftest.py
```diff
@@ -0,0 +1,98 @@
...
+            ),
+        )
+        for day, status in enumerate(statuses, start=1)
+    ]
+
+
+@pytest.fixture
+def example_pivot_table(example_analytics):
+
+    status_field = example_analytics.fields.get(alias="statuuuuuus")
```
anehx, line 79: Is this a typo?

caluma/caluma_analytics/models.py
```diff
@@ -15,13 +15,66 @@ class AnalyticsTable(SlugModel):
 
     disable_visibilities = models.BooleanField(default=False)
     name = LocalizedField(blank=False, null=False, required=False)
-    starting_object = models.CharField(max_length=250, choices=STARTING_OBJECT_CHOICES)
+
+    starting_object = models.CharField(
+        max_length=250, choices=STARTING_OBJECT_CHOICES, null=False
```
anehx, line 20:
```suggestion
        max_length=250, choices=STARTING_OBJECT_CHOICES
```

`null=False` is the default

caluma/caluma_analytics/models.py
```diff
@@ -15,13 +15,66 @@ class AnalyticsTable(SlugModel):
...
+    )
+
+    def is_extraction(self):
+        return all(
+            [
+                field.function == AnalyticsField.FUNCTION_VALUE
+                for field in self.fields.all()
+            ]
+        )
```
anehx, line 29: This is to avoid db hits, right?

caluma/caluma_analytics/schema.py
```diff
@@ -115,19 +124,41 @@ def resolve_available_fields(obj, info, prefix=None, depth=None, **kwargs):
...
 
+    result_data = graphene.Field(AnalyticsOutput)
+
     @staticmethod
     def resolve_result_data(obj, info):
-        return simple_table.SimpleTable(obj)
+        if obj.is_extraction():
+            return simple_table.SimpleTable(obj)
+        else:
+            return pivot_table.PivotTable(obj)
```
anehx, line 134:
```suggestion
        if obj.is_extraction():
            return simple_table.SimpleTable(obj)

        return pivot_table.PivotTable(obj)
```

Not sure if this is a thing in python but in our frontend code we have a rule to avoid else returns

caluma/caluma_analytics/serializers.py
```diff
@@ -10,9 +10,19 @@ def __init__(self, **kwargs):
...
+
+
 class SaveAnalyticsTableSerializer(serializers.ModelSerializer):
     starting_object = StartingObjectField(required=True)
 
+    def validate(self, data):
+        validated_data = super().validate(data)
+        # TODO remove if we don't have anything to validate
+        return validated_data
+
```
anehx, line 25: Do what the comment says :wink:

caluma/caluma_analytics/serializers.py
```diff
@@ -67,24 +79,6 @@ def _validate_field_exists(self, data):
...
-            )
-        if existing_fields.filter(data_source=data["data_source"]).exists():
-            raise exceptions.ValidationError(
-                {
-                    "dataSource": [
-                        "Cannot use the same data source twice within the same table"
-                    ]
-                }
-            )
-
```
anehx, line 87: Why isn't this necessary anymore?

caluma/caluma_analytics/serializers.py
```diff
@@ -106,6 +101,11 @@ class Meta:
 class RemoveAnalyticsFieldSerializer(serializers.ModelSerializer):
     id = serializers.GlobalIDField()
 
+    def validate(self, data):
+        validated_data = super().validate(data)
+        # TODO remove if we don't have anything to validate
+        return validated_data
+
```
anehx, line 108: Remove

caluma/caluma_analytics/simple_table.py
```diff
@@ -86,9 +87,23 @@ def source_path(self) -> List[str]:
...
-        Only needs to be overloaded when required.
-        Used in GraphQL context only
+        Ensure date / time types are using the correct timezone,
+        and numeric types are int or float.
         """
+        if isinstance(value, Decimal):
+            # Postgres sometimes returns Decimal types that
+            # need to be converted for "proper" representation
+            # in the output
+            return float(value)  # pragma: no cover
```
anehx, line 97: This seems like something that should be covered..

caluma/caluma_analytics/sql.py
```diff
@@ -85,7 +87,7 @@ def self_alias(self):
         this query.
         """
         if isinstance(self.from_, str):
-            return self.from_
+            return self.from_  # pragma: no cover
```
anehx, line 90: Why isn't this covered anymore?

caluma/caluma_analytics/sql.py
```diff
@@ -395,7 +412,7 @@ def _join_source(self, from_spec):
         if isinstance(from_spec, str):
             # by defintion, only "our own" table can be passed as string.
             # subqueries MUST be query objects
-            return (from_spec, {}, self.query.outer_alias())
+            return (from_spec, {}, self.query.outer_alias())  # pragma: no cover
```
anehx, line 415: Why isn't this covered anymore?

anehx (changes requested):
Awesome work as far as I can tell! Just some minor comments and questions..

Otherwise LGTM

caluma/caluma_analytics/serializers.py
winged: In other parts of Caluma, we let the ORM do such uniqueness checks, which has the same result, just with a bit less-nice of an error message. But since end-users don't see this either way, it's not needed

caluma/caluma_analytics/sql.py
winged: This code path was used when referencing the DB table table directly, but now we're always going via the queryset from the visibilities, so `self.from_` is always a queryset/subquery. I'd like to keep it in for now, but might decide later to remove this completely.

caluma/caluma_analytics/simple_table.py
winged: Sadly, this depends on the postgres version - so this code path is not covered in all variants we test in CI at the moment

caluma/caluma_analytics/conftest.py
anehx: Fair enough

caluma/caluma_analytics/models.py
```diff
@@ -15,13 +15,65 @@ class AnalyticsTable(SlugModel):
...
+
     starting_object = models.CharField(max_length=250, choices=STARTING_OBJECT_CHOICES)
 
+    def is_extraction(self):
+        # We are an extraction table if there are ONLY
+        # `FUNCTION_VALUE` type fields.
+        return (
+            not self.fields.all()
+            .exclude(function=AnalyticsField.FUNCTION_VALUE)
+            .exists()
```
anehx, line 27: `.all()` is not necessary

caluma/caluma_analytics/serializers.py
anehx: :+1:

caluma/caluma_analytics/simple_table.py
anehx: Damn, makes sense. Maybe change the comment so this is clear

caluma/caluma_analytics/sql.py
anehx: Add a comment for that

caluma/caluma_analytics/sql.py
winged: same reason as above. Comment added

caluma/caluma_analytics/sql.py
winged: Done

anehx (approved):

fugal-dy (approved):
Nice!

Nothing much to add from my side. Looks excellent

caluma/caluma_analytics/sql.py
```diff
@@ -347,6 +352,12 @@ def _order_by(self):
             return f"ORDER BY {order_sql}"
         return ""
 
+    def _group_by(self):
+        if self.query.group_by:
+            group_sql = ",\n       ".join(self.query.group_by)
```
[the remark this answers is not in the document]
winged: Nope, the 7 spaces are just to align with everything else, particularly below the `SELECT` statement. select = 6 chars, plus one space :-)

caluma/caluma_analytics/simple_table.py
```diff
@@ -86,9 +88,24 @@ def source_path(self) -> List[str]:
...
+            # Postgres sometimes returns Decimal types that
+            # need to be converted for "proper" representation
+            # in the output. Depends on PostgreSQL version
+            # whether this is triggered or not
+            return float(value)  # pragma: no cover
+
+        if isinstance(value, datetime):
+            # psycopg2 returns "generic" tz info dates that sometimes
+            # are weird when comparing to the ones Django expects
+            # (Note they're still correct, just not labeled in an
```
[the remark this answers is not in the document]
winged: Fixed

caluma/caluma_analytics/sql.py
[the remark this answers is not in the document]
fugal-dy: ahh okay, I go all like: how does that not factor by four, is that even a real indent :rofl:

PR #1725, opened 2022-03-21
fix(document): remove duplicate filtering on answer

anehx:
This caused some kind of query loop that resulted in very slow DB queries.

### DIFF
--- a/caluma/caluma_form/models.py
+++ b/caluma/caluma_form/models.py
@@ -4,7 +4,6 @@ from functools import wraps
 from django.contrib.postgres.fields import ArrayField
 from django.contrib.postgres.indexes import GinIndex
 from django.db import models, transaction
-from django.db.models import Q
 from django.utils.functional import cached_property
 from localized_fields.fields import LocalizedField, LocalizedTextField
 from minio import S3Error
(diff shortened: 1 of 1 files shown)


caluma/caluma_form/models.py
```diff
@@ -350,7 +349,7 @@ def last_modified_answer(self):
         """
 
         return (
-            Answer.objects.filter(Q(document=self) | Q(document__family=self))
+            Answer.objects.filter(document__family=self)
```
czosel, line 352: I think this breaks the behavior for table rows - see comment above: 

> For root documents, we want to look through the whole document, **while for row documents only the table-local answer is wanted.**

winged: @anehx Potentially, we could solve this by doing `Q(document__id=self) | Q(document__family=self)`, but I'm afraid Django is too smart and will generate the same SQL.

@czosel How often do we need this feature anyway? Exactly in this form? Might be worth the sacrifice, given the performance penalty we pay for that

winged (approved):
Nice work! Still don't know what causes PostgreSQL to slow down THAT much in this query...

PR #1726, opened 2022-03-22
fix(graphene): crash in allForms query

JohnL17:


caluma/caluma_form/tests/test_form.py
```diff
@@ -63,6 +63,52 @@ def test_query_all_forms(
...
+                }
+              }
+            }
+          }
+        }
+    """
+
+    result = schema_executor(
+        query,
+    )
```
winged, line 106: You're not passing the `$slug` input variable here - is this intentional?

This causes the `slug` filter to be disregarded, causing the error.

winged: Added a check in the filter code to avoid crash

caluma/caluma_form/tests/test_form.py
```diff
@@ -63,6 +63,52 @@ def test_query_all_forms(
...
+                name
+                meta
+                isPublished
+                source {
+                  slug
+                }
+              }
+            }
+          }
+        }
```
anehx, line 101: Let's keep the test simple:

```suggestion
        query Form($slug: String) {
          allForms(filter: [{ slug: $slug }]) {
            totalCount
          }
        }
```

caluma/caluma_form/tests/test_form.py
```diff
@@ -63,6 +63,52 @@ def test_query_all_forms(
...
+    form,
+    form_factory,
+    form_question,
+    form_question_factory,
+    question,
+    schema_executor,
+):
+    form_factory(name="3rd", description="Second result")
+    form_factory(name="2nd", description="Second result")
+    form_question_factory(form=form)
```
anehx, line 78:
```suggestion
    form_factory.create_batch(5)
```

caluma/caluma_form/tests/test_form.py
```diff
@@ -63,6 +63,52 @@ def test_query_all_forms(
...
+def test_query_root_forms(
+    db,
+    snapshot,
+    form,
+    form_factory,
+    form_question,
+    form_question_factory,
+    question,
+    schema_executor,
+):
```
anehx, line 75:
```suggestion
def test_optional_filter(db, form_factory, question, schema_executor):
```

caluma/caluma_form/tests/test_form.py
```diff
@@ -63,6 +63,52 @@ def test_query_all_forms(
...
+          }
+        }
+    """
+
+    result = schema_executor(
+        query,
+    )
+
+    assert not result.errors
+    snapshot.assert_match(result.data)
```
anehx, line 109:
```suggestion
    assert result.data["allForms"]["totalCount"] == 5
```

anehx (changes requested):
Let's keep the test generic and simple..

winged:
@anehx Thanks for the review, applied your suggestions

anehx (approved):

PR #1730, opened 2022-03-24
Feat: analytics summary

winged:
Add a summary output to pivot tables. This gives the same aggregates,
but without grouping, thus applying the aggregates over all results.

Note: Stacks on top of #1719 and must be rebased before merge.

### DIFF
--- a/caluma/tests/__snapshots__/test_schema.ambr
+++ b/caluma/tests/__snapshots__/test_schema.ambr
@@ -153,6 +153,7 @@
   
   type AnalyticsOutput {
     records(before: String = null, after: String = null, first: Int = null, last: Int = null): AnalyticsTableContentConnection
+    summary: AnalyticsRowConnection
   }
   
   type AnalyticsRowConnection {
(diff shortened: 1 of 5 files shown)


caluma/tests/__snapshots__/test_schema.ambr
```diff
@@ -153,6 +153,7 @@
   
   type AnalyticsOutput {
     records(before: String = null, after: String = null, first: Int = null, last: Int = null): AnalyticsTableContentConnection
+    summary(before: String = null, after: String = null, first: Int = null, last: Int = null): AnalyticsRowConnection
```
anehx, line 156: More of a general question: We extend the standard connection types in analytics which means we don't have a total count and I'm not sure that the standard implementation of the pagination (especially the `PageInfo`) works properly - is that intended?

winged: For now, we don't do pagination or total count output, correct. Will add this later on if needed

anehx: If we don't do pagination, why are we exposing pagination params? Either implement it or omit it..

anehx (approved):

PR #1736, opened 2022-03-31
feat(analytics): support for additional starting objects

winged:
We now also support:
  * Work items and everything "below"
  * Documents and their answers


caluma/caluma_analytics/simple_table.py
```diff
@@ -740,6 +754,116 @@ def root_fields(self):
         return {**direct_fields, **form_fields, **workitem_fields}
 
 
+class WorkItemsStartingObject(BaseStartingObject):
+    identifier = "workitems"
```
anehx, line 758:
```suggestion
    identifier = "work_items"
```

anehx (changes requested):
Minor nitpicks, apart from that LGTM as far as I can tell!

winged:
Note: Contains #1737 which should be reviewed first

caluma/caluma_analytics/simple_table.py
```diff
@@ -740,6 +754,116 @@ def root_fields(self):
         return {**direct_fields, **form_fields, **workitem_fields}
 
 
+class WorkItemsStartingObject(BaseStartingObject):
+    identifier = "work_items"
+    label = "Work Items"
+
+    def get_query(self):
+        return self.visibility_source.workitems()
```
anehx, line 762: Can we rename this to `.work_items()` as well to be consistent?

winged: Done

caluma/caluma_analytics/simple_table.py
```diff
@@ -740,6 +754,116 @@ def root_fields(self):
...
+            ),
+            "meta": MetaField(
+                parent=None,
+                identifier="meta",
+                label="Meta",
+                visibility_source=self.visibility_source,
+            ),
+            "form_id": AttributeField(
+                # TODO can we use "form" as identifier and "form_id" as actual field?
+                # Alternative - join form and present all of the attributes it has, in turn?
```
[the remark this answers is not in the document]
winged: Thought a bit about this - Leaving it with "task_id" / "form_id", so we can add a "Join field" later on with the name "task" or "form" respectively, leading to datasource names like "task.slug" or "task.meta.foo" or similar, without breaking stuff.

anehx (approved):

PR #1737, opened 2022-04-01
feat(test): fixture for running a block of code at a given faked time

winged:
Similar to the existing at_date() fixture, the new set_date() fixture
works as a context manager and allows more complex code to run within
a given faked datetime.

### DIFF
--- a/caluma/conftest.py
+++ b/caluma/conftest.py
@@ -1,3 +1,4 @@
+import contextlib
 import datetime
 import functools
 import inspect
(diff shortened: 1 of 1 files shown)


caluma/conftest.py
```diff
@@ -377,8 +378,37 @@ def custom_snapshot(*args):
...
+
+            with set_date('2022-01-01'):
+                assert datetime.now() == datetime(2022,1,1)
+
+            assert datetime.now() == datetime(2022,2,22)
+    """
+
+    @contextlib.contextmanager
+    def make_context(date):
+        old_now = datetime.datetime.now()
```
anehx, line 401: This might cause a runtime warning because it's a naive date, no?

winged: No, those dates are not used in Django`DateTimeField`s. It's only used to set a fake time during testing

anehx: :+1:

anehx (approved):

PR #1739, opened 2022-04-01
test: allow minio cert checks to fail if badssl.com is down

winged:
We should avoid having this external dependency, but for now, it
servers it's purpose (unless it's down, as of right now).

Since the test is not verifying an overly critical behaviour, we
allow it to fail if the external dependency is not available.

See https://github.com/chromium/badssl.com/issues/500

### DIFF
--- a/caluma/caluma_form/tests/test_minio.py
+++ b/caluma/caluma_form/tests/test_minio.py
@@ -1,14 +1,24 @@
 import minio.error
 import pytest
+import requests
 import urllib3
 
 from .. import storage_clients
+
+
+def is_badssl_down():
+    """Return True if badssl.com is not reachable or down."""
+    try:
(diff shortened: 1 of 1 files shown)


anehx (approved):
Wow, this is sexy :dancers: Didn't know that xfail takes a function as condition, nice!

anehx (changes requested):
However, the scope of the commit message shouldn't be `fix` as it doesn't fix anything for the users. Let's use `test(minio): ...`

winged:
> However, the scope of the commit message shouldn't be `fix` as it doesn't fix anything for the users. Let's use `test(minio): ...`

Done, had some coverage issues in the check still...

anehx (approved):

PR #1740, opened 2022-04-01
chore: remove unused fixture

winged:
I still kind of like it, but it's currently unused, and just keeping stuff
around because it "might" be used later on is kind of a bad idea


anehx (approved):

winged:
This will be merged as part of #1736 as otherwise I'd need to mess around quite a bit with `# pragma: no cover` lines

PR #1752, opened 2022-04-19
fix(deps): lock graphql-core to 3.1.x to avoid import error

anehx:
`graphene-django` uses a removed function `format_error` from
`graphql-core` but accepts the version in which it was removed as a
valid dependency. This commit makes sure we only use working versions of
`graphql-core`.

https://github.com/graphql-python/graphql-core/commit/09ff14f68bdce1e9cd71decc871fd38274b01971


winged (approved):
LGTM

Issue #1489, opened 2021-04-30
Clean up the connection fields


winged:
A rather long while ago, we decided to put all the filters into a single `filter` type, per connection. We should now remove the old filter types on the connections, so that we have only the new syntax available.
This should be easily implemented by removing the parent class `filterset_class` from the generated filterset collections in the `CollectionFilterSetFactory`.

At the same time, we should also move the connection types below the top-level ones to the same connection types. For example, the `forms` connection on the `Question` node does not even expose the `filters` field, and thus behaves and looks different to the top-level connections. For example:

```diff
-    forms = DjangoFilterConnectionField(
-        "caluma.caluma_form.schema.Form", filterset_class=filters.FormFilterSet
-    )
+    forms = DjangoFilterConnectionField(
+        "caluma.caluma_form.schema.Form",
+        filterset_class=CollectionFilterSetFactory(
+            filters.FormFilterSet, orderset_class=filters.FormOrderSet
+        ),
+    )
```

open-dynaMIX:
Yes, it's about time, we clean this up.

PR #1754, opened 2022-04-21
feat(filters): clean up all the old filter syntax

anehx:
BREAKING CHANGE:

We have for a long time had two syntaxes for filtering, one of which was
deprecated for over a year now and only kept for backwards
compatibility.  Finally, the old form has to go.

Old:
```graphql
query foo {
    allWorkItems(status: $status) {...}
}
```

New:
```graphql
query foo {
    allWorkItems(filter: [{status: $status}]) {...}
}
```

The new syntax also allows reusing the same filter type twice to narrow
down the result set even more. It also supports inverting the filters to
implement explicit exclusion. It is now time to get rid of the old
syntax.

The following changes are breaking:

- Removed deprecated `slug` filters in favor or `slugs` filters
- Removed top level filters
- Removed `orderByQuestionAnswerValue` filter in favor of
`documentAnswer` in the top level orderset
- Removed `orderBy` in filtersets in favor of top level `order` orderset
- Removed ordering by static meta attributes (configured with
`META_FIELDS` in favor of `meta` in the top level orderset

Resolves https://github.com/projectcaluma/caluma/issues/1489.
Closes https://github.com/projectcaluma/caluma/pull/1490


caluma/caluma_core/models.py
```diff
@@ -109,18 +108,4 @@ class Meta:
...
-    """
-
-    def formfield(self, **kwargs):
-        # Postfixing newly created filter with Argument to avoid conflicts
-        # with query nodes
-        meta = self.model._meta
-        name = to_camel_case(f"{meta.object_name}_{self.name}_argument")
-
-        return super().formfield(label=name, **kwargs)
+    pass
```
winged, line 111: Do we even need this anymore?

caluma/caluma_form/filters.py
```diff
@@ -486,7 +482,7 @@ class VisibleAnswerFilter(Filter):
     field_class = BooleanField
 
     def filter(self, qs, value):
-        if not value or not qs.exists():
+        if value in EMPTY_VALUES or not qs.exists():  # pragma: no cover
```
winged, line 485: I was just gonna suggest removing the `qs.exists()` check. But given that it triggers loading a whole document structure and evaluating the visibility JEXL, maybe we should add a comment instead to explain why we're doing this here, while we're at it :-) Might be a bit unrelated but useful for the next person to look at this

winged (commented):
Nice work on picking that up again, looking good so far!

winged (approved):

caluma/caluma_core/models.py
anehx: Yes, for backwards compatiblity in the migrations. I added a comment and removed all implementation in favor of `models.CharField`

caluma/caluma_form/filters.py
anehx: Added a comment

winged:
\m/

PR #1763, opened 2022-05-04
test(analytics): fix flaky test using creation date

anehx:
This test checks the month of a creation date which just changed some days ago. By freezing the date for the test this won't be an issue in the future.


winged (approved):
Ouch :rofl:

PR #1764, opened 2022-05-04
fix(build): conftest should be in generated package

winged:
When an application uses Caluma as a Django package, it is often very
beneficial to be able to reuse Caluma's fixtures. In fact, we do have
projects that do depend on Caluma's conftest. So, this should be built
into the package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,8 +16,6 @@ classifiers = [
 ]
 include = ["CHANGELOG.md"]
 exclude = [
-    "caluma/conftest.py",
-    "caluma/**/conftest.py",
     "caluma/**/tests",
 ]
 


winged:
Further discussion: Packing the tests into the release shouldn't hurt either, it's not 1990 anymore where we're overly space-limited

czosel:
Regarding conftest: Sounds reasonable! What would be the use case for the tests?

winged:
> Regarding conftest: Sounds reasonable! What would be the use case for the tests?

No particular use case right now, just didn't think it's required to hide them

czosel:
Alright, I don't mind either way :upside_down_face:

anehx (approved):

PR #1779, opened 2022-06-03
fix(pivot_table): pass context to child query

velrest:


winged (approved):
LGTM, just needed to `black` it

PR #656, opened 2019-08-21
feat(searchAnswers): Allow multiple searches

winged:
BREAKING: Rename the SearchAnswerFilterType to SearchAnswersFilterType
BREAKING: searchAnswers filter now takes a list instead of an object

fixes #655

PR #1787, opened 2022-06-10
fix(form): set questions of searchAnswers as required

Yelinz:
fixes bug from #656

not sure if breaking?

### DIFF
--- a/caluma/caluma_form/filters.py
+++ b/caluma/caluma_form/filters.py
@@ -299,7 +299,7 @@ class SearchLookupMode(Enum):
 class SearchAnswersFilterType(InputObjectType):
     """Lookup type to search in answers."""
 
-    questions = List(graphene.ID)
+    questions = List(graphene.ID, required=True)
     value = graphene.types.generic.GenericScalar(required=True)
     lookup = SearchLookupMode(required=False)
 
--- a/caluma/tests/__snapshots__/test_schema.ambr
+++ b/caluma/tests/__snapshots__/test_schema.ambr
@@ -2792,7 +2792,7 @@
   
   """Lookup type to search in answers."""
   input SearchAnswersFilterType {
-    questions: [ID]
+    questions: [ID]!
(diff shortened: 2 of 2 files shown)


anehx (changes requested):
Yes, this is breaking - please mark it as such. However I don't think the test is necessary since the schema test kind of already does that.

Otherwise LGTM.

Yelinz:
> Yes, this is breaking - please mark it as such. However I don't think the test is necessary since the schema test kind of already does that.
> 
> Otherwise LGTM.

Added breaking to commit msg

anehx:
@Yelinz as per the [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/) the body must lead with `BREAKING CHANGE:`

Yelinz:
> @Yelinz as per the [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/) the body must lead with `BREAKING CHANGE:`

Done

Yelinz:
Might be worth considering leaving it optional and implementing a search over all cases, will be discussed by @winged and @open-dynaMIX

winged (approved):
Nice catch, LGTM!

winged:
> Might be worth considering leaving it optional and implementing a search over all cases, will be discussed by @winged and @open-dynaMIX

Had an offline discussion with @open-dynaMIX - best solution would be to keep it "required", but add additional syntax to explicitly state that a search across all answers is desired. We would perfer it to be explicit than implicit by just not specifying questions.

anehx:
> Had an offline discussion with @open-dynaMIX - best solution would be to keep it "required", but add additional syntax to explicitly state that a search across all answers is desired. We would perfer it to be explicit than implicit by just not specifying questions.

I'd prefer that option as well, since a search across all answers might be very costly in terms of performance. This should only happen if explicitly required.

anehx (approved):

PR #1798, opened 2022-06-20
fix(deps): update dependencies of dependencies

anehx:
This will update graphene which results in a change of the schema which should be uncritical.

### DIFF
--- a/caluma/caluma_form/historical_schema.py
+++ b/caluma/caluma_form/historical_schema.py
@@ -7,7 +7,6 @@ from ..caluma_core.relay import extract_global_id
 from ..caluma_core.types import ConnectionField, CountableConnectionBase
 from . import models
 from .schema import (
-    QUESTION_ANSWER_TYPES,
     Answer,
     DateAnswer,
     FileAnswer,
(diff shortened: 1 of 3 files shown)


anehx:
The change in the type resolution function in the history is needed because of the graphene upstream change in the `resolve_type` function:
Before:
https://github.com/graphql-python/graphene/blob/27f19e5a905f95f3703a187aaabb67f1623d4a92/graphene/types/schema.py#L376-L391
After:
https://github.com/graphql-python/graphene/blob/efe4b89015d56a0b6c81abc6c0b4e46984aefd11/graphene/types/schema.py#L376-L383

The type needs to be a class, before we simply returned a string which will now resolve to the default type which would be `HistoricalStringAnswer`

caluma/caluma_form/historical_schema.py
```diff
@@ -221,3 +219,19 @@ class Query:
...
+    models.Question.TYPE_DATE: HistoricalDateAnswer,
+    models.Question.TYPE_CHOICE: HistoricalStringAnswer,
+    models.Question.TYPE_TEXTAREA: HistoricalStringAnswer,
+    models.Question.TYPE_TEXT: HistoricalStringAnswer,
+    models.Question.TYPE_TABLE: HistoricalTableAnswer,
+    models.Question.TYPE_FILE: HistoricalFileAnswer,
+    models.Question.TYPE_DYNAMIC_CHOICE: HistoricalStringAnswer,
+    models.Question.TYPE_DYNAMIC_MULTIPLE_CHOICE: HistoricalListAnswer,
+    models.Question.TYPE_CALCULATED_FLOAT: HistoricalFloatAnswer,
+}
```
winged, line 237: Just a suggestion - I'd move this bit up to the top of the module

winged (approved):
Looks good!

PR #1799, opened 2022-06-21
Remove various deprecations

### LINKED ISSUE #930: Deprecations

anehx:
We need to discuss whether the python API `start_case` should be renamed to `save_case` to be consistent with the GraphQL API. What is your take on this @winged @open-dynaMIX @fugal-dy @czosel ?

Resolves #930

### DIFF
--- a/caluma/caluma_workflow/serializers.py
+++ b/caluma/caluma_workflow/serializers.py
@@ -193,7 +193,7 @@ class SaveCompleteTaskFormTaskSerializer(SaveTaskSerializer):
         fields = SaveTaskSerializer.Meta.fields + ["form"]
 
 
-class CaseSerializer(SendEventSerializerMixin, ContextModelSerializer):
+class SaveCaseSerializer(ContextModelSerializer):
     workflow = serializers.GlobalIDPrimaryKeyRelatedField(
         queryset=models.Workflow.objects.prefetch_related("start_tasks")
     )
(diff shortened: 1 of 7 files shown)


caluma/caluma_workflow/serializers.py
```diff
@@ -226,34 +226,11 @@ def create(self, validated_data):
...
-        return instance
-
-    @transaction.atomic
-    def update(self, instance, validated_data):
-        self.send_event(
-            events.pre_create_case, case=instance, validated_data=validated_data
-        )
-        instance = super().update(instance, validated_data)
-        self.send_event(events.post_create_case, case=instance)
-        return instance
```
anehx, line 249: FYI this has been completely removed since it's IMO wrong to fire pre and post **create** events in an update mutation.

czosel:
I discussed this with @anehx shortly - we lean towards leaving `start_case`, because that's basically the only use-case for the method. For updates, we'd work with the django model directly. Our feeling is that the python API doesn't have to mirror the GraphQL API exactly and should be optimized to be as intuitive as possible.

winged:
Code looks good - I have no opinion regarding "startcase" vs "savecase". Aesthetically speaking, I think keeping the distinction has it's merits, as the intention of a mutation/request is communicated clearly. But technically, I agree that it can and probably should be dropped. Less code = less bugs

anehx:
Can I get a review on this @winged @open-dynaMIX @fugal-dy @czosel ?

czosel (approved):
LGTM, I'd say let's get this merged before we release a new stable version. @winged do you agree?

winged:
Alright, let's merge it!

PR #1800, opened 2022-06-23
fix: reopenCase should disregard WorkItems in status REDO

open-dynaMIX:


winged (approved):
Nice work! LGTM, but let's cc @anehx and @czosel for a quick glance at the "business case"

anehx:
Hmm, should we maybe d

anehx:
Sorry, wrong button :wink: I was wondering whether it makes sense to have `REDO` work items in a completed case or not. I tend to say yes, to keep the "history" correct..

czosel:
I'm not too familiar with the whole REDO pattern, so I can't really comment on this. If the feature is considered stable, we should extend it's documentation a bit (https://caluma.gitbook.io/caluma-docs/workflow/redo-pattern). I'd be happy to help with that, but I'd need someone on my side who knows the feature better than I do :upside_down_face:

PR #1804, opened 2022-06-24
Dynamic option fixes

anehx:


winged (approved):
Nice catch & good work! LGTM

Issue #1780, opened 2022-06-03
RFC: New question type "MultipleFileQuestion"


czosel:
When using file questions, it is often desired that more that one file can be uploaded. Currently this is only possible in Caluma using TableQuestions, which doesn't yield a good UX when the table form consists of just one FileQuestion.

We'd propose to add a new question type "MultipleFileQuestion" which is very similar to the existing "FileQuestion", but allows uploading more than one file.

The API could look like this:

### FileAnswer (existing, just for reference)

Request:
```json
{
    "operationName": "SaveDocumentFileAnswer",
    "variables": {
        "input": {
            "question": "some-slug",
            "document": "7da68336-8ccd-47cb-8800-81e4444bc649",
            "value": "test.pdf"
        }
    },
    "query": "mutation SaveDocumentFileAnswer($input: SaveDocumentFileAnswerInput!) { ... }"
}
```

Response:
```json
{
  "answer": {
    "id": "some-id",
    "question": {
      "slug": "some-slug",
      "__typename": "FileQuestion"
    },
    "fileValue": {
      "id": "some-id",
      "uploadUrl": "",
      "downloadUrl": "",
      "metadata": null,
      "name": "pdf-test (copy).pdf",
      "__typename": "File"
    },
  },
}
```

### MultipleFileAnswer

Request:
```json
{
    "operationName": "SaveDocumentMultipleFileAnswer",
    "variables": {
        "input": {
            "question": "some-id",
            "document": "7da68336-8ccd-47cb-8800-81e4444bc649",
            "value": [
              // existing file
              {
                "id": <File UUID>,
                "name": "pdf-test (copy).pdf"
              },
              {
                "name": "pdf-test (copy).pdf" // new file
              }
              // deleting files: just leave them out
            ]
        }
    },
    "query": "mutation SaveDocumentFileAnswer($input: SaveDocumentFileAnswerInput!) { ... }"
}
```

Response:
```json
{
  "answer": {
    "id": "some-id",
    "question": {
      "slug": "some-slug",
      "__typename": "MultipleFileQuestion"
    },
    "multipleFileValue": [
      {
        "id": "some_id",
        "uploadUrl": "",
        "downloadUrl": "",
        "metadata": null,
        "name": "pdf-test (copy).pdf",
        "__typename": "File"
      },
      ...
    ]
  }
}
```

StephanH90:
Does this allow you to rename the files as well by supplying an `ID` with a different `name`?

[...]

PR #1809, opened 2022-07-01
feat(form): multiple files in file questions

winged:
Change the model to allow multiple files per answer to be stored. Migrate
the existing questions to the new question type, and migrate the existing
answers to the new schema.

Updated the GraphQL schema to allow saving and updating multiple files
on an answer. The answer's value is a list of dicts, each containing
a file name, and optionally an ID. When you're creating new files,
add an entry without ID to the value. Caluma will then save the file,
and return it back WITH and ID.

BREAKING CHANGE: This renames the question type constant for file questions,
and changes the semantics of the answer value for file questions as well:
It is now a list of dicts instead of a single string. The response type
for querying file(s) answers now also is a list instead of a single dict.

For details / RFC see #1780


caluma/caluma_form/tests/test_validators.py
```diff
@@ -298,11 +300,11 @@ def test_validate_data_source(
         (Question.TYPE_INTEGER, None, None),
         (Question.TYPE_FLOAT, None, None),
         (Question.TYPE_DATE, None, None),
-        (Question.TYPE_FILE, None, None),
+        (Question.TYPE_FILES, None, []),
         (Question.TYPE_TEXTAREA, None, None),
     ],
 )
-@pytest.mark.parametrize("answer__date,answer__file", [(None, None)])
+@pytest.mark.parametrize("answer__date", [(None)])
```
fugal-dy, line 307:
```suggestion
@pytest.mark.parametrize("answer__date", [None])
```

fugal-dy (approved):
LGTM as far as I can tell. relly just nitpickery

winged:
Merging, as @open-dynaMIX approved off-github

PR #1810, opened 2022-07-05
fix(form): use potentially prefetched answer document set for table rows

anehx:
This allows prefetching the answer documents (table rows) to avoid DB
hits when generating the form structure.

### DIFF
--- a/caluma/caluma_form/structure.py
+++ b/caluma/caluma_form/structure.py
@@ -3,7 +3,7 @@ import weakref
 from functools import singledispatch
 from typing import List, Optional
 
-from .models import AnswerDocument, Question
+from .models import Question
 
 
 def object_local_memoise(method):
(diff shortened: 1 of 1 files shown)


caluma/caluma_form/structure.py
```diff
@@ -116,7 +116,10 @@ def children(self):
         if not self.answer:
             return []  # pragma: no cover
 
-        rows = AnswerDocument.objects.filter(answer=self.answer).order_by("sort")
+        rows = sorted(
+            self.answer.answerdocument_set.all(),
+            key=lambda answer_document: answer_document.sort,
+        )
```
winged, line 122:
```suggestion
        # We're sorting in python-space here to reuse the already-executed query
        # for the answerdocument_set. Sorting in DB would re-issue the query
        rows = sorted(
            self.answer.answerdocument_set.all(),
            key=lambda answer_document: answer_document.sort,
        )
```

winged (approved):
Comment why it's sorted in python, otherwise LGTM. Feel free to merge once comment is added

caluma/caluma_form/structure.py
anehx: Done

PR #1813, opened 2022-07-06
feat(form): multiple files in file questions

anehx:
Change the model to allow multiple files per answer to be stored. Migrate
the existing questions to the new question type, and migrate the existing
answers to the new schema.

Updated the GraphQL schema to allow saving and updating multiple files
on an answer. The answer's value is a list of dicts, each containing
a file name, and optionally an ID. When you're creating new files,
add an entry without ID to the value. Caluma will then save the file,
and return it back WITH and ID.

BREAKING CHANGE: This renames the question type constant for file questions,
and changes the semantics of the answer value for file questions as well:
It is now a list of dicts instead of a single string. The response type
for querying file(s) answers now also is a list instead of a single dict.


anehx:
Please merge this simultaneously with https://github.com/projectcaluma/ember-caluma/pull/2040

derrabauke:
There is a question by @StephanH90 in the RFC to this feature. https://github.com/projectcaluma/caluma/issues/1780#issuecomment-1177103089 

Should we add the requested feature, @winged ?

winged:
This PR is missing some commits from the original PR, thus failing CI. I've re-created it here: #1821

So now we can close this one

PR #1814, opened 2022-07-07
fix(workflow): remove and fix ordering keys

Yelinz:
remove non exsisting ordering keys for work item and
task, fix relational ordering keys for case and work item

removes allow_all_forms and is_published from work item and task
rename is_archived on work item to task__is_archived
append workflow__ to keys on case which are from workflow relation

### DIFF
--- a/caluma/caluma_workflow/filters.py
+++ b/caluma/caluma_workflow/filters.py
@@ -109,13 +109,8 @@ class CaseOrderSet(BaseFilterSet):
         fields=[
             "created_at",
             "modified_at",
-            "allow_all_forms",
-            "description",
-            "is_archived",
-            "is_published",
-            "name",
             "status",
-            "slug",
+            "document__form__name",
         ],
     )
     document_answer = AnswerValueOrdering(document_via="document")
@@ -140,12 +135,10 @@ class TaskOrderSet(BaseFilterSet):
         fields=[
             "created_at",
             "modified_at",
-            "allow_all_forms",
             "lead_time",
             "type",
             "description",
             "is_archived",
-            "is_published",
             "name",
             "slug",
         ],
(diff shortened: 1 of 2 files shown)


caluma/caluma_workflow/filters.py
```diff
@@ -109,13 +109,18 @@ class CaseOrderSet(BaseFilterSet):
...
+            "workflow__name",
+            "workflow__description",
+            "workflow__is_archived",
+            "workflow__is_published",
+            "workflow__allow_all_forms",
+            "document__form__slug",
+            "document__form__name",
+            "document__form__description",
+            "document__form__is_archived",
+            "document__form__is_published",
```
winged, line 123: I'm not sure those are all useful. It's only about ordering after all. For example, are you sure we will (soon-ish) need to sort cases by the fact whether it's document's form is actually published? Or the form's description? I don't think so. 

Therefore please only add new order-by fields that are either immediately used or easily conceivable of a use case. Most of those are not, IMHO

caluma/caluma_workflow/filters.py
```diff
@@ -200,14 +203,12 @@ class WorkItemOrderSet(BaseFilterSet):
...
             "closed_at",
-            "allow_all_forms",
             "description",
-            "is_archived",
-            "is_published",
             "name",
             "deadline",
             "status",
             "slug",
+            "task__is_archived",
```
winged, line 211: Same here - do we really need to sort work items by whether the associated task is archived? This is not something the users of work items would be interested in

winged (changes requested):

caluma/caluma_workflow/filters.py
Yelinz: removed

caluma/caluma_workflow/filters.py
Yelinz: only kept slug and name

winged: Still not sure we really need those

winged (approved):

PR #1815, opened 2022-07-07
feat: searchanswers in forms

open-dynaMIX:
This is the result of a hacking-session with @winged. I didn't have the time to finish this before the holidays.

@Yelinz This is needed for mySAGW. Would be great if you could push this forward, maybe with additional help from a backender.

### DIFF
--- a/caluma/caluma_form/filters.py
+++ b/caluma/caluma_form/filters.py
@@ -20,7 +20,7 @@ from ..caluma_core.filters import (
     SearchFilter,
 )
 from ..caluma_core.ordering import AttributeOrderingFactory, MetaFieldOrdering
-from ..caluma_form.models import Answer, DynamicOption, Question
+from ..caluma_form.models import Answer, DynamicOption, Form, Question
 from ..caluma_form.ordering import AnswerValueOrdering
 from . import models, validators
 
@@ -299,7 +299,8 @@ class SearchLookupMode(Enum):
 class SearchAnswersFilterType(InputObjectType):
     """Lookup type to search in answers."""
 
-    questions = List(graphene.ID, required=True)
+    questions = List(graphene.ID)
+    forms = List(graphene.ID)
     value = graphene.types.generic.GenericScalar(required=True)
     lookup = SearchLookupMode(required=False)
 
@@ -325,6 +326,9 @@ class SearchAnswersFilter(Filter):
 
     def __init__(self, *args, **kwargs):
         self.document_id = kwargs.pop("document_id")
+        self.UNSUPPORTED_QUESTION_TYPES = [
+            type for type in Question.TYPE_CHOICES if type not in self.FIELD_MAP.keys()
+        ]
         super().__init__(*args, **kwargs)
 
     def filter(self, qs, value):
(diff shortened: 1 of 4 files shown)


caluma/caluma_form/filters.py
```diff
@@ -325,6 +326,9 @@ class SearchAnswersFilter(Filter):
 
     def __init__(self, *args, **kwargs):
         self.document_id = kwargs.pop("document_id")
+        self.UNSUPPORTED_QUESTION_TYPES = [
+            type for type in Question.TYPE_CHOICES if type not in self.FIELD_MAP.keys()
+        ]
```
winged, line 331: This can be moved to class scope as a constant - no need to do it every time the filter is instantiated.

Better yet, remove the constant - it's the exclusion of FIELD_MAP, so we can just use that one to check for supported question types

caluma/caluma_form/filters.py
```diff
@@ -339,9 +343,26 @@ def filter(self, qs, value):
...
+            raise exceptions.ValidationError(
+                '"forms" and/or "questions" parameter must be set'
+            )
 
-        questions = self._validate_and_get_questions(value["questions"])
+        if raw_forms:
+            forms = self._validate_and_get_forms(raw_forms)
+            questions = Form.get_all_questions(forms).exclude(
+                type__in=self.UNSUPPORTED_QUESTION_TYPES
+            )
```
winged, line 358: Wouldn't it be a good idea to validate the questions list and reject the call if the client supplies unsupported questions? Otherwise, we might get unexpected behaviour

winged (changes requested):
Looks good so far, just some suggestions for cleanup

caluma/caluma_form/filters.py
Yelinz: refactored to check before, so the exclude isnt needed

caluma/caluma_form/filters.py
Yelinz: refactored away

winged (changes requested):
Nice work! I'd just like to see a bit more explicit comments / docstrings (see suggestions)

caluma/caluma_form/filters.py
```diff
@@ -339,13 +340,36 @@ def filter(self, qs, value):
...
 
-        questions = self._validate_and_get_questions(value["questions"])
+        if raw_questions:
+            questions = self._validate_and_get_questions(raw_questions)
+
+        if raw_forms:
+            forms = self._validate_and_get_forms(raw_forms)
+            form_questions = Form.get_all_questions(forms)
+            # combine querysets
+            questions = questions | form_questions
```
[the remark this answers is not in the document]
open-dynaMIX: Done

caluma/caluma_form/filters.py
```diff
@@ -299,7 +299,8 @@ class SearchLookupMode(Enum):
 class SearchAnswersFilterType(InputObjectType):
     """Lookup type to search in answers."""
```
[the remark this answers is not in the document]
open-dynaMIX: Done

winged (approved):

PR #1828, opened 2022-07-19
fix(analytics): provide more correct supported functions in analytics

winged:
Identifier fields cannot be sorted or have string functions applied to them in a
meaningful manner - nobody should require the "max" of an UUID for example.

Also, Meta fields (not their contents!) cannot have any useful aggregate
function applied to them either


open-dynaMIX (approved):
Nice work! :tada:

PR #1835, opened 2022-08-03
feat(workflow): add ids filter to case

Yelinz:


caluma/caluma_workflow/filters.py
```diff
@@ -86,6 +87,7 @@ class Meta:
 
 class CaseFilterSet(MetaFilterSet):
     id = GlobalIDFilter()
```
winged, line 89: I think we could deprecate the `id` filter - no need to have two of them that do the same thing

winged (approved):
LGTM!

caluma/caluma_workflow/filters.py
Yelinz: Removed id filter

winged: I thought deprecate, not straight out remove ;-) But given that we're not in a stable `main` anyway right now, I think we *should* be OK with this. Not sure how often it's used in application code and will break (I think normal direct-access use would be via `node` anyway)..

What do you guys think @czosel @anehx ?

czosel: I’d prefer to keep breaking changes to a minimum, so I’m in favor of keeping it. I wouldn’t even mind keeping both filters in the long run since they have zero maintenance overhead.

Yelinz: Added deprecation warning instead of removing

PR #1836, opened 2022-08-04
fix: multiple fixes for analytics

open-dynaMIX:


winged (changes requested):
Looks good! However, for future archeology, I'd like to see a more detailed description in the commit message:

* What has changed
* Why

Something along the lines of:

> Add form identifier as an available field in documents.
> Manually calculate summary row to avoid SQL complexity/bugs

open-dynaMIX:
@winged True, done!

winged (approved):

winged:
> @winged True, done!

Awesome! let's merge

PR #1840, opened 2022-08-11
chore: make sure rollback happens also for failed tests

open-dynaMIX:


winged (approved):
Nice find!

winged:
This might have been the cause of some flaky tests in the past...

PR #1843, opened 2022-08-12
Redo work item / reopen case improvements

luytena:
- Add `is_redoable` property to work item schema
- Allow empty `next` on flows if `redoable` is given
- Allow reopening child cases if the parent work item is ready
- Don't create a new work item after completing a redone work item if there are succeeding work items in status redo

### DIFF
--- a/caluma/caluma_workflow/domain_logic.py
+++ b/caluma/caluma_workflow/domain_logic.py
@@ -180,8 +180,16 @@ class CompleteWorkItemLogic:
             )
 
             if all_siblings_complete:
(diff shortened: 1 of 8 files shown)


czosel:
Commit 2-4 LGTM, commit 1 still gets a refactoring (property instead of filter) as discussed with @anehx

winged (approved):
Looks good so far, I'm just not sure how to feel about that redoable filter. This will potentially cause a huge performance hit if it's called in a too-broad context.

winged:
Awesome! I like this approach much better

PR #1846, opened 2022-08-15
fix(form): fix migration problem with historicalanswer

winged:
We cannot handle historical answers and "current" answers the same way
when migrating their files: Historical answer's files need to be fetched
from history as well, and their history needs to be updated also.

When fetching the historical files of an answer, we need to fetch
exactly the files that were there at the given date/time, not all that
have ever been created (but may have been deleted in the mean time)

This supersedes #1839


open-dynaMIX (approved):
Awesome :tada:

PR #1849, opened 2022-08-18
fix: fix multifile migration again

open-dynaMIX:
The previous approach ignored all HistoricalFiles that belong to an
Answer that doesn't exist anymore.
This approach takes all HistoricalFiles into account and hence provides
a more robust procedure.


winged (approved):
iteration++ ;-D

PR #1852, opened 2022-08-19
feat(caluma): add new work item order key

Yelinz:


caluma/caluma_workflow/filters.py
```diff
@@ -201,6 +201,7 @@ class WorkItemOrderSet(BaseFilterSet):
             "deadline",
             "status",
             "slug",
+            "case__document__form__name"
```
open-dynaMIX, line 204:
```suggestion
            "case__document__form__name",
```

Or just run black.

open-dynaMIX (approved):
Otherwise LGTM :+1:

caluma/caluma_workflow/filters.py
Yelinz: used black

open-dynaMIX (approved):
Awesome :tada:

PR #1856, opened 2022-08-22
fix(analytics): handle None and ignore 0 in summary

open-dynaMIX:


caluma/caluma_analytics/pivot_table.py
```diff
@@ -106,7 +106,7 @@ def get_records(self):
             return result
 
     def _update_summary(self, field, value):
-        if field.function in [field.FUNCTION_SUM, field.FUNCTION_COUNT]:
+        if value and field.function in [field.FUNCTION_SUM, field.FUNCTION_COUNT]:
```
winged, line 109: I'd add a comment here, it's not filtering out just `None` as the PR says, but Zeroes (0) as well. Which isn't wrong, as we're just summing up counts and sums here, but might confuse future readers (did for me for a second)

open-dynaMIX: Done

winged (approved):
LGTM now!

PR #1862, opened 2022-08-26
fix(permissions): make sure user permission classes call super

### LINKED ISSUE #1830: Permission classes `IsAuthenticated` and `CreatedByGroup` do not call super `has_permission` / `has_object_permission`
Inheriting from `IsAuthenticated` or `CreatedByGroup` in custom permission classes causes `has_permission` and `has_object_permission` of the `BasePermission` class to be overridden and since the super method isn't called, any further mutation permissions defined through `permission_for` or `object_permission_for` are never called. To allow custom permission classes to make use of the `IsAuthenticated` / `CreatedByGroup` and still further restrict mutation permissions, the classes would need to call the `BasePermission` super methods. Are there any further considerations that should be taken into account?

`IsAuthenticated`, `CreatedByGroup`
https://github.com/projectcaluma/caluma/blob/3bc878cbd90efb0c8ce8ea10b819eb43464c7ecb/caluma/caluma_user/permissions.py#L1-L15

`BasePermission`
https://github.com/projectcaluma/caluma/blob/3bc878cbd90efb0c8ce8ea10b819eb43464c7ecb/caluma/caluma_core/permissions.py#L124-L136

anehx:
Fixes #1830

### DIFF
--- a/caluma/caluma_user/permissions.py
+++ b/caluma/caluma_user/permissions.py
@@ -5,11 +5,16 @@ class IsAuthenticated(BasePermission):
     """Only allow authenticated users to execute mutations."""
 
     def has_permission(self, mutation, info):
-        return info.context.user.is_authenticated
+        return info.context.user.is_authenticated and super().has_permission(
+            mutation, info
+        )
 
 
 class CreatedByGroup(BasePermission):
     """Only allow mutating data that belongs to same group as current user."""
 
     def has_object_permission(self, mutation, info, instance):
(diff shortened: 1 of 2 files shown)


winged (approved):
How did this not bite us earlier? Nice catch, and code LGTM as well!

anehx:
> How did this not bite us earlier? Nice catch, and code LGTM as well!

I think most of the time we extended the `BasePermission` class directly..

PR #1870, opened 2022-09-09
fix(events): don't emit create work item events when updating

anehx:
BREAKING CHANGE: The `saveWorkItem` mutation doesn't emit `pre_create_work_item` `post_create_work_item` anymore since this is not the expected behaviour.

### DIFF
--- a/caluma/caluma_workflow/serializers.py
+++ b/caluma/caluma_workflow/serializers.py
@@ -452,7 +452,7 @@ class CancelWorkItemSerializer(ContextModelSerializer):
         fields = ["id", "context"]
 
 
-class SaveWorkItemSerializer(SendEventSerializerMixin, ContextModelSerializer):
+class SaveWorkItemSerializer(ContextModelSerializer):
     work_item = serializers.GlobalIDField(source="id")
     name = CharField(
         required=False,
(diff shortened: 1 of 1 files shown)


luytena:
@anehx Couldn't it be a breaking change if another project depends on the pre_create / post_create hooks on work-item updates?

caluma/caluma_workflow/serializers.py
```diff
@@ -467,17 +467,6 @@ class SaveWorkItemSerializer(SendEventSerializerMixin, ContextModelSerializer):
         help_text=models.WorkItem._meta.get_field("description").help_text,
     )
 
-    @transaction.atomic
-    def update(self, instance, validated_data):
-        self.send_event(
```
luytena, line 472: The SendEventSerializerMixin might no longer be necessary either

anehx: Yes, fixed!

anehx:
> @anehx Couldn't it be a breaking change if another project depends on the pre_create / post_create hooks on work-item updates?

Absolutely, marked it as breaking change

luytena (approved):

PR #1885, opened 2022-09-27
fix(deps): pin graphene-django to v3.0.0b7

anehx:
We need to pin this dependency in order to not break consuming apps because we are somehow not compatible with later versions yet.


winged (approved):

winged:
Let's get this released ASAP - I'm working on fixing it properly, but might take a bit longer. This seems like an appropriate fix for now

anehx:
Alright, I'll draft a release

PR #1887, opened 2022-09-27
chore(ci): use built in cache of setup-python action

anehx:
This is the recommended way of caching poetry dependencies in github actions


winged (approved):
LGTM

PR #1914, opened 2022-11-07
feat(analytics): add support for form name extraction

winged:
While at it, for better "generic" extraction, we now also allow extracting answers from documents without specifying the form in the data source, using `document[*].answer_slug` syntax.

Note: This is kind-of breaking, because we replaced the `form_id` field on documents by `caluma_form.slug`. However, it's not "officially" breaking as the analytics part is still considered unstable.

I'm still debating whether a data migration should be created to convert any existing `form_id` analytics fields, but then again, it's probably not worth it...


open-dynaMIX:
> I'm still debating whether a data migration should be created to convert any existing form_id analytics fields, but then again, it's probably not worth it...

Please also add a migration. Caluma analytics is already running in production in at least one customer project.

winged:
> > I'm still debating whether a data migration should be created to convert any existing form_id analytics fields, but then again, it's probably not worth it...
> 
> Please also add a migration. Caluma analytics is already running in production in at least one customer project.

I have now also added a corresponding migration, with a test

open-dynaMIX (approved):
:tada:

PR #1917, opened 2022-11-11
feat(data_source): add question and context arguments for data sources

anehx:
https://github.com/projectcaluma/ember-caluma/pull/2205

This implements further context for data sources since they currently only have the `user` as context. We added the `question` and a JSON object `context` as new arguments for the data source `get_data` method. This way we have much more possibilities when writing data sources.

### DIFF
--- a/caluma/caluma_data_source/schema.py
+++ b/caluma/caluma_data_source/schema.py
@@ -1,7 +1,8 @@
-from graphene import ConnectionField, String
+from graphene import ConnectionField, JSONString, String
 from graphene.types import ObjectType
 
 from ..caluma_core.types import CountableConnectionBase
+from ..caluma_form.models import Question
 from .data_source_handlers import get_data_source_data, get_data_sources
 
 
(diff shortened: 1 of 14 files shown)


caluma/caluma_data_source/schema.py
```diff
@@ -33,4 +33,4 @@ def resolve_all_data_sources(self, info, **kwargs):
         return get_data_sources()
 
     def resolve_data_source(self, info, name, **kwargs):
-        return get_data_source_data(info.context.user, name)
+        return get_data_source_data(info.context.user, name, None, None)
```
winged, line 36: Shouldn't we accept `question` and `context` here as well in the GQL schema? I think we could still make it optional, but depending on the data source, it might not be able to provide data without the new inputs available

winged (changes requested):
Looks awesome! Just a small-ish nitpick

caluma/caluma_data_source/schema.py
anehx: Good idea, I'll try to add this!

anehx: Done!

winged (changes requested):
Awesome!

One final issue though: this will break all existing data source implementations, so the commit should be marked as breaking.

anehx:
@winged why will it break them? because of the additional arguments?

winged:
> @winged why will it break them? because of the additional arguments?

Exactly. You're passing two new arguments that the old DS implementations do not accept yet. This will cause the call to fail

anehx:
@winged good point! I marked the commit as breaking.

winged (approved):

PR #1921, opened 2022-11-18
Fix analytics sql generation

winged:
Analytics uses generated SQL from the visibility layer of the application. However, only the part after the WHERE is actually used. Now if a visibility layer adds expressions that affect the FROM bit of an SQL statement, and these expressions contain query parameters, we have too many parameters in for the remaining SQL fragment.

Those parameters need to be discarded, as we don't use the SQL that came along with them.

Then, as a drive-by fix: It doesn't make sense to extract a summary row for simple tables - there is nothing to summarize in a meaningful way. Fixe a crash where the frontend would request a summary nonetheless.


open-dynaMIX (approved):
:tada:

PR #1922, opened 2022-11-21
feat: allow using client id as user for client-secret auth

winged:
When authentication happens with a "service user" instead of a regular user, we may not get a username. However, we can now configure Caluma to use the client ID as username instead.

This is optional to ensure that the behaviour is intended, and no "wrong" fallback happens in case the username claim was just configured incorrectly.

### DIFF
--- a/caluma/caluma_user/models.py
+++ b/caluma/caluma_user/models.py
@@ -24,7 +24,16 @@ class OIDCUser(BaseUser):
         super().__init__()
 
         self.claims, self.claims_source = self._get_claims(userinfo, introspection)
-        self.username = self.claims[settings.OIDC_USERNAME_CLAIM]
+
+        if (
+            self.claims_source == "introspection"
+            and settings.OIDC_CLIENT_AS_USERNAME
+            and settings.OIDC_USERNAME_CLAIM not in self.claims
+        ):
+            self.username = self.claims[settings.OIDC_CLIENT_CLAIM]
+        else:
+            self.username = self.claims[settings.OIDC_USERNAME_CLAIM]
+
         self.groups = self.claims.get(settings.OIDC_GROUPS_CLAIM)
         self.group = self.groups[0] if self.groups else None
         self.token = token
--- a/caluma/caluma_user/tests/test_views.py
+++ b/caluma/caluma_user/tests/test_views.py
@@ -92,3 +92,33 @@ def test_no_client_id(rf, requests_mock, settings):
     request = rf.get("/graphql", HTTP_AUTHORIZATION=authentication_header)
     response = views.AuthenticationGraphQLView.as_view()(request)
     assert response.status_code == status.HTTP_401_UNAUTHORIZED
+
+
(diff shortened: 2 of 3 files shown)


caluma/caluma_user/models.py
```diff
@@ -24,7 +24,16 @@ def __init__(self, token: str, userinfo: dict = None, introspection: dict = None
...
-        self.username = self.claims[settings.OIDC_USERNAME_CLAIM]
+
+        if (
+            self.claims_source == "introspection"
+            and settings.OIDC_CLIENT_AS_USERNAME
+            and settings.OIDC_USERNAME_CLAIM not in self.claims
+        ):
+            self.username = self.claims[settings.OIDC_CLIENT_CLAIM]
+        else:
+            self.username = self.claims[settings.OIDC_USERNAME_CLAIM]
```
open-dynaMIX, line 35: I prefer the default to be on top, tather than buried inside an `else`.

```suggestion
        self.username = self.claims[settings.OIDC_USERNAME_CLAIM]
        if (
            self.claims_source == "introspection"
            and settings.OIDC_CLIENT_AS_USERNAME
            and settings.OIDC_USERNAME_CLAIM not in self.claims
        ):
            self.username = self.claims[settings.OIDC_CLIENT_CLAIM]
```

caluma/caluma_user/tests/test_views.py
```diff
@@ -92,3 +92,33 @@ def test_no_client_id(rf, requests_mock, settings):
...
+    requests_mock.post(
+        settings.OIDC_INTROSPECT_ENDPOINT, text=json.dumps({"client_id": "test_client"})
+    )
+
+    request = rf.get("/graphql", HTTP_AUTHORIZATION="Bearer foo")
+    if expect_success:
+        views.AuthenticationGraphQLView.as_view()(request)
+        assert request.user.is_authenticated == expect_success
+    else:
+        with pytest.raises(Exception):
```
open-dynaMIX, line 123: Might make sense to assert a more specific Exception.

open-dynaMIX (changes requested):
Just two minor nitpickts. Otherwise LGTM :+1:

caluma/caluma_user/models.py
winged: This way it will crash, as `settings.OIDC_USERNAME_CLAIM in self.claims == False`

But we could invert the `if` statement to achieve the same thing

open-dynaMIX: Ah, I see it now :see_no_evil: I've updated my suggestion above, so it wouldn't crash. But you can also do what you suggested in your comment.

winged: Then again, the "default" is also the "fallback" in case we *do* have an introspection source, but `settings.OIDC_USERNAME_CLAIM in self.claims == True` so after thinking about it, I'd rather keep it as is

open-dynaMIX (approved):

PR #1928, opened 2022-11-30
fix(analytics): fix commandline output crash with certain aliases

winged:
The run_analytics command used the column aliases as keys in the format string for outputting tabular data. This crashes if the aliases contain values (like dots, etc) that are not compatible with format string syntax. Thus, we hash the aliases before using it in the format string. A little less debuggable, but guaranteed valid keys.


open-dynaMIX (approved):
:tada:

PR #1936, opened 2022-12-12
chore: release 8.0.0-beta.22

open-dynaMIX:


winged (approved):
LGTM!

But I feel like we should start thinking about making a non-beta release again soon... I think we got all the breaking stuff in that we wanted :)

PR #1937, opened 2022-12-15
fix: properly quote aliases - small additional fixes

open-dynaMIX:


fugal-dy (approved):
now that's properly quoted.

aka LGTM

PR #1939, opened 2022-12-21
fix(analytics): fix form identifier; disable subform form

open-dynaMIX:
Displaying form information for subforms is broken. Since there is no pressing use case for this anyway, this commit disables it alltogether.


winged (approved):
Nice catch, LGTM

PR #1940, opened 2022-12-21
Updates

anehx:


pyproject.toml
```diff
@@ -43,36 +43,36 @@ urllib3 = "^1.26.8"
 uWSGI = "^2.0.20"
 graphql-core = "~3.1.7"
 
-[tool.poetry.dev-dependencies]
-black = "22.10.0"
+[tool.poetry.group.dev.dependencies]
```
winged, line 46:
```suggestion
[tool.poetry.group.dev.dependencies]
# Dev dependencies are always pinned to an exact version (No caret or tilde matching)
# to ensure everyone uses the same version of the tools, especially linters & formatters
```

winged (approved):
LGTM otherwise

pyproject.toml
anehx: Thanks for the text! Done.

Issue #1953, opened 2023-01-27
AnalyticsTable description


open-dynaMIX:
I'd like to have a description field on the `AnalyticsTable`. To be consistent, it should also be localized. This should also be exposed over the API and be integrated in the frontend.

PR #1954, opened 2023-01-31
feat(analytics): Add nullable field description to analytics table

gh-PonyM:
Implements #1953


winged (changes requested):
Nice work, let's see what CI thinks :)

The commit doesn't look like [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) yet, but otherwise LGTM!

winged (approved):

DadaDaMotha:
Running pytest in parallel causes the issue. As you pointed out last time, running them with `poetry run pytest -n 0 --reuse-db` inside docker passes the tests. What should we do in this case?

caluma/caluma_analytics/filters.py
```diff
@@ -53,7 +50,7 @@ class AnalyticsTableOrderSet(FilterSet):
     meta = MetaFieldOrdering()
     attribute = AttributeOrderingFactory(
         models.AnalyticsTable,
-        fields=["created_at", "modified_at", "slug", "name"],
+        fields=["created_at", "modified_at", "slug", "name", "description"],
```
winged, line 53: Does it make sense to allow ordering by description? No need to fix, I just don't think it's particularly useful

caluma/caluma_analytics/tests/test_table.py
```diff
@@ -71,15 +83,25 @@ def test_create_table(db, snapshot, schema_executor):
...
     )
 
     assert not result.errors
-    assert models.AnalyticsTable.objects.filter(pk="test-table").exists()
+    item = models.AnalyticsTable.objects.filter(pk="test-table").first()
+    assert str(item.description) == "Foo bar"
 
     snapshot.assert_match(result.data)
+    assert result.data["saveAnalyticsTable"]["analyticsTable"]
+    print(result.data)
```
winged, line 98: Print should be removed in "production" code

winged (changes requested):
Some minor cleanup required, otherwise LGTM

caluma/caluma_analytics/filters.py
open-dynaMIX: Sorry, should have marked as `draft`. Wanted to first fix CI, that's why I pushed early, even before I took a closer look at the changes :sweat_smile:

winged (approved):

PR #1955, opened 2023-02-03
fix(analytics): handle missing dates

open-dynaMIX:


caluma/caluma_analytics/simple_table.py
```diff
@@ -824,7 +824,7 @@ def parse_value(self, value):
         same datatype which leads to inconsistency. E.g versions < 12 returns a
         float while later versions return a decimal.
         """
-        return int(value)
+        return int(value) if value else value
```
winged, line 827:
```suggestion
        return int(value) if value is not None else None
```
Otherwise, `0` (zero) might not be correctly converted

open-dynaMIX: Fixed. But technically `0` would already be `int` ;)

winged: But not `decimal.Decimal(0)`, which is why this code even exists

winged (approved):

caluma/caluma_analytics/simple_table.py
open-dynaMIX: True

PR #1963, opened 2023-02-16
fix: fix calc answer evaluation when adding new table rows

open-dynaMIX:
* fix: fix calc answer evaluation when adding new table rows
* chore(tests): do not use nested documents
   In the past we've had some misconceptions about document structure and
   introduced nested documents. This commit cleansup a remnant of this
   mistake, that was forgotten back then.
 * chore(tests): migrate to current state after migration tests
   This commit introduces a `post_migrate_to_current_state`-fixture to properly
   cleanup after our migration tests.

### DIFF
--- a/caluma/caluma_form/migrations/0047_recalculate_calc_answers.py
+++ b/caluma/caluma_form/migrations/0047_recalculate_calc_answers.py
@@ -1,0 +1,54 @@
+from django.db import migrations
+
+from caluma.caluma_form.jexl import QuestionJexl
+from caluma.caluma_form.structure import FieldSet
+
+
+def recalculate_answer(answer):
+    """
+    Recalculate a single calc-answer.
+
+    This is more or less a copy/paste from `_update_or_create_calc_answer` in order to
+    use the old models.
+    """
+    root_doc = answer.document.family
+
+    struc = FieldSet(root_doc, root_doc.form)
+    field = struc.get_field(answer.question.slug)
+
+    # skip if question doesn't exist in this document structure
+    if field is None:  # pragma: no cover
+        return
+
+    jexl = QuestionJexl(
+        {"form": field.form, "document": field.document, "structure": field.parent()}
+    )
+
+    # Ignore errors because we might evaluate partially built forms. So we might be
+    # missing some answers or the expression might be invalid, in which case we return
+    # None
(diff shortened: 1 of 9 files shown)


caluma/caluma_form/migrations/0047_recalculate_calc_answers.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    # Ignore errors because we might evaluate partially built forms. So we might be
+    # missing some answers or the expression might be invalid, in which case we return
+    # None
+    value = jexl.evaluate(field.question.calc_expression, raise_on_error=False)
+
+    AnswerModel.objects.update_or_create(
+        question=answer.question, document=field.document, defaults={"value": value}
+    )
```
winged, line 34: You already have the answer object, so why the `update_or_create()` call?

Since this only affects already-calculated answers, you could just write 

```suggestion
    answer.value = value
    answer.save()
```

caluma/caluma_form/tests/test_migration.py
```diff
@@ -429,3 +431,82 @@ def test_migrate_answer_history_question_type(transactional_db):
...
+    calc_answer.save()
+
+    # migrate back and forth in order to run the data-migration
+    executor = MigrationExecutor(connection)
+    app = "caluma_form"
+    migrate_from = [(app, "0046_file_answer_reverse_keys")]
+    migrate_to = [(app, "0047_recalculate_calc_answers")]
+    executor.migrate(migrate_from)
+    executor.loader.build_graph()
+    executor.migrate(migrate_to)
```
winged, line 504: Since you clearly factored out the actual migration code, you should be able to simply call the function instead of doing the actual migration. This would probably also fix the flaky test in the health check again.

winged (changes requested):
Looking good so far, I have some small suggestions though

caluma/caluma_form/migrations/0047_recalculate_calc_answers.py
open-dynaMIX: Done.

caluma/caluma_form/tests/test_migration.py
open-dynaMIX: As discussed in person.

winged (approved):

PR #1965, opened 2023-02-17
fix: replace faulty migration for calc answers with a command

open-dynaMIX:
The migration `0047_recalculate_calc_answers` was error prone and took a long time to execute. This commit replaces it with a management command that is more optimized and stable. Still this operation will take some time to finish.


winged (approved):
Looking good!

Some suggestions (can be done in a follow-up):

* The command's docstring says "all calculated", which is not entirely true. I think it might be a useful tool, but for now, it solves a specific problem. I think we should develop this further as a generic useful tool where you can recalculate some questions when needed, not just fixing this specific bug
* The `post_migrate_to_current_state` fixture is not required in the test anymore

open-dynaMIX:
You are absolutely correct. I fixed it.

PR #1969, opened 2023-02-22
feat(workflow): add case_document_forms filter to work item

Yelinz:


caluma/caluma_workflow/filters.py
```diff
@@ -167,6 +167,7 @@ class WorkItemFilterSet(MetaFilterSet):
 
     document_has_answer = HasAnswerFilter(document_id="document__pk")
     case_document_has_answer = HasAnswerFilter(document_id="case__document__pk")
+    case_document_forms = MultipleChoiceFilter(field_name="case__document__form_id")
```
open-dynaMIX, line 170: For consistency reasons, I'd also add a `case_document_form`-filter as in the `CaseFilterSet`.

open-dynaMIX (changes requested):
Great work :tada:

Just a small addition.

caluma/caluma_workflow/filters.py
Yelinz: Some single value filters are deprecated in favor of their multi value ones.
For example:
https://github.com/projectcaluma/caluma/blob/e351de3b9c4c48e70f529295b13a61cc5052daec/caluma/caluma_workflow/filters.py#L90
So we do continue implementing both single and multi value?

open-dynaMIX: Yes I saw that, but the `document_form`-filter is not deprecated (yet?). What's our strategy here @anehx @winged? We just need the multivalue-filter. But I think we should either deprecate the `document_form`-filter on the `CaseFilterSet` or implement both on the `WorkItemFilterSet`.

winged: I think we should only provide the single-value variant if it really makes sense, which is almost never. I suggest deprecating `CaseFilterSet.document_form` in this case - but maybe in another MR

open-dynaMIX (approved):

PR #1970, opened 2023-03-02
feat(filters): Support word combinations in SearchAnswersFilter

### LINKED ISSUE #1927: [filters] Support word combinations in `SearchAnswersFilter`
In (at least) one of our projects, there is a demand to search for word combinations. So that one could search for `"Foo Bar"` and only receive results that contain both words in the same order.

I've implemented this once in a DRF project and we would need to extract the search terms somewhere along those lines:

```python
    ...
    @staticmethod
    def _split(value):
        lex = shlex.shlex(value, posix=True)
        lex.whitespace_split = True
        lex.commenters = ""
        lex.quotes = '"'
        return [i.strip() for i in lex if i.strip()]

    def get_search_terms(self, request):
        params = request.query_params.get(self.search_param, "")
        params = params.replace("\x00", "")  # strip null characters
        try:
            return self._split(params)
        except ValueError as e:
            if e.args[0] == "No closing quotation":
                return self._split(f'{params}"')
    ...
```

This is potentially breaking, but I fail to see a usecase where it makes sense to search for literal quotation marks.

open-dynaMIX:
This commit implements searching for exact word combinations if words are enclosed in quotations. If the amount of quotation marks is uneven, we implicitly add an ending quotation at the end of the search string.

Closes #1927

### DIFF
--- a/caluma/caluma_form/filters.py
+++ b/caluma/caluma_form/filters.py
@@ -1,3 +1,4 @@
+import shlex
 from functools import reduce
 
 import graphene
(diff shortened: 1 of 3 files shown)


winged (approved):
Awesome work! :tada:

PR #1972, opened 2023-03-03
chore: fix the link for wait-for-it.sh to a known safe commit

StephanH90:
In our Dockerfile we simply fetch `wait-for-it.sh` from master. However, we can't be sure that it hasn't been updated, and malicious code has been injected.

Instead, we should point it to a known safe commit.


open-dynaMIX (changes requested):
`wait-for-it` is in the debian repositories. So we can just `apt-get install wait-for-it`.

PR #1977, opened 2023-03-23
feat(form): order questions when building document structure

fugal-dy:
This allows using the structure in other contexts where the order of elements is important.

e. g. exporting a document to a json object


czosel:
I think @anehx' feedback on this would be important, especially regarding potential performance regressions.

winged:
Closing this one. The newly merged structure code does this correctly

PR #1983, opened 2023-04-13
fix(analytics): field alias length should allow 100 chars

open-dynaMIX:


winged:
We'll probably need to enforce a much shorter alias length than currently allowed (100 chars). See https://stackoverflow.com/questions/8213127/maximum-characters-in-labels-table-names-columns-etc

winged:
> @winged This is tested tho, right?
> 
> https://github.com/projectcaluma/caluma/blob/cdca090e6a8cafbeb3dde5aad29bb1c866dea518/caluma/caluma_analytics/tests/test_simple_table.py#L131

Indeed, you're right! Didn't look too closely at the test.

winged (approved):

PR #1991, opened 2023-05-11
feat(analytics): support accessing sub-cases and their parent

open-dynaMIX:


winged (approved):
Awesome work! :tada: :100:

PR #2006, opened 2023-05-26
feat(auth)!: only use userinfo for auth (drop introspect)

open-dynaMIX:
For OIDC-auth it's not necessary to have a fallback to the introspect endpoint, so this commit drops it. For this to work, make sure to include the `openid`-scope in the request.

BREAKING CHANGE: setting the `OIDC_INTROSPECT_*` and `OIDC_CLIENT_AS_USERNAME` env vars is obsolete now. Auth now only calls the userinfo endpoint without any fallback to the introspect endpoint. This will possibly have an impact on the `username` property of the user object. If your extensions depend on that property, please update them if necessary.
Additionally, all requests sent to Caluma must include the `openid`-scope.

### DIFF
--- a/caluma/caluma_user/models.py
+++ b/caluma/caluma_user/models.py
@@ -20,34 +20,14 @@ class AnonymousUser(BaseUser):
 
 
 class OIDCUser(BaseUser):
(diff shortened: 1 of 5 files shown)


winged (approved):
Nice work, I like! :100:

PR #2034, opened 2023-07-17
chore(ci): add compatibility tests for postgres v15

anehx:


winged (approved):
LGTM! Could we maybe drop v10 (or even v11) at this point, or is someone still using it?

Issue #1699, opened 2022-03-02
Handle failed or omitted file uploads


open-dynaMIX:
When saving a `FileAnswer`, Caluma assumes that the file is uploaded to the bucket afterwards. There are cases where this doesn't happen and Caluma has data about a file that doesn't exist in the bucket.

## Proposal
In [Alexandria](https://github.com/projectcaluma/alexandria/) we let MinIO call an endpoint on our side after completed uploads. There we use it to generate Thumbnails (https://github.com/projectcaluma/alexandria/blob/main/alexandria/core/views.py#L125). The same mechanism could be used in Caluma to know if a file really was uploaded.

### Process

The process would look like this:

1. Client saves `FileAnswer` in Caluma
2. Caluma saves the `File` and marks it as "draft"
3. Client uploads file to the bucket
4. MinIO send a request to Caluma, informing about the newly uploaded file
5. Caluma unmarks the `File` as draft.

Caluma then could treat draft `Files` as non-existent and also ignore the corresponding `FileAnswer`.

PR #2035, opened 2023-07-20
fix(form): Fixes projectcaluma/caluma#1699

maria-pw:


caluma/caluma_core/models.py
```diff
@@ -10,11 +10,10 @@ def _history_user_getter(historical_instance):
...
 def _history_user_setter(historical_instance, user):
     request = getattr(HistoricalRecords.thread, "request", None)
-    user = None
+    user = "AnonymousUser"
     if request is not None:
-        user = request.user.username
-        if request.user.__class__.__name__ == "AnonymousUser":
-            user = "AnonymousUser"
+        if hasattr(request, "user"):
+            user = request.user.username
```
winged, line 16:
```suggestion
        if hasattr(request, "user"):
            user = request.user.username or "AnonymousUser"
```

This should clean up the failing tests, as otherwise, we're now setting the user to `None` in some cases where it was `"AnonymousUser"` before

caluma/caluma_core/views.py
```diff
@@ -44,3 +47,29 @@ def health_check_status(request):
...
+        try:
+            file = File.objects.get(pk=file_pk)
+        except File.DoesNotExist:
+            return HttpResponse(status=HTTP_400_BAD_REQUEST)
+
+        file.is_draft = False
+        file.save()
+        status = HTTP_201_CREATED
+
+    return HttpResponse(status=status)
```
winged, line 75: This should probably be moved to `caluma.caluma_form.views` as it's strictly forms-related. `caluma_core` is mainly infrastructure

caluma/caluma_form/tests/__snapshots__/test_history.ambr
```diff
@@ -28,7 +28,7 @@
           dict({
             'node': dict({
               '__typename': 'HistoricalStringAnswer',
-              'historyUserId': 'AnonymousUser',
+              'historyUserId': None,
```
winged, line 31: This should be fixed (see note in the `_history_user_setter` function)

caluma/caluma_logging/middleware.py
```diff
@@ -22,17 +22,20 @@ def __call__(self, request):
...
+            AccessLog.objects.create(
+                username=request.user.username,
+                query=body.get("query"),
+                variables=body.get("variables"),
+                status_code=response.status_code,
+                has_error=response.status_code >= 400,
+                **vis.values,
+            )
+        except AttributeError:
+            pass
```
winged, line 38: I'd like to see a short comment here explaining why this might fail

winged (changes requested):
Awesome work! Just some small nitpicks and cleanups to do.

Also, as we're using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), you should reword the commit message to use `feat(form) ...` instead of `fix`, as it's a new feature, not a bugfix.

Other than that, great work and thanks for the contribution :-)

PR #2043, opened 2023-08-10
fix(analytics): fix meta field extraction

### LINKED ISSUE #2017: [Analytics] Support displaying multiple meta-values of same object
When trying to display more than one value from the meta field of the same object (e.g. WorkItem), Caluma raises an error. I assume this is similar to the problem we had with multiple date extractions.

/cc @winged

open-dynaMIX:
This commit implements support for using multiple keys from the meta field.

Closes #2017


winged (approved):
:tada: :100:

PR #2047, opened 2023-08-11
feat(core): add meta intersection filter

Yelinz:
Motivation: Just like Answers, filtering with an intersection in the meta is useful

### DIFF
--- a/caluma/caluma_core/filters.py
+++ b/caluma/caluma_core/filters.py
@@ -6,6 +6,7 @@ from django import forms
 from django.contrib.postgres.fields.hstore import KeyTransform
 from django.contrib.postgres.search import SearchVector
 from django.db import models
+from django.db.models import Q
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.expressions import OrderBy
 from django.db.models.fields.json import KeyTextTransform
@@ -378,6 +379,7 @@ class JSONLookupMode(Enum):
     CONTAINS = "contains"
     ICONTAINS = "icontains"
     IN = "in"
+    INTERSECTS = "intersects"
     GTE = "gte"
     GT = "gt"
     LTE = "lte"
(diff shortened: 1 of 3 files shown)


czosel (approved):
:+1: LGTM

caluma/caluma_core/filters.py
```diff
@@ -425,10 +427,19 @@ def filter(self, qs, value):
                     ),
                 )
                 lookup = {f"field_val__{lookup_expr}": expr["value"]}
+            elif lookup_expr == JSONLookupMode.INTERSECTS:
+                lookup = {}
+                exprs = [
+                    Q(**{f"{self.field_name}__{expr['key']}__contains": val})
```
anehx, line 433: Did you check whether we need the transformation above (L422-L429) or not? Maybe add a test using strings as values instead of integers.

caluma/caluma_core/filters.py
```diff
@@ -425,10 +427,19 @@ def filter(self, qs, value):
...
+                    for val in expr["value"]
+                ]
+                # connect all expressions with OR
+                qs = qs.filter(reduce(lambda a, b: a | b, exprs))
```
winged, line 437: In this branch, you're updating `qs` directly. The other two branches set the `lookup` dict, and `qs.filter()` runs after.

You should probably move the `qs.filter()` call inside the branch in every case, so it's not called twice (See Line 443 below)

winged (changes requested):
A small inconsistency to clean up, otherwise LGTM!

caluma/caluma_core/filters.py
anehx: Nice! In that case, could you shortly check if we even need the transform above anymore?

Yelinz: Tried commenting out the transform and it broke the contains, seems still necessary

winged (approved):
LGTM now, nice work!

PR #2057, opened 2023-08-31
fix(migrations): make app prefix migration language independent

anehx:
Currently, the migration to add prefixes failed on initial migrations if the database server is not in English. This commit removes the dependency on the error message and instead uses the underlying psycopg exception which is more specific.


open-dynaMIX (approved):
:100:

PR #2069, opened 2023-09-18
fix(healthz): do not warn about expected events

open-dynaMIX:
As part of the health-check routines, a file is removed in minio and afterwards it's asserted that the file is gone. This call to `Minio.stat_object()` led to a warning being issued during every health-check iteration like the following:

```
Minio error, cannot stat object 'healthz_tmp-object_{UUID}': NoSuchKey
```

This commit extends `Minio.stat_object()`, so that issuing of the log-warning can be suppressed by profiding the argument `warn_missing=False`. Default behaviour doesn't change and will still issue this warning.

### DIFF
--- a/caluma/caluma_core/health_checks.py
+++ b/caluma/caluma_core/health_checks.py
@@ -58,7 +58,7 @@ def _check_media_storage_service():
 
         # remove object
         storage_client.remove_object(object_name)
-        assert not storage_client.stat_object(object_name)
+        assert not storage_client.stat_object(object_name, suppress_warning=True)
 
         return {"ok": True}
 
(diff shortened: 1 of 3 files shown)


winged (approved):
Code LGTM, but honestly I feel like our health checks do a bit too much work. I think we should check connectivity, but not whether we can actually store data (same as with the DB as well). 

That's something for another PR however :-)

winged:
On a another note, I'd kinda like the flag to be inverted.. something like `suppress_warning=False` in the parameters. But that's just my personal taste, so feel free to ignore

open-dynaMIX:
> Code LGTM, but honestly I feel like our health checks do a bit too much work. I think we should check connectivity, but not whether we can actually store data (same as with the DB as well).
> 
> That's something for another PR however :-)

YES YES YES YES!!!

open-dynaMIX:
> On a another note, I'd kinda like the flag to be inverted.. something like `suppress_warning=False` in the parameters. But that's just my personal taste, so feel free to ignore

Done.

PR #2071, opened 2023-09-20
feat(healthz)!: overhaul health-checks for them to be less intrusive

open-dynaMIX:
This commit contains no changes. Its only purpose is to mark the commit `54c545f3d484661f8c9b29b1ce20384b8df5bfc5` as breaking, because this was forgotten.

BREAKING CHANGE: In the healthz response, the key `database models` has been dropped.


winged (approved):
:D :tada:

PR #2077, opened 2023-09-27
chore: add search answer lookup test (wip)

czosel:


caluma/caluma_form/tests/test_search_answers.py
```diff
@@ -94,6 +94,49 @@ def _search(q_slugs, f_slugs, word, expect_count):
...
 
 
+@pytest.mark.parametrize(
+    "answer_value,search_text,lookup,expect_count",
+    [
+        # ("foo", "foo", "EXACT_WORD", 1),
+        # ("foo", "fo", "EXACT_WORD", 0),
+        # ("foo", "oo", "CONTAINS", 1),
+        # ("foo", "bar", "CONTAINS", 1),
+        ("foo", '"fo', "STARTSWITH", 1),
```
winged, line 104: This is how it works now, but it's wrong. I think we'll need something like `KeyTextTransform` in the filter to correctly extract it. See the `JSONValueFilter.filter()` method to see an application of it

PR #2081, opened 2023-10-06
Define date range of the date questions

MitanOmar:


winged:
@MitanOmar is this PR ready to merge? Are you gonna add validation for the date range in the answers as well, or is this intended more as a range limiter for the input field?

caluma/caluma_form/serializers.py
```diff
@@ -240,12 +240,28 @@ class Meta(SaveQuestionSerializer.Meta):
 
 
 class SaveDateQuestionSerializer(SaveQuestionSerializer):
+    min_date = DateField("%Y-%m-%d", required=False, allow_null=True)
+    max_date = DateField("%Y-%m-%d", required=False, allow_null=True)
+
     def validate(self, data):
```
MitanOmar, line 246: @winged there is a validation in the backend, but since i finish this PR, it's waiting for review

MitanOmar: https://github.com/projectcaluma/ember-caluma/pull/2521

here is the implementation of the frontend side

winged: Right, but I think if we do this, we should add validation to the *answers* as well. This only validates the question, but no restriction happens when people actually fill out the forms.

Look at [the answer validator](https://github.com/projectcaluma/caluma/blob/1a47a86868a252951d3ec9e66ea94bc63f298f42/caluma/caluma_form/validators.py#L84C9-L84C32) to see what I mean.

There's already `_validate_question_float()` and `_validate_question_integer()` that implement the same pattern. I think therefore, you should extend `_validate_question_date()` to match.

MitanOmar: that's right, i will implement that too, and i will include it in my todo for this week

winged:
As discussed today: Since this doesn't stem from a direct (client) requirement, we will stop development for now, and pick it back up when we do have a need for this feature. Otherwise, we may have a breaking change to modify this in a way that actually benefits the user.

Ideas that may be useful in this context, that aren't possible with the current approach:

* Set upper or lower limit to "today" or "right now"
* Set upper or lower limit to the value of another question (potentially with an offset in either direction)

These limits may possibly also be interesting for float / integer questions as well, but as we're not clear on any actual requirements, let's not run off in the wrong direction here.

PR #2088, opened 2023-10-17
feat: implement flat_answer_map

StephanH90:
# Description

Adds a `flat_answer_map` property to a document which can be used in the same was as we do it in `ember_caluma`.

### DIFF
--- a/caluma/caluma_form/models.py
+++ b/caluma/caluma_form/models.py
@@ -334,6 +334,40 @@ class Document(core_models.UUIDModel):
     )
     meta = models.JSONField(default=dict)
 
+    def flat_answer_map(self):
+        """
+        Return a dictionary with the flattened answer map for this document.
+
+        The keys are the question IDs, and the values are either the answer value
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/models.py
```diff
@@ -334,6 +334,19 @@ class Document(core_models.UUIDModel):
     )
     meta = models.JSONField(default=dict)
 
+    @property
+    def flat_answer_map(self):
```
winged, line 338: I'd rather have this as a method, not a `@property`. Just to make clear to the caller that this *does* stuff. Also, because multiple calls to the property will re-evaluate the code.

And a docstring would be nice too

winged (changes requested):
Nice work, just like it to be a method, not a property

caluma/caluma_form/models.py
StephanH90: done

winged (approved):
Nice!

PR #2108, opened 2023-12-14
fix: only save updated fields in post_complete case logic

kaldras:
This prevents untouched fields of the case model which were updated in a event to be overwritten with an old value.


anehx (approved):

luytena (approved):

luytena:
:partying_face:

PR #2135, opened 2024-01-30
chore: drop deprecated id filter on cases filterset

winged:
BREAKING CHANGE: you now must use the `ids` filter instead

This should go in before the 10.0.0 release


open-dynaMIX (approved):
Finally :sweat_smile:

PR #2138, opened 2024-01-31
feat(settings): add setting to limit query depth

winged:
This should avoid overly comples GraphQL queries which might overload the backend. Note: We do not yet have experience with the exact useful value of this settings, so the default is zero (which implies the limit is disabled)

Edit: Ran some manual tests (It's hard to do automated tests for stuff like that). The setting works, but I'm still not sure what a reasonable limit would be. I assume it will be around 15-20 realistically

image


derrabauke (approved):
LGTM :+1:

winged:
Let's wait with this until 10.0.0 is released and tagged (just to avoid having to update/redo the changelog etc)

PR #2148, opened 2024-02-05
fix: defer calculation of calc answers until after creating a new doc

open-dynaMIX:


caluma/caluma_form/api.py
```diff
@@ -92,3 +93,62 @@ def copy_form(
...
+    )
+
+
+def recalculate_answers_from_document(instance):
+    if instance.family.meta.get("_defer_calculation"):
+        return
+    for question in models.Form.get_all_questions(
+        [(instance.family or instance).form_id]
+    ).filter(type=models.Question.TYPE_CALCULATED_FLOAT):
+        update_or_create_calc_answer(question, instance)
```
winged, line 154: I think we should split the PR into the refactoring bit and the actual bugfix bit, so make it two commits. That way, we won't have any confusion when we're digging for bugs in the future

Also, I think we don't really need to move the `update_calc_dependents()` function, thus avoiding the circular import here.

winged (changes requested):
LGTM, was a nice pairing session! I'd split out the commits as noted tho

caluma/caluma_form/api.py
open-dynaMIX: > Also, I think we don't really need to move the update_calc_dependents() function, thus avoiding the circular import here.

The problem isn't `update_calc_dependents()`, but the import from `caluma.caluma_form.api` inside `caluma.caluma_form.domain_logic`. I've solved it with a delayed import. Will split the commit now and the make CI pass.

winged (approved):
:rocket:

PR #2155, opened 2024-02-08
fix(validation): do not block __typename when introspection is disabled

winged:
The `DisableIntrospection` validator rejects everything that could lead to insight into the schema. Sadly, our frontends rely on having `__typename` available, thus we need our own validator that allows this specific introspection key (but not anything else)


derrabauke (approved):
LGTM :rocket:

winged:
In addition to the unit test, performed some manual tests as well:

* Set DISABLE_INTROSPECTION to true
* Ran manual GQL with __schema -> rejected :heavy_check_mark: 
* Ran manual GQL with __typename -> accepted :heavy_check_mark: 
* Did some queries and mutations in ember-caluma (using __typename in it's queries) -> works :heavy_check_mark:

anehx (approved):

PR #2204, opened 2024-04-24
feat(jexl): add main_case_form to info object

czosel:
This is convenient when you'd like to write a JEXL expression in a task form attached to some work item, that depends on the case's main documents' form.

Related PR in ember-caluma: https://github.com/projectcaluma/ember-caluma/pull/2703

Once we merge this, we should spend a minute to update the caluma docs: https://caluma.gitbook.io/caluma-docs/forms/validation#variables-available-for-evaluation

### DIFF
--- a/caluma/caluma_form/structure.py
+++ b/caluma/caluma_form/structure.py
@@ -134,7 +134,7 @@ class RowField(Field):
 
 
 class FieldSet(Element):
-    aliases = {"formMeta": "form_meta"}
+    aliases = {"formMeta": "form_meta", "mainCaseForm": "main_case_form"}
 
     def __init__(self, document, form, question=None, parent=None):
         super().__init__(parent)
@@ -144,6 +144,18 @@ class FieldSet(Element):
         self.question = question
         self._fields = None
         self._sub_forms = None
+        self._main_case_form = "NOTSET"
+
+    @property
+    def main_case_form(self):
+        if self._main_case_form == "NOTSET":
+            try:
+                self._main_case_form = (
+                    self.document.family.work_item.case.family.document.form.slug
+                )
+            except Exception:  # pragma: no cover
+                self._main_case_form = None
+        return self._main_case_form
 
     @property
     def fields(self):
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/structure.py
```diff
@@ -144,6 +144,18 @@ def __init__(self, document, form, question=None, parent=None):
...
         self._fields = None
         self._sub_forms = None
+        self._main_form = None
+
+    @property
+    def main_form(self):
+        if self._main_form is None:
+            if hasattr(self.document.family, "work_item") and hasattr(
+                self.document.family.work_item.case.family, "document"
+            ):
```
czosel, line 154: there's probably a cleaner way to do this in python?

caluma/caluma_form/structure.py
```diff
@@ -144,6 +144,18 @@ def __init__(self, document, form, question=None, parent=None):
...
+                self._main_form = (
+                    self.document.family.work_item.case.family.document.form.slug
+                )
+        return self._main_form
```
winged, line 158: Two things: One, naming: I think it might be confusing, as the "main form" could also be the the main form of the current document, for example if you're in a row-doc or similar. Suggestion: "main_case_form" or something similar maybe?

Second: If for some reason the inner `if` does not find the main form, it will not cache the result in `self._main_form` and try to do the whole resolution-chain again, possibly triggering quite a few DB queries. I'd suggest caching that case as well

winged (changes requested):
Nice work, just some minor issues found

caluma/caluma_form/structure.py
winged: There's [operator.attrgetter](https://docs.python.org/3/library/operator.html#operator.attrgetter), but it doesn't make things much shorter.

Since the default/fallback is to just have *no value", you could just as well fetch the whole attribute chain and just try/except around it

czosel: :+1: done

caluma/caluma_form/structure.py
czosel: Fixed the naming and caching :+1:

winged (approved):
LGTM!

PR #2205, opened 2024-04-25
feat(form): add more case info to jexl context

czosel:
As suggested in https://github.com/projectcaluma/ember-caluma/pull/2703#discussion_r1579985434

### DIFF
--- a/caluma/caluma_form/structure.py
+++ b/caluma/caluma_form/structure.py
@@ -49,6 +49,8 @@ class Element:
             out = out()
         if isinstance(out, Element) or isinstance(out, dict):
             return out
+        if out is None:
+            return None
         return str(out)
 
 
@@ -134,7 +136,7 @@ class RowField(Field):
 
 
 class FieldSet(Element):
-    aliases = {"formMeta": "form_meta", "mainCaseForm": "main_case_form"}
+    aliases = {"formMeta": "form_meta"}
 
     def __init__(self, document, form, question=None, parent=None):
         super().__init__(parent)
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/structure.py
```diff
@@ -144,18 +150,18 @@ def __init__(self, document, form, question=None, parent=None):
...
-        if self._main_case_form == "NOTSET":
+    def case(self):
+        if self._case == "NOTSET":
             try:
-                self._main_case_form = (
-                    self.document.family.work_item.case.family.document.form.slug
-                )
+                case = self.document.family.work_item.case
+                parent = case.parent_work_item.case if case.parent_work_item else None
+                self._case = Case(case, parent=Case(parent))
```
czosel, line 161: @winged For some reason, the `parent` weakref was dead when trying to read it, which made some of the tests fail. I removed the weakref abstraction just to get the tests green, but I'm sure that's not what we actually want :-)

winged: Yeah we need to find another solution for that. Making `parent` a regular ref turns the structure into a reference loop that the GC can't collect, causing memory leaks.

For the `case` here though, I think we could get away with it, as the `Case` element does not present access to the form, so is not circular

caluma/caluma_form/structure.py
```diff
@@ -144,18 +144,38 @@ def __init__(self, document, form, question=None, parent=None):
         self.question = question
         self._fields = None
         self._sub_forms = None
-        self._main_case_form = "NOTSET"
+        self._case = "NOTSET"
+        self._parent_case = None
```
winged, line 148:
```suggestion
```

That line can go

caluma/caluma_form/structure.py
czosel: Fixed in pair programming session, thanks again @winged

winged (changes requested):
LGTM, just a tiny left-over line

caluma/caluma_form/structure.py
czosel: fixed

caluma/caluma_form/structure.py
```diff
@@ -49,6 +49,8 @@ def get(self, name, default=None):
             out = out()
         if isinstance(out, Element) or isinstance(out, dict):
             return out
+        if out is None:
+            return None
```
czosel, line 53: @winged I still added coverage for the "disconnected document" case - I had to add this line to make it pass, because otherwise the returned value is the string `"None"` instead of `null`.

winged: Ah yes makes perfect sense

PR #2220, opened 2024-05-31
feat(options): implement is_hidden jexl on options

open-dynaMIX:
This commit adds an `is_hidden` jexl to Options. This nwill be evaluated and enforced on saving of answers. Addiotionally, the Options got a new filter `visible_in_document`.


caluma/caluma_form/filters.py
```diff
@@ -64,8 +65,67 @@ class Meta:
         fields = ("meta", "attribute")
 
 
+class VisibleOptionFilter(Filter):
+    """
+    Lookup type display options in the context of a document.
```
winged, line 70: That sentence reads a bit weird.. maybe rephrase?

```suggestion
    Filter options to only show ones whose `is_hidden` JEXL evaluates to false
```

caluma/caluma_form/filters.py
```diff
@@ -64,8 +65,67 @@ class Meta:
...
+
+    This will make sure all the `is_hidden`-JEXLs on the options are evaluated in the
+    context of the provided document.
+
+    Note:
+    This filter can only be used if the options in the QuerySet all belong only to
+    one single question. Generally forms are built that way, but theoretically,
+    options could be shared between questions. In that case it will throw a
+    `ValidationError`.
+
```
winged, line 80:
```suggestion
    Note:
    This filter can only be used if the options in the QuerySet all belong only to
    one single question. Generally forms are built that way, but theoretically,
    options could be shared between questions. In that case it will throw a
    `ValidationError`.

    Also note that this evaluates JEXL for all the options of the question, which
    has a good bit of performance impact.
```

caluma/caluma_form/validators.py
```diff
@@ -87,17 +89,63 @@ def _validate_question_date(self, question, value, **kwargs):
...
+
+        options = qs if qs else question.options.all()
+
+        # short circuit if none of the options has an `is_hidden`-jexl
+        if not options.exclude(
+            Q(is_hidden="false") | Q(is_hidden__isnull=True)
+        ).exists():
+            return [o.slug for o in options]
+
+        validation_context = validation_context or _validation_context(document)
```
winged, line 117: Nice!

winged (changes requested):
Awesome work! I have some suggestions regarding the filter docstring, but other than that, LGTM! Very nice

caluma/caluma_form/filters.py
open-dynaMIX: Oh, good catch, I was in the middle of editing this sentence. It was still a mix from when I experimented with the FilterTypes ;)

Fixed.

caluma/caluma_form/filters.py
open-dynaMIX: Done :+1:

winged (approved):

PR #2250, opened 2024-07-15
feat: allow calculated fields to depend on other calculated fields

czosel:


open-dynaMIX:
Interesting. That was something a customer asked for in one of our projects, but we ended up not doing it.

How do you make sure they are calculated in the right order?

Let's say we have Questions `A` and `B`. Then we have a calc question `calc-1` with the expression (simplified) `A + B`.  Another calc question `calc-2` has following expression: `calc-1 + 5`.

If we now change the answer to question `A`, the recalculation only works, if we make sure `calc-1` is evaluated first. If `calc-2` is evaluated first, we will end up with a wrong answer for it.

How many levels of referencing calc-question should be supported? -> calc question that depends on calc-question that depends on calc-question ...
What about circular dependencies? :smiling_imp:

czosel:
As discussed: Since calculation is implemented with forward references (`calc_dependents`) and signals, the order should always be correct. Circular references are indeed an issue, but he have the same problem already since a long time with the `hidden` JEXLs (since hidden question's receive a `null` value when executing JEXL, this is the same issue as with calculated questions) - and this has not been an issue in practice yet. If we decide to catch circular references in the future, that is possible (and it would probably make sense to apply the check to hidden expressions at the same time).

@open-dynaMIX or @winged ready for review :slightly_smiling_face:

winged (approved):
Nice work!

open-dynaMIX (approved):

PR #2256, opened 2024-07-18
Fix deprecations

Yelinz:
update dependencies and fix deprecations

use newer ruff version

### DIFF
--- a/caluma/caluma_analytics/tests/test_migrate_form_slugs.py
+++ b/caluma/caluma_analytics/tests/test_migrate_form_slugs.py
@@ -48,7 +48,7 @@ def test_migrate_rename_form_slugs(post_migrate_to_current_state):
         # `form_id` data source anymore.
         runner.get_records()
     assert exc_info.value.args == ("form_id",)
-    assert type(exc_info.value) == KeyError
+    assert type(exc_info.value) is KeyError
 
     # Migrate forwards.
     executor.loader.build_graph()  # reload.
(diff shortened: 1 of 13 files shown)


caluma/caluma_analytics/tests/test_migrate_form_slugs.py
```diff
@@ -48,7 +48,7 @@ def test_migrate_rename_form_slugs(post_migrate_to_current_state):
         # `form_id` data source anymore.
         runner.get_records()
     assert exc_info.value.args == ("form_id",)
-    assert type(exc_info.value) == KeyError
+    assert type(exc_info.value) is KeyError
```
winged, line 51:
```suggestion
    assert isinstance(exc_info.value, KeyError)
```

that would be more correct IMO

caluma/caluma_form/factories.py
```diff
@@ -196,9 +196,11 @@ def files(self, create, extracted, **kwargs):
                 self.files.set(
                     FileFactory.create_batch(num_files, **kwargs, answer=self)
                 )
+            self.save()
```
winged, line 199: Is this really needed? `self.files` is a M:N relationship, and the intermediate table should be written directly.

caluma/caluma_core/tests/test_mutation.py
```diff
@@ -77,7 +77,7 @@ class Meta:
             serializer_class = MySerializer
 
     result = NoModelMutation.mutate_and_get_payload(None, info, name="test")
-    assert type(result) == NoModelMutation
+    assert type(result) is NoModelMutation
```
winged, line 80: Same here again, unless you specifically don't want to allow any subtypes you can use `isinstance()` instead

winged (changes requested):
Nice work, LGTM, just some minor suggestions

PR #2269, opened 2024-08-15
fix: add missing "is_hidden" to SaveOptionSerializer

StephanH90:


winged (approved):
LGTM, probably will need a snapshot update on the schema though

PR #2272, opened 2024-08-16
chore: migrate from uwsgi to gunicorn

c0rydoras:


Dockerfile
```diff
@@ -34,4 +33,4 @@ COPY . $APP_HOME
 
 EXPOSE 8000
 
-CMD /bin/sh -c "wait-for-it $DATABASE_HOST:${DATABASE_PORT:-5432} -- poetry run python manage.py migrate && poetry run uwsgi"
+CMD /bin/sh -c "wait-for-it $DATABASE_HOST:${DATABASE_PORT:-5432} -- poetry run python manage.py migrate && poetry run gunicorn --bind :8000 caluma.wsgi"
```
derrabauke, line 36:
```suggestion
CMD /bin/sh -c "wait-for-it $DATABASE_HOST:${DATABASE_PORT:-5432} -- poetry run python manage.py migrate && poetry run gunicorn --workers 10 --access-logfile - --limit-request-line 16384 --bind :8000 caluma.wsgi"
```
Let's spin up at least 10 workers. The old config had 16, but [we did the same in mySAGW](https://github.com/adfinis/mySAGW/commit/7f5c73d50b6f428d1663523a3ab1055271b6f136)

Further we should increase the request size limit, in case we have large listing in the query string (we did that on all our caluma projects).

derrabauke (changes requested):
Looks good so far to me. But let's have a second opinion from @open-dynaMIX we he's back.

derrabauke:
While we are at it already, could you fix the naming of the project in the comment, please? :-) 
https://github.com/projectcaluma/caluma/blob/main/caluma/wsgi.py#L2

pyproject.toml
```diff
@@ -40,7 +40,7 @@ pyjexl = "^0.3.0"
 python-memcached = "^1.62"
 requests = "^2.31.0"
 urllib3 = ">=1.26.8,<3.0.0"
-uWSGI = "^2.0.26"
+gunicorn = "^23.0.0"
```
open-dynaMIX, line 43: The dependencies should be in alphabetical order (expect for `python`).

open-dynaMIX (changes requested):
Awesome work :tada: 

Just a veeeeery small nitpick :see_no_evil:

pyproject.toml
c0rydoras: done

open-dynaMIX (approved):

PR #2278, opened 2024-08-30
feat(jexl): add length transform

anehx:
Needs https://github.com/projectcaluma/ember-caluma/pull/2784


czosel:
LGTM - just an open question concerning naming: should we call it `length`, to make it more aligned with JS? (JEXL = JavaScript expression language after all :wink:)

czosel (approved):

PR #2295, opened 2024-09-26
fix(form: speed up copying of documents

Yelinz:
Instead of triggering calculations every time a table answer is copied, only once at the end then the calculation is run.


winged (approved):
nice catch, lgtm!

PR #2309, opened 2024-10-16
feat: use uuid7 instead of uuid4 from now on

winged:
UUID7 has the nice property of being semi-sequential, so it will be much nicer on DB indexes, and following that, improve query performance as well

Note we don't fully replace UUID4 everywhere: In tests, we don't care too much, so we don't build a UUID7 faker for those IDs, and in the Analytics module, we use UUID4 as a simple "random string" utility for generated identifiers (and using only the left few chars of it would break when using UUID7).


winged:
Further information and reasoning for UUID7 can be found here https://uuid7.com/

nlzet (commented):
LGTM, only have to fix the migration conflict imo.

(And I added one minor comment for a todo that can be removed)

anehx (approved):

PR #2319, opened 2024-10-28
fix: update (and pin) graphene dependency

winged:
Graphene 3.4.0 had a bug where UTC dates with the `Z` suffix wouldn't be parsed and were rejected. This was fixed in Graphene 3.4.1, so instead of relying on it as a transitive dependency, we now pin it explicitly.


winged:
Rel: https://github.com/graphql-python/graphene/issues/1580

czosel (approved):

PR #2338, opened 2024-12-02
feat(workflow): add indices to created_at and deadline of work item

StephanH90:
adds 2 indices to the work item which should be useful for most users of caluma.


winged (approved):
LGTM!

PR #2339, opened 2024-12-02
fix(form): optimize calculated question performance

czosel:
- drop several signals in favor of explicit calls in domain logic to allow reuse of form structure
- add prefetching for document structure
- apply topological sorting before calculating answers on document creation


luytena:
@czosel Note: Instance creation is still slow, and seems to be triggered by create in SaveDocumentLogic. I'm having a look.

luytena:
Form changes (e.g. through form-builder) are probably also still slow.

caluma/caluma_form/signals.py
```diff
@@ -93,67 +85,3 @@ def remove_calc_dependents(sender, instance, **kwargs):
...
-
-
-@receiver(post_save, sender=models.FormQuestion)
-@disable_raw
-@filter_events(
-    lambda instance: instance.question.type == models.Question.TYPE_CALCULATED_FLOAT
-)
-def update_calc_from_form_question(sender, instance, created, **kwargs):
-    for document in instance.form.documents.all():
-        update_or_create_calc_answer(instance.question, document)
```
czosel, line 122: I dropped these two handlers completely. On my local instance, updating a calculated question took about 10s, even though I only had 10 documents in the DB. Since in typical prod scenarios we have thousands of documents, this kind of thing should be done, if needed, explicitly in a management command.

caluma/caluma_form/tests/test_question.py
```diff
@@ -1109,34 +1106,6 @@ def test_recursive_calculated_question(
...
-        question__calc_expression="'sub_question'|answer + 1",
-    ).question
-
-    calc_ans = document.answers.get(question_id="calc_question")
-    assert calc_ans.value == 101
-
-    calc_question.calc_expression = "'sub_question'|answer -1"
-    calc_question.save()
-    calc_ans.refresh_from_db()
-    assert calc_ans.value == 99
```
czosel, line 1137: This functionality was dropped. When updating the calc expression of a calculated question, the answers are not updated automatically anymore.

caluma/caluma_form/tests/test_document.py
```diff
@@ -511,269 +509,35 @@ def test_save_document(
...
+        (Question.TYPE_DYNAMIC_MULTIPLE_CHOICE, {}, "MyDataSource", [], ["not in data"], None, "SaveDocumentStringAnswer", False),
+        (Question.TYPE_DYNAMIC_CHOICE, {}, "MyDataSource", [], "5.5", None, "SaveDocumentStringAnswer", True),
+        (Question.TYPE_DYNAMIC_CHOICE, {}, "MyDataSource", [], "not in data", None, "SaveDocumentStringAnswer", False),
+        (Question.TYPE_TEXT, {}, None, ["email"], "some text", None, "SaveDocumentStringAnswer", False),
+        (Question.TYPE_TEXT, {}, None, ["email"], "test@example.com", None, "SaveDocumentStringAnswer", True),
+        (Question.TYPE_TEXTAREA, {}, None, ["email"], "some text", None, "SaveDocumentStringAnswer", False),
+        (Question.TYPE_TEXTAREA, {}, None, ["email"], "test@example.com", None, "SaveDocumentStringAnswer", True),
     ],
-)
+)  # fmt:skip
```
czosel, line 540: drive-by fix: exclude this block from autoformatting so that all parametrizations can stay one one line

caluma/caluma_form/tests/test_question.py
winged: Not sure this is a good idea though. Sure shouldn't happen when the form is already being used, but having data lying around that doesn't actually reflect the calculated result is kinda problematic IMHO.

I think we seriously need to consider doing more background processing (Celery) for this kind of stuff though, as it's clear we can't do more and more work in a still very "sync" fashion.

caluma/caluma_form/tests/test_question.py
```diff
@@ -1220,3 +1198,25 @@ def test_save_action_button_question(db, snapshot, question, schema_executor):
...
+
+    form_question_factory(
+        form=form,
+        question__slug="calc_question",
+        question__type=models.Question.TYPE_CALCULATED_FLOAT,
+        question__calc_expression="'table'|answer|mapby('column')|sum + 'top_question'|answer + 'sub_question'|answer",
+    )
+
+    with django_assert_num_queries(38):
+        api.save_answer(questions_dict["top_question"], document, value="1")
```
winged, line 1222: could we instead snapshot the queries? With just the number of queries is just too easy to overlook/ignore small changes (and update the test "just to make it green again")

caluma/caluma_form/utils.py
```diff
@@ -1,7 +1,79 @@
...
 from caluma.caluma_form import models, structure
 from caluma.caluma_form.jexl import QuestionJexl
 
 
+def build_document_prefetch_statements(prefix="", prefetch_options=False):
+    """Build needed prefetch statements to performantly fetch a document.
+
+    This is needed to reduce the query count when almost all the form data
+    is needed for a given document, e.g. when recalculating calculated
+    answers.
```
winged, line 12: Docstring needs an example I think.

Also I feel like this is not quite the right abstraction level just yet. Callers still have quite a bit of work to do to get it right

caluma/caluma_form/domain_logic.py
```diff
@@ -151,6 +152,26 @@ def post_save(answer: models.Answer) -> models.Answer:
...
+
+        root_doc = answer.document.family
+        root_doc = (
+            models.Document.objects.filter(pk=answer.document.family_id)
+            .prefetch_related(
+                *utils.build_document_prefetch_statements(prefetch_options=True)
+            )
+            .first()
+        )
+        struc = structure.FieldSet(root_doc, root_doc.form)
```
winged, line 168: How about we refactor this dance into the fieldset itself? So this becomes something along the lines of

```py
struc = structure.FieldSet.with_prefetch(root_doc, root_doc.form)
```

The implementation could then be something like

```py
class FieldSet(...):
    ...
    @classmethod
    def with_prefetch(cls, root_doc, form=None):
        root_doc = (
            models.Document.objects.filter(pk=answer.document.family_id)
            .prefetch_related(
                *utils.build_document_prefetch_statements(prefetch_options=True)
            )
            .first()
        )
        return cls(root_doc,form)
```

caluma/caluma_form/domain_logic.py
```diff
@@ -276,7 +301,42 @@ def create(
...
+
+        # fetch all related questions in one query, but iterate according
+        # to pre-established sorting
+        _questions = models.Question.objects.in_bulk(sorted_question_slugs)
+        for slug in sorted_question_slugs:
+            print("question", slug)
+            update_or_create_calc_answer(
+                _questions[slug], document, struc, update_dependents=False
+            )
+
```
winged, line 339: I'd like to have this factored out into it's own (private if needed) method

winged (changes requested):
Nice work so far! Not sure we're all the way there yet tho :-)

caluma/caluma_form/signals.py
open-dynaMIX: What's the setup for those calc-questions for it to take so long? We use this in SAGW, where the users edit their forms with the FormBuilder. So it's actually needed to be recalculated without involving devs.
As a test, I've created 36 documents with a calc question. I've created around 30 documents. Updating the calc-question now in the FormBuilder takes ~500ms.

I totally agree, that this kind of work should be defered to Celery or similar and I would much prefer this over just dropping it. If there is enough reason to drop this from core, I could still implement those handlers in the project they're needed tho.

czosel: @open-dynaMIX we discussed this yesterday and arrived at the same conclusion - we'll keep the feature as is for now (I'll make the changes soon). Probably the time taken per document depends on the number of dependend calculated questions as well. Either way, once you have several thousands of documents, things will blow up without Celery :upside_down_face:

caluma/caluma_form/tests/test_question.py
czosel: reverted as discussed

caluma/caluma_form/tests/test_question.py
czosel: done

caluma/caluma_form/utils.py
czosel: I tried to add an example and changed the callsite

caluma/caluma_form/domain_logic.py
czosel: Changed the callsite to `prefetch_document` instead, as discussed.

caluma/caluma_form/domain_logic.py
czosel: done

winged (changes requested):
Nice work so far, just a few suggestions

caluma/caluma_form/tests/test_question.py
czosel: Turns out this is not feasible (unstable UUIDs and SAVEPOINTS) - reverted

caluma/caluma_form/signals.py
open-dynaMIX: Awesome! Totally agree :+1:

caluma/caluma_form/utils.py
```diff
@@ -1,7 +1,87 @@
...
 from caluma.caluma_form import models, structure
 from caluma.caluma_form.jexl import QuestionJexl
 
 
+def prefetch_document(document_id):
+    return (
+        models.Document.objects.filter(pk=document_id)
+        .prefetch_related(*build_document_prefetch_statements(prefetch_options=True))
+        .first()
+    )
```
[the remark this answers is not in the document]
czosel: moved the docstring up

caluma/caluma_form/signals.py
```diff
@@ -105,8 +102,8 @@ def remove_calc_dependents(sender, instance, **kwargs):
 @disable_raw
 @filter_events(lambda instance: instance.type == models.Question.TYPE_CALCULATED_FLOAT)
 def update_calc_from_question(sender, instance, created, update_fields, **kwargs):
-    # TODO optimize: only update answers if calc_expression is updated
-    # needs to happen during save() or __init__()
+    if not instance.calc_expression_changed:
+        return
```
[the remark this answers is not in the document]
czosel: done

caluma/caluma_form/signals.py
```diff
@@ -104,10 +101,8 @@ def remove_calc_dependents(sender, instance, **kwargs):
 @receiver(post_save, sender=models.Question)
 @disable_raw
 @filter_events(lambda instance: instance.type == models.Question.TYPE_CALCULATED_FLOAT)
+@filter_events(lambda instance: getattr(instance, "calc_expression_changed"))
```
winged, line 104:
```suggestion
@filter_events(lambda instance: getattr(instance, "calc_expression_changed", False))
```

czosel:
@winged I think we should squash this on merge. I enabled squashing in the repo settings, just not sure if it'll mess up the commit message.

winged (approved):

PR #2352, opened 2025-01-09
chore: version bump

winged:


winged:
self-merging, just a changelog/version update

PR #2362, opened 2025-01-20
feat(float): add step configuration

velrest:


caluma/caluma_form/serializers.py
```diff
@@ -335,6 +335,7 @@ class Meta(SaveQuestionSerializer.Meta):
 class SaveFloatQuestionSerializer(SaveQuestionSerializer):
     min_value = FloatField(required=False, allow_null=True)
     max_value = FloatField(required=False, allow_null=True)
+    step = FloatField(required=False, allow_null=True)
 
     def validate(self, data):
```
luytena, line 340: As discussed: Let's add a validation to prevent negative steps

caluma/caluma_form/tests/__snapshots__/test_question.ambr
```diff
@@ -245,6 +245,7 @@
             '__typename': 'FloatQuestion',
             'floatMaxValue': 1.0,
             'floatMinValue': 0.0,
+            'floatStep': None,
```
luytena, line 248: Can we add a test case that includes a value?

velrest: Done

caluma/caluma_form/tests/__snapshots__/test_question.ambr
```diff
@@ -881,6 +883,27 @@
...
+        'defaultAnswer': dict({
+          'value': 0.3,
+        }),
+        'hintText': 'test',
+        'id': 'RmxvYXRRdWVzdGlvbjplbnZpcm9ubWVudGFsLXRlbg==',
+        'label': 'Bonnie Moreno',
+        'maxValue': None,
+        'meta': dict({
+        }),
+        'minValue': None,
```
luytena, line 901: shouldn't the step value be in here as well?

velrest: No the value is validated to be `>0` and the `step` value here is `-0.3`. Second param is `step`:
> `# name: test_save_float_question[float-question__configuration2-0.3-True]`

luytena (approved):

PR #2368, opened 2025-01-28
chore: version bump

velrest:


velrest:
self-merging, just a changelog/version update

PR #2391, opened 2025-03-04
Fix indirect calculations

winged:
### fix(structure): correctly update structure in-place for recalculations

When calculating a field, we must take better care to update
the dependents, invalidate all the cached values, and ensure the
intermediate values are correctly used.

### refactor(structure): rename `update_calc_dependents()`

Rename `update_calc_dependents()` to `recalculate_dependents()`, because
there's a similarly-named function in caluma_form.utils that does
something else, which is confusing

### DIFF
--- a/caluma/caluma_form/structure.py
+++ b/caluma/caluma_form/structure.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import collections
 import copy
+import itertools
 import typing
 import weakref
 from abc import ABC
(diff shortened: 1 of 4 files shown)


caluma/caluma_form/structure.py
```diff
@@ -175,9 +176,13 @@ def _store_question(self, question):
                 continue
             # Evaluator is only used for extracting answer transforms & friends,
             # so is field-independent here
-            for dependency_slug in self._evaluator.extract_referenced_questions(
-                jexl_expr
-            ):
+            referenced_questions = set(
+                itertools.chain(
+                    self._evaluator.extract_referenced_questions(jexl_expr),
+                    self._evaluator.extract_referenced_mapby_questions(jexl_expr),
```
czosel, line 182: Let's refactor this in the future: "extract_referenced_questions" should return _all_ referenced questions.

czosel (approved):

PR #2393, opened 2025-03-04
chore(ci): deprecate python and postgres versions

winged:
Python version 3.9 and PostgreSQL 12 are being deprecated. In their place, start testing with Python 3.13 and PostgreSQL 17


winged:
@anehx said it's ok, merging

PR #2396, opened 2025-03-05
fix(perf): don't load full document in memory if not needed for options

winged:
Question options can have a JEXL to decide whether they're hidden or not. Most of them however do not use this feature, and for those, loading the full document into memory just to evaluate their (non-existing) `is_hidden` expression is just a waste of resources/time.

### DIFF
--- a/caluma/caluma_form/validators.py
+++ b/caluma/caluma_form/validators.py
@@ -130,6 +130,14 @@ class AnswerValidator:
             # and use the structure code)
             return [o.slug for o in question.options.all()]
 
+        elif not validation_context:
+            # First see if we really need to build up the validation context. If
+            # not, we can speed things up significantly
+            all_options = list(question.options.all())
+            if all(opt.is_hidden in ("false", "") for opt in all_options):
+                # no JEXL evaluation neccessary
+                return [o.slug for o in all_options]
+
         field, _root = self._structure_field(document, question, validation_context)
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/validators.py
```diff
@@ -130,6 +130,14 @@ def _evaluate_options_jexl(
             # and use the structure code)
             return [o.slug for o in question.options.all()]
 
+        elif not validation_context:
+            # First see if we really need to build up the validation context. If
+            # not, we can speed things up significantly
+            all_options = list(question.options.all())
+            if all(opt.is_hidden in ("", None, "false") for opt in all_options):
```
anehx, line 137: Can `.is_hidden` even be an empty string or `None`?

anehx (changes requested):

caluma/caluma_form/validators.py
winged: You're right, it can't be `None`, and the default is `"false"`. I'd still take the empty string as possible shortcut nonetheless though, as a safety measure. It'd evaluate to `None` which is falsy, and even though it's unlikely (currently not used in any of our projects) I'd rather keep it

anehx: In theory we should also cover the case where all options `.is_hidden` is `"true"` but let's not as that really is no use case.

anehx (approved):

PR #2397, opened 2025-03-06
PoC: feat(models): add possibility to provide custom managers

anehx:


winged:
To be honest I'm not too keen on this idea. It introduces another complexity level that's hard to understand, as it literally extends the model classes "under the hood" and makes things "magically" appear (or they may be missing and nobody knows why).

It is also really hard to test correctly, as we'd need to have a reverse `add_to_class()` to avoid leaking.

Instead, I'd propose recommending the use of custom querysets without modifying the managers on the models. The following works just fine:

```python
class MyCustomQuerySet(QuerySet):
    def only_ready(self):
        return self.filter(status=WorkItem.STATUS_READY)

    class Meta:
        model = WorkItem
        name = "my_queryset"

...
def get_ready_workitems():
    return MyCustomQuerySet(model=WorkItem).only_ready()
```

anehx:
@StephanH90 @czosel @luytena Opinions to this? This would allow host applications to use something like `WorkItem.inquiries.for_instance(...)`

PR #2399, opened 2025-03-06
 chore(ci): update minimal python version in github actions

nlzet:
I believe this version should be updated accordingly to https://github.com/projectcaluma/caluma/pull/2393


anehx (approved):

winged (approved):
LGTM - but since we run compatibility tests with all supported python versions, we could probably bump this up to at least 3.12

PR #2415, opened 2025-04-02
fix(history): make history aware of potential proxy models

### LINKED ISSUE #2421: Historic model problem when using a proxy model
Change implementation of `HistoricalRecord` to something that works with proxy models e.g.  https://www.github.com/jazzband/django-simple-history/issues/544#issuecomment-1633269152

anehx:
`django-simple-history` currently creates historical records for proxy
models as well instead of using the historical records of the base
model. This issue is tracked in https://github.com/jazzband/django-simple-history/issues/544 but
doesn't seem to get any attention currently which is why we applied the
workaround from this comment:

https://github.com/jazzband/django-simple-history/issues/544#issuecomment-1538615799

This will make sure that changes on a proxy model will be tracked in the
historical records of the base model.

Resolves #2421

### DIFF
--- a/caluma/caluma_core/models.py
+++ b/caluma/caluma_core/models.py
@@ -1,6 +1,7 @@
 import uuid_extensions
+from django.apps import apps
 from django.db import models
-from simple_history.models import HistoricalRecords
+from simple_history.models import HistoricalRecords, registered_models
 
 
 def _history_user_getter(historical_instance):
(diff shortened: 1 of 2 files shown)


winged (changes requested):
Code LGTM (Nice work btw!), commit message is a bit short.. Would you mind adding a line or two explaining the change? I mean you have it in the docstring, but I find it useful to have an explanation in the commit messages as well

anehx:
> Code LGTM (Nice work btw!), commit message is a bit short.. Would you mind adding a line or two explaining the change? I mean you have it in the docstring, but I find it useful to have an explanation in the commit messages as well

Sure, makes sense. Will do.

winged (approved):
Awesome work, LGTM now!

PR #2378, opened 2025-02-21
feat(copy): adds on_copy to datasources to alter dynamic values

nlzet:
Add a on_copy feature for datasources to optionally alter the dynamic option answer value and label

PR #2420, opened 2025-04-04
fix(copy); change order of document-answer on_copy

nlzet:
In my previous PR https://github.com/projectcaluma/caluma/pull/2378 a new feature `on_copy` was introduced on an Answer, in order for a datasource to be able to change an answer value when a form gets copied (in case the answers change when copying a document, e.g. for dynamic answers in the form).

**Process of copying answers:**

- A document gets copied (main document to start with)
- Each answer in the document gets copied
- If an answer is a table answer, the AnswerDocument's for that table gets copied, recursing back to the document copy

**The problem with this setup:**

The order in which these answers are copied is depending on the recursive form structure, therefore an answer might get copied, before other dynamic answers are copied, and the on_copy process cannot complete correctly. In that version I already sorted the answers to copy non dynamic answers first to prevent this, but because of the recursion per document this can still fail depending on the form structure.

**New version:**

The copying of answers is reverted like it worked before (copy the answers without changing, and always copy AnswerDocument's as well). In the document copy process after the main document is copied (then the subforms are also done copying) the new answers are fetched and processed with the `on_copy` method. Answer values and dynamic options are updated on change, and dynamic options are removed on discard.

### DIFF
--- a/caluma/caluma_form/models.py
+++ b/caluma/caluma_form/models.py
@@ -422,34 +422,94 @@ class Document(core_models.UUIDModel):
             modified_by_user=user.username if user else None,
             modified_by_group=user.group if user else None,
         )
+
+        is_main_doc = False
         if not family:
+            is_main_doc = True
             family = new_document
 
-        non_dynamic_answers = self.answers.exclude(
-            question__type__in=[
-                Question.TYPE_DYNAMIC_CHOICE,
-                Question.TYPE_DYNAMIC_MULTIPLE_CHOICE,
-            ]
-        )
-        dynamic_answers = self.answers.filter(
-            question__type__in=[
-                Question.TYPE_DYNAMIC_CHOICE,
-                Question.TYPE_DYNAMIC_MULTIPLE_CHOICE,
-            ]
-        )
-
-        # first copy all non dynamic answers, then the dynamic answers
-        # because dynamic answers can reference to values of non dynamic answers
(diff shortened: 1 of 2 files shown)


caluma/caluma_form/models.py
```diff
@@ -422,33 +422,86 @@ def copy(self, family=None, user=None):
...
+                        new_dynamic_option.save()
+
+                    # delete the dynamic option if the slug is None
+                    elif not new_slug:
+                        new_dynamic_option.delete()
+
+                # save the answer value if changed after checking all dynamic options
+                if answer_value_changed:
+                    new_answer.save(update_fields=["value"])
+
```
winged, line 504: I think this whole block of code should be moved into it's own method - the method is otherwise way too overloaded. Create a new method `_update_dynamic_answers_after_copy()` (or similar - maybe you have a better idea for a name :-) )

winged (changes requested):
Nice catch on finding this problem! Just a small suggestion, otherwise LGTM I think.

Also, I would like to see the PR description also in the commit message. Makes code archeology much easier (Not the whole story about the previous PR, but a clear description of the change)

nlzet:
@winged updated to extract the logic into a new method, and updated the commit description to describe the changes

winged (approved):
LGTM now, nice work!

anehx (approved):
Nothing to add from my side :+1:

PR #2466, opened 2025-09-04
fix(schema): fix expected graphql type for is_hidden on options

anehx:


caluma/caluma_form/schema.py
```diff
@@ -169,6 +169,7 @@ def resolve_type(cls, instance, info):
 
 class Option(FormDjangoObjectType):
     meta = generic.GenericScalar()
+    is_hidden = QuestionJexl(required=True)
```
StephanH90, line 172: Ackchyually not a `QuestionJexl` but I will let it slide.

anehx: Actually it is. Option uses the same JEXL class the questions use.

PR #2489, opened 2025-09-25
fix(jexl): remove whitespace in stringify transform

luytena:
Remove whitespace in the result of stringify transform to ensure that the result of the stringify transform in caluma corresponds to the stringify transform in ember-caluma. The JSON.stringify method on an array doesn't include any whitespaces, while json.dumps of an array includes whitespaces between the elements by default, which leads to issues when comparing the result with values in a jexl. This can be set through the separators parameter in the json.dumps method.


maria-pw (commented):
`test_simple_transforms` also needs to be updated, otherwise LGTM

maria-pw (approved):

PR #2537, opened 2026-02-12
Validation graph improvements and fixes

anehx:
- **feat(validation): add graph to validate a single answer**
  This will mainly be used in the frontend to validate questions using format validators as those can't be validated in the frontend.
- **fix(validation): make sure validity graphs pass full context**
  When validating a table row document, the validation was only scoped to that specific document. This lead to errors when a question in that document referenced to other questions in the root document.


nlzet (approved):
LGTM, nice work! 👍

PR #2569, opened 2026-05-07
fix(validation): fix date and files answer validation

anehx:
When validating a whole document, answers of the type date or files were not being validated as their value (from a custom property) was not passed by the document validator.


nlzet (approved):
Nice, LGTM 👍

PR #2594, opened 2026-07-14
perf(form): recalculate only once in table rows

winged:
When updating a calcualted value inside a table, we must take extra care to only update when needed, as this can have compounding effects and slow down recalculation significantly.

We must avoid recalculation if a dependency of a calculation, as well as the calc field itself, are inside the same table, but in a different row.


nlzet (approved):
LGTM 👍

PR #2595, opened 2026-07-15
chore: rowset stringification should use slug from question, not form

winged:
In the structure code, the stringification (mostly used for debugging)
should use the question slug, not the (row) form slug.

Reason being that when printing/listing the structure (see the tests
that needed to be changed), the fieldsets below the rowset already
contain the row form's slug. Therefore, this change removes some
redundancy and increases useful information at the same time.

Note: Stacks on top of #2594 which needs to be merged first


anehx (approved):
LGTM 👍

PR #2596, opened 2026-07-15
chore: bump version for new perf improvements

winged:


winged:
(Draft, until verified that the effect of the perf fix is actually enough and we don't need to optimize more for the current issue)

anehx:
Superseeded by release v12.5.0 from today: https://github.com/projectcaluma/caluma/releases/tag/v12.5.0

PR #2597, opened 2026-07-20
fix(jexl): respect default value in answer transform for hidden fields

anehx:
Previously, the answer transform ignored any passed default value when the field was hidden, returning the empty value instead. Now hidden fields behave the same as empty fields and fall back to the default when one is given.

This matches the frontend, which already uses the passed default for the same case, and is what users expect.

https://github.com/projectcaluma/ember-caluma/blob/d1165a89a67a4d57baa2fb90bb23cc9ac15d8074/packages/form/addon/lib/document.js#L287-L323

### DIFF
--- a/caluma/caluma_form/jexl.py
+++ b/caluma/caluma_form/jexl.py
@@ -66,12 +66,7 @@ class QuestionJexl(JEXL):
                     f"Question `{question_slug}` could not be found in form {self.field.get_form()}"
                 )
 
-        if field.is_hidden():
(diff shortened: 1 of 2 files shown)


winged (approved):
Nice catch, don't know why I split those cases back then..

anehx:
> Nice catch, don't know why I split those cases back then..

@winged Because this has been wrong since the beginning - this is not a regression of your refactoring but an issue in the initial implementation - I checked :wink:

PR #2617, opened 2026-07-29
perf(form): improve calculated-question performance significantly

winged:
This should greatly improve performance, as recalculation of calculated
fields is only done when absolutely necessary. This is done via a few
mechanisms:

A) don't do recalculation recursively anymore, but instead build up a
   topoligically-ordered list of the fields to be recalculated. This
   avoids recalculation of the same field within the same change (Request
   or similar)

B) Culling of dependent calc questions if none of their input changed:
   If we come to a question in the ordered list (see above) and the
   result of the recalculation of all it's dependencies has not yielded
   a change, we don't need to recalculate said question either.

C) After recalculation, don't refresh all calc dependents. This is not
   needed in our structured recalculation anymore, and slows us down
   by a big margin, as it could in theory hit the DB for dozens of answers
   that are going to be recalculated in a few milliseconds anyway.

Drive-By changes & cleanups:

- Move calc question code into it's own module
- optimize get_path and add row number to fieldsets, so we can easily
  lookup the row number of a table row
- refactor global recalculation code as well (the stuff that recalculates
  when a calc_expression changes, or a calc question gets added to a form)
- keep row number in field sets that are part of a RowSet, and output it
  in the print utility functions (print_structure() and list_structure())

### DIFF
--- a/caluma/caluma_form/domain_logic.py
+++ b/caluma/caluma_form/domain_logic.py
@@ -1,4 +1,3 @@
-from graphlib import TopologicalSorter
 from logging import getLogger
 from typing import Optional
 
@@ -10,7 +9,10 @@ from caluma.caluma_core.exceptions import ConfigurationError
 from caluma.caluma_core.models import BaseModel
 from caluma.caluma_core.relay import extract_global_id
 from caluma.caluma_form import models, validators
-from caluma.caluma_form.utils import recalculate_field
+from caluma.caluma_form.calc_questions import (
+    recalculate_all_calc_fields,
+    recalculate_dependent_fields,
+)
 from caluma.caluma_user.models import BaseUser
 from caluma.utils import update_model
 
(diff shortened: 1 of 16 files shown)


winged:
FYI: In a complex form structure, this caused an improvement of a save action from 90s down to ca 3s.

caluma/caluma_form/structure.py
```diff
@@ -610,10 +606,11 @@ def evaluate_jexl(self, expression: str, raise_on_error=True):
...
         except Exception as exc:
-            log.error(
-                f"Error while evaluating expression on question {self.slug()}: "
-                f"{expression!r}: {str(exc)}"
-            )
+            if settings.DEBUG:  # pragma: no cover
+                log.debug(
+                    f"Error while evaluating expression on question {self.slug()}: "
+                    f"{expression!r}: {str(exc)}"
+                )
```
nlzet: Few lines below for the raised exception it says `The system log contains more information`. This could be not true for this case after this log level change I think

nlzet (changes requested):

caluma/caluma_form/domain_logic.py
```diff
@@ -188,47 +190,8 @@ def recalculate_dependents(answer, update_info):
...
-
-                if need_recalc:
-                    log.debug(
-                        "update_calc_dependents(%s): updating question %s",
-                        answer,
-                        dep_field.question.pk,
-                    )
-                    recalculate_field(dep_field)
+
+        recalculate_dependent_fields(field)
```
nlzet: Not entirely sure here, but the part of the old version `# We treat all children of the table as "changed"`, and it uses the `update_info["created"]` to collect all calculated float fields in those rows, as far as I can tell this part is no longer present.

If I'm correct this could miss a recalculation of a new table row, which I guess is fine unless the calculation expression references a field on the root document ?

nlzet (changes requested):

caluma/caluma_form/structure.py
winged: Removed the `if settings.DEBUG` check again - this was added during analysis of the issue, and we needed to suppress these spammy messages. So the exception string below is correct again

caluma/caluma_form/domain_logic.py
winged: Will have to think about this for a while. I'm quite confident in the new code thought, but maybe will need another test case.

winged: You were right, I've added a test case for this explicit situation. Needed a bit of a rework, but I think we're fine now :-)

nlzet (approved):
From code perspective it LGTM 👍

PR #2618, opened 2026-07-29
chore: disable tests for old migrations

winged:
These tests are rather slow and don't provide any value, as the code being tested (old migrations) are not expected to run anymore anywhere, as they happened long ago.


anehx (approved):

nlzet (approved):
LGTM 👍

PR #2619, opened 2026-07-30
 fix(calc-questions): robust recalculation on deletion and form changes

winged:
Several operations did not correctly trigger the recalculation
of dependent calculated fields, leading to stale data.

Improve the test suite to cover these issues, and fix the found issues.

Refactor answer and document deletion by moving the code into their own
domain logic classes, as other operations have been done before.
This allows us to update the dependent calc questions more easily where
needed.

Add `post_save` and `post_delete` signals for `FormQuestion` and `Question`
to automatically trigger recalculations when questions are added to,
or removed from forms. This allows calculated fields to react to
newly available dependencies or missing ones.

Improved recalculation by refining the culling logic in
`DependencyList.perform_recalculation()`. Now it always recalculates
fields that lack an answer. This handles new table rows and schema
changes better, without requiring manual root collection in the domain
logic.

### DIFF
--- a/caluma/caluma_form/calc_questions.py
+++ b/caluma/caluma_form/calc_questions.py
@@ -38,13 +38,13 @@ def update_calc_dependents(slug, old_expr, new_expr):
     questions = models.Question.objects.filter(pk__in=list(to_add | to_remove))
 
     for question in questions:
-        deps = set(question.calc_dependents)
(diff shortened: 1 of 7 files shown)


nlzet (approved):
LGTM 👍

PR #2628, opened 2026-08-04
Improve JEXL context

anehx:
### test(jexl): replace context snapshot with explicit assertions

Assert the concrete values and the full set of context paths instead of
comparing against a snapshot. Snapshots obscured which parts of the
context are actually contractual and made unintended changes easy to
accept blindly.

Parametrize over how the document is attached (case, work item, child
case) so the `info.case` block is covered too, and build the context via
`DocumentValidator.get_validation_context` so the global context is
populated the same way as in production.

The path assertions also pin down two known warts: the context exposes
`info.document` and `info.work_item`, the latter with the work item's
entire model data. Those will be fixed in the next commit.

### chore(jexl): remove undocumented `work_item` and `document` from JEXL context

Neither property was documented, nor referenced by the frontend, so they
were never meant to be used in expressions. Both also leaked model
internals into the context: `document` passed the Django model instance
directly, and `work_item` exposed every single attribute of the work
item via its `__dict__`. We don't want that data polluting the context,
so we remove it.

### feat(jexl): expose case and work item meta in JEXL context

Add `info.case.meta` and `info.case.root.meta`, and reintroduce
`info.workItem` in a curated form with only `task` and `meta` - as
opposed to the full model `__dict__` that was removed before.

- [Frontend PR](https://github.com/projectcaluma/ember-caluma/pull/3071)
- [Docs PR](https://github.com/projectcaluma/docs/pull/12)


winged (approved):
LGTM!

